PR #6, opened 2018-10-01
Herokify

palmfjord:
Adds basic auth and heroku build steps. This *should* work ™️


hedvigoscar:
Looks like there's 2 branch builds going on, what's up with that?

palmfjord:
@hedvigoscar i know, it keeps happening on all repos i connect travis to :( it builds both PR and branches even though they're the same commits?

PR #7, opened 2018-10-02
State session storage

palmfjord:
Like the title says ☝️ Stores the state in a session cookie


Iamsamwhoami:
Just a tought, maybe persistance should be a concept that belongs to https://github.com/diegohaz/constate? similar to `redux-persist` and the likes :)

palmfjord:
@Iamsamwhoami i agree, but maybe a do-later thing?

palmfjord:
also i noticed the reset doesn't reset the form and current steps, so i should probably fix that hehe 🙈

Iamsamwhoami:
@palmfjord yep yep

palmfjord:
@Iamsamwhoami ok im ready. plz gief another look?

src/pages/Chat/ChatConversation.tsx
```diff
@@ -1,35 +1,102 @@
...
+        prop('step1'),
+        isNil,
+        not,
+      ),
+      step1PropIsntNil('firstName'),
+      step1PropIsntNil('lastName'),
+      step1PropIsntNil('age'),
+    ])(steps)
+  ) {
+    return 5
```
Iamsamwhoami: what does this magic number mean? ✨

palmfjord: hehe, idk how to do it more neat currently, but it's the index of the message that we should just `appear` ( = this is where they are in the flow). May need to refactor this into a more clever solution when we get a more complex flow with branches n stuff. But is it OK for now or not really?

src/pages/Chat/steps/NameAgeInput.tsx
```diff
@@ -18,86 +23,96 @@ interface Actions {
...
+  pipe<React.ChangeEvent<HTMLInputElement>, Step1State[K], void>(
+    pathOr<string>('', ['target', 'value']),
+    (val: string | number) => chatState.setStep1Prop(field, val),
+  )
 
-export const NameAgeInput: React.SFC<Props> = ({ onSubmit }) => (
+const isDone = (
+  values: FormValues = { firstName: '', lastName: '', age: '' },
+) =>
+  pathOr(0, ['firstName', 'length'], values) > 0 &&
```
hedvigoscar: I'm not super fond of this. Could we perhaps use something like `yup` instead?

The equivalent would be something like this:

```
const schema = yup.object({
  firstName: yup.string().required(),
  lastName: yup.string().required(),
  age: yup
    .number()
    .positive()
    .moreThan(10)
    .required(),
})

const isDone = (values) => schema.isValidSync(values)
```

src/pages/Chat/steps/NameAgeInput.tsx
```diff
@@ -18,86 +23,96 @@ interface Actions {
...
+export const NameAgeInput: React.SFC<Props> = ({ onSubmit, appear }) => (
   <ChatContainer>
-    {({ setStep1 }) => (
+    {(chatState: ChatState & ChatActions) => (
       <Container<State, Actions>
         initialState={{ hasFocused: false }}
         actions={{ setFocused: () => (_) => ({ hasFocused: true }) }}
       >
         {(state) => (
-          <Formik<FormValues>
```
hedvigoscar: Did `Formik` become a bad fit with the dynamic nature of this form?

src/components/hedvig/chat.tsx
```diff
@@ -62,46 +62,69 @@ export const Typing: React.SFC<{ status: TransitionStatus }> = ({ status }) => (
...
 
 export interface ChatMessageProps {
+  appear?: boolean
   typingDuration?: number
   onTyped?: () => void
 }
 
+const HedvigIcon: React.SFC = () => (
+  <HedvigWrapper>
+    <Hedvig src="https://cdn.hedvig.com/identity/graphics/hedvig-symbol-color.svg" />
```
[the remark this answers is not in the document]
palmfjord: Maybe we should have a webpack loader (that also jams all svgs into one where we just reference the elements) instead?

src/pages/Chat/steps/NameAgeInput.tsx
palmfjord: Yes :( it keeps its own state, meaning we need some kind of event emitter/subscriber system to be able to reset it a level "above" in the component tree

src/components/hedvig/chat.tsx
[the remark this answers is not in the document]
hedvigoscar: I think that would be good. For example, this webpack loader? https://github.com/smooth-code/svgr/tree/master/packages/webpack :wink:

src/pages/Chat/steps/NameAgeInput.tsx
hedvigoscar: Reasonable. Formik is likely better for static-type forms, like we have in the dashborad.

src/pages/Chat/state.tsx
```diff
@@ -1,33 +1,68 @@
-import { Container, ContainerProps, StateUpdater } from 'constate'
+import { Container, ContainerProps, EffectProps } from 'constate'
+import { propOr } from 'ramda'
 import * as React from 'react'
+import { StorageContainer } from '../../utils/StorageContainer'
+
+export interface Step1State {
```
hedvigoscar: I think we should name the steps with descriptive names - having them referred to by index is likely to get painful down the line.

palmfjord: true!

src/pages/Chat/steps/NameAgeInput.tsx
palmfjord: yup!

src/components/hedvig/chat.tsx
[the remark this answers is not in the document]
palmfjord: Giving this some 2nd thought after i've woken up a little hehe. Idk if i wanna shoehorn svgs into the js bundle, if you get what i mean? Like currently we only have JS bundled with webpack, and i think that has its beauty right? Maybe we should use th svgr tool and bundle all svgs together and reference the same svg file, without bundling it with webpack?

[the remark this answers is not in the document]
palmfjord: also, maybe we could push that? hehe. I think this pr is growing to much already

PR #12, opened 2018-10-04
Feature/existing insurance + force to select selects

palmfjord:
like the title says  ☝️ This makes the flow go the whole way until the 'show offer', next steps are to actually get an offer based on the input and connect it to #8 🤘

PR #13, opened 2018-10-05
Giraffe + text keys

palmfjord:
Like the title says, throw in Apollo and hook it up to the 🦒 and txt 🔑 service.

What's good? SSR + text keys ofc 💥

Thinking i'll replace all texts with text keys in a different request, this is merely the infrastructure for using them.

<strike>I branched off #12 so i figured i'd set that as base branch until its merged</strike>


src/utils/apolloClient.ts
```diff
@@ -0,0 +1,29 @@
+import { InMemoryCache } from 'apollo-cache-inmemory'
+import { ApolloClient } from 'apollo-client'
+import { BatchHttpLink } from 'apollo-link-batch-http'
+
+export const GIRAFFE_ENDPOINT =
+  typeof window !== 'undefined' &&
+  (window as any).GIRAFFE_ENDPOINT !== undefined
+    ? (window as any).GIRAFFE_ENDPOINT
+    : process.env.GIRAFFE_ENDPOINT || 'https://graphql.dev.hedvigit.com/graphql'
```
Iamsamwhoami: Maybe we should throw if we can't find the giraffe? would be sad if we by mistake went to the dev environment in prod  😄

palmfjord: Very good point, that would suck very much. Maybe a check for wether we're in a dev build? I feel like i don't wanna specify these kind of env vars every time i start a dev build of the project

Iamsamwhoami: @palmfjord sounds good!

hedvigoscar: We could maybe use this principle in giraffe as well, where the config file defaults to dev-ish values right now in lieu of envvars.

src/utils/apolloClient.ts
```diff
@@ -1,12 +1,27 @@
 import { InMemoryCache } from 'apollo-cache-inmemory'
 import { ApolloClient } from 'apollo-client'
 import { BatchHttpLink } from 'apollo-link-batch-http'
+import { notNullable } from './nullables'
 
-export const GIRAFFE_ENDPOINT =
+let realGiraffeEndpoint: string
```
Iamsamwhoami: move this into a func and we can avoid the let maybe? 👯‍♂️

palmfjord: sure 🙌

hedvig.config.js
```diff
@@ -9,5 +9,5 @@ module.exports = {
   port: 8038, // The WDS port
   developmentPublicPath: 'http://0.0.0.0:8038/', // Client public path during development, i.e. "http://0.0.0.0:8081/". Port must match the port directive
   productionPublicPath: undefined, //  Client public path in production, i.e. "/assets/"
-  envVars: ['USE_AUTH', 'AUTH_NAME', 'AUTH_PASS', 'GIRAFFE_ENDPOINT'],
+  envVars: ['USE_AUTH', 'AUTH_NAME', 'AUTH_PASS', 'getGiraffeEndpoint'],
```
hedvigoscar: 🤔

palmfjord: hahahaha 😂 webstorm is a bit too eager sometimes hehe

PR #17, opened 2018-10-08
Kill lottie typing animation in favor of pure-css animation

palmfjord:
☝️
Lottie takes up about 200KB (non-gz), and even if we're code splitting it's used in a critical path so it needs to load before the user can do anything. I see no reason to use it for an animation this simple. But let's keep the `LazyLottie` component in case we want to use it later?


Iamsamwhoami:
@palmfjord could we maybe build this with `animated`? then we can get rid of the typing animation being a Lottie in the app as well :)

hedvigoscar:
I like this idea! Then we could probably use `react-native-animated-components` as well for web which would be really cool.

Iamsamwhoami:
@palmfjord yes, I was to write that yesterday, have we hooked this up to heroku? so that I can see how it looks.

palmfjord:
@Iamsamwhoami i'm afraid you'll have to run it locally, since the giraffe is down atm

PR #18, opened 2018-10-08
Input screen text keys

palmfjord:
first iteration. Thoughts?

### DIFF
--- a/src/pages/Chat/ResetButton.tsx
+++ b/src/pages/Chat/ResetButton.tsx
@@ -1,8 +1,15 @@
+import { TranslationsConsumer } from '@hedviginsurance/textkeyfy'
 import * as React from 'react'
 import { ChatContainer } from './state'
 
 export const ResetButton = () => (
   <ChatContainer>
-    {({ reset }) => <button onClick={reset}>reset session</button>}
+    {({ reset }) => (
+      <button onClick={reset}>
+        <TranslationsConsumer textKey="CHAT_RESET_LABEL">
+          {(t) => t}
+        </TranslationsConsumer>
+      </button>
+    )}
   </ChatContainer>
 )
--- a/src/pages/Chat/steps/CurrentInsuranceInput.test.tsx
+++ b/src/pages/Chat/steps/CurrentInsuranceInput.test.tsx
@@ -1,6 +1,7 @@
 import { Provider } from 'constate'
 import { mount } from 'enzyme'
 import * as React from 'react'
+import { MockTextKeyProvider } from 'utils/MockTextKeyProvider'
 import { createSession } from 'utils/sessionStorage'
 import { MockStorage } from 'utils/storage/MockStorage'
 import { WithStorageProps } from '../../../App'
@@ -12,7 +13,15 @@ it('handles form changes', () => {
     <Provider<WithStorageProps>
       initialState={{ storage: { session: createSession(new MockStorage()) } }}
     >
-      <CurrentInsuranceInput />,
+      <MockTextKeyProvider
+        textKeys={{
+          CHAT_INPUT_CURRENT_INSURANCE_TEXT: '{toggle}',
+          CHAT_INPUT_CURRENT_INSURANCE_TEXT_HAS_INSURANCE:
+            '{toggle} {currentInsurerMaybe}',
+        }}
+      >
+        <CurrentInsuranceInput />
+      </MockTextKeyProvider>
     </Provider>,
   )
   wrapper
--- a/src/pages/Chat/steps/CurrentInsuranceInput.tsx
+++ b/src/pages/Chat/steps/CurrentInsuranceInput.tsx
@@ -1,4 +1,8 @@
 import {
+  TranslationsConsumer,
+  TranslationsPlaceholderConsumer,
+} from '@hedviginsurance/textkeyfy'
+import {
   UserResponse,
   UserSelectInput,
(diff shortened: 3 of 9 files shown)


src/pages/Chat/steps/LivingSituationInput.test.tsx
```diff
@@ -5,6 +5,7 @@ import { mount } from 'enzyme'
 import * as React from 'react'
 import { createSession } from 'utils/sessionStorage'
 import { MockStorage } from 'utils/storage/MockStorage'
+import { MockTextKeyProvider } from '../../../utils/MockTextKeyProvider'
```
hedvigoscar: -`../../.....`

src/pages/Chat/steps/LivingSituationInput.tsx
```diff
@@ -44,7 +43,7 @@ const validationSchema = yup
     numberOfPeople: yup
       .number()
       .moreThan(0)
-      .lessThan(7, 'TOO MANY PEEPS') // TODO improve message
+      .lessThan(7, 'UNKNOWN_ERROR') // This should never happen since it s a dropdown
```
hedvigoscar: I think we can safely remove the error message from this validation rule 😉

src/pages/Chat/steps/CurrentInsuranceInput.tsx
```diff
@@ -96,20 +100,35 @@ export const CurrentInsuranceInput: React.SFC<CurrentInsuranceInputProps> = ({
...
+            }
+            replacements={{
+              toggle: (
+                <UserSelectInput
+                  onChange={chatState.setHasCurrentInsurance}
+                  value={getHasCurrentInsuranceInputValue(chatState)}
+                  id="hasCurrentInsurance"
+                >
+                  <option value="select" disabled />
+                  <option value="yes">Ja</option>
```
palmfjord: this should prolly be translated huh

src/utils/MockTextKeyProvider.tsx
```diff
@@ -0,0 +1,15 @@
+import { TranslationsContext } from '@hedviginsurance/textkeyfy'
+import * as React from 'react'
+
+export interface MockTextKeyProviderProps {
+  textKeys: { [key: string]: string }
+}
+
+export const MockTextKeyProvider: React.SFC<MockTextKeyProviderProps> = ({
```
palmfjord: this should probably be moved to `@hedviginsurance/textkeyfy` but maybe later

src/pages/Chat/steps/CurrentInsuranceInput.tsx
```diff
@@ -108,15 +108,23 @@ export const CurrentInsuranceInput: React.SFC<CurrentInsuranceInputProps> = ({
...
-                <UserSelectInput
-                  onChange={chatState.setHasCurrentInsurance}
-                  value={getHasCurrentInsuranceInputValue(chatState)}
-                  id="hasCurrentInsurance"
-                >
-                  <option value="select" disabled />
-                  <option value="yes">Ja</option>
-                  <option value="no">Nej</option>
-                </UserSelectInput>
+                <TranslationsConsumer textKey="CHAT_INPUT_CURRENT_INSURANCE_HAS_CURRENT_INSURANCE_TRUE">
```
Iamsamwhoami: you could wrap the individual <options /> instead 💅

src/pages/Chat/ResetButton.tsx
```diff
@@ -1,8 +1,15 @@
+import { TranslationsConsumer } from '@hedviginsurance/textkeyfy'
 import * as React from 'react'
 import { ChatContainer } from './state'
 
 export const ResetButton = () => (
   <ChatContainer>
-    {({ reset }) => <button onClick={reset}>reset session</button>}
+    {({ reset }) => (
+      <button onClick={reset}>
+        <TranslationsConsumer textKey="CHAT_RESET_LABEL">
```
Iamsamwhoami: matter of taste, but i tend to wrap the component needing text as a child, so doing:

```
<Consumer>
{text => <button>{text}</button>}
</Consumer>
```

But I'm fine with whatever :)

palmfjord: haha i think i've done mix, nice catch. Lets be consistent

src/pages/Chat/steps/CurrentInsuranceInput.test.tsx
```diff
@@ -12,7 +13,15 @@ it('handles form changes', () => {
     <Provider<WithStorageProps>
       initialState={{ storage: { session: createSession(new MockStorage()) } }}
     >
-      <CurrentInsuranceInput />,
+      <MockTextKeyProvider
```
Iamsamwhoami: What's the use-case for mocking? :) Shouldn't textkeyfy just return the passed text key? or does it throw due to it not having a provider? (in that case we should just fix that)

palmfjord: Well, we could pull in `MockedProvider` from `react-apollo/test-utils` and mock the network request. The problem is that the test depends on the items that the text key replaces, meaning the text key itself isn't good enough. So we need the actual text, not the text key.

Iamsamwhoami: Ah, ok fine then!

PR #21, opened 2018-10-09
Feature/signup design

frapperino:
The design for the signup screen.


src/pages/Sign/index.tsx
```diff
@@ -0,0 +1,22 @@
+import { TopBar } from 'components/TopBar'
+import * as React from 'react'
+import styled from 'react-emotion'
+import { SignUp } from './sections/SignUp'
+
+const Wrapper = styled('div')({})
+
+export const Sign: React.SFC<{}> = () => (
```
palmfjord: just fyi, the generic defaults to `{}` so  `React.SFC` is sufficient. Not that it matters much but you can save a few keystrokes here and there :D

src/pages/Sign/sections/SignUp.tsx
```diff
@@ -0,0 +1,180 @@
...
+  fontSize: '16px',
+})
+
+interface Props {
+  title: string
+  adress: string
+  buttonText: string
+  inputTitleEmail: string
+  inputTitlePersonalNumber: string
+  errorMessage: string
```
palmfjord: optional maybe? in case we dont want to show it

src/pages/Sign/sections/SignUp.tsx
```diff
@@ -0,0 +1,180 @@
...
+              <Header>
+                {props.title} {props.adress}
+              </Header>
+            </HeaderWrapper>
+            <Form onSubmit={state.handleSignup}>
+              <InputTitle>{props.inputTitleEmail}</InputTitle>
+              <InputField
+                type="text"
+                value={state.email}
+                onChange={() => state.updateEmail(state.email)}
```
palmfjord: this looks suspicious 🤨 shouldn't this be `onChange={event => state.updateEmail(event.target.value)` for the form to work?

palmfjord: But if we're refactoring we might as well throw in Formsik (as discussed irl)

frapperino: yes to both:)

src/pages/Sign/sections/SignUp.tsx
palmfjord: i think we may need different error messages depending on which error we get back. so then we could just remove it if we want to hide it. Also i don't think we'll use a cms that traditional, we'll use text keys instead so i think its a non-issue

src/pages/Sign/sections/SignUp.tsx
```diff
@@ -102,79 +102,81 @@ const ErrorText = styled('div')({
...
   inputTitlePersonalNumber: string
-  errorMessage: string
+  errorMessage?: string
 }
 
-interface State {
-  email: string
-  personalNumber: number
-  signUpError: string
+const validatePersonalNumber = (value: any) => {
```
palmfjord: imo we should use `yup` for these type of things. see ie https://github.com/HedvigInsurance/web-onboarding/blob/master/src/pages/Chat/steps/LivingSituationInput.tsx#L36

src/pages/Sign/sections/SignUp.tsx
```diff
@@ -116,21 +117,15 @@ interface Props {
...
-  if (!/^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$/i.test(value)) {
-    errorMessage = 'Whops! Formatet verkar vara fel på angiven email.'
-  }
-  return errorMessage
-}
+const UserSchema = Yup.object().shape({
+  personalNumber: Yup.string()
+    .min(12, 'Personnumret måste vara 12 siffor')
+    .max(12, 'Whops! Är det endast 12 siffor?')
+    .required('Krävs för signering'),
```
palmfjord: maybe max 13? if someone adds ie. a dash?
alternatively something like `.matches(/^\d{8}[\s\-]?\d{4}$/, ...)` instead?

frapperino: agree, solved and changed to interval between 6-8 to allow for 10 and 12 digit formats!

palmfjord: ah good!

palmfjord: nice!

palmfjord: just dont forget text keys for validation messages too :D

PR #25, opened 2018-10-11
Add ICE middleware and a panic room

palmfjord:
Panic room = intentionally throw errors in case we want to see that error reporting works as expected
ICE reverse proxies https://github.com/HedvigInsurance/cdn-files/pull/7 so that needs to be merged and CF needs to be invalidated be4 this'll work

ping @HedvigInsurance/clients

### DIFF
--- a/src/server/middleware/enhancers.ts
+++ b/src/server/middleware/enhancers.ts
@@ -51,6 +51,7 @@ export const inCaseOfEmergency: Middleware = async (ctx, next) => {
   try {
     await next()
   } catch (e) {
+    ctx.status = 500
     try {
       ;(ctx.state.getLogger('emergency') as Logger).error(
         'Uncaught error in request, requesting 500 page',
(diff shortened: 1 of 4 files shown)


src/server/middleware/enhancers.ts
```diff
@@ -36,3 +38,41 @@ export const logRequestMiddleware: Middleware = async (ctx, next) => {
     throw e
   }
 }
+
+const handleSuperFatalError = (ctx: Context) => (superFatalError: Error) => {
+  ;(ctx.state.getLogger('emergency') as Logger).fatal(
```
palmfjord: Do we have to cast here? :(

src/serverEntry.tsx
```diff
@@ -45,6 +48,12 @@ reactPageRoutes.forEach((route) => {
   server.router.get(route.path, getPage)
 })
 
+server.router.get('/panic-room', async () => {
```
palmfjord: This is going away I assume? :)

palmfjord: no, i was thinking we can keep it in order to test web onboarding panics

src/server/middleware/enhancers.ts
palmfjord: yes and no, koa ctx = always any, so just to to get the autocomplete hehe

PR #28, opened 2018-10-15
Chat steps functionality improved

palmfjord:
Basically ask one thing at a time instead of asking massive questions.

@HedvigInsurance/clients


src/pages/Chat/steps/ApartmentTypeInput.tsx
```diff
@@ -0,0 +1,156 @@
...
+
+const validationSchema = yup
+  .object<Partial<LivingSituationState>>({
+    apartmentType: yup
+      .mixed()
+      .oneOf(Object.keys(ApartmentType))
+      .required(),
+    size: yup
+      .mixed()
+      .test({
```
hedvigoscar: What's up here? 🙂

palmfjord: This can be an empty string _or_ a number, and it may not be over 250. So it's quite tricky to validate since emptystring becomes NaN if we make it a number, so this is the workaround :)

hedvigoscar: Isnt empty string an invalid case though? Given that we `.test()` away that.

palmfjord: sorry, i'm mixing up possible states and valid states when trying to explain it 😅

the input _could be_:
a) an empty string
b) a number

the input _is valid when_ it's:
a) a number
b) > 0
c) < 250

palmfjord: the alternative would be to make the input _number only_, but that would mean we need to start at 0 and they can't really remove the 0, which isn't pretty for the end user

palmfjord: (after trying to get rid of this, following an irl discussion)
We cant get rid of this, because `empty string` -> `yup.number()` = ``size must be a `number` type, but the final value was: `NaN` (cast from the value `""`).`` :(

hedvigoscar: That's all I need to hear - lets keep it as is! :+1:

PR #29, opened 2018-10-15
Feature/hook up signing

frapperino:


src/pages/Sign/sections/SignUp.tsx
```diff
@@ -137,62 +196,216 @@ export const SignUp: React.SFC = () => (
...
+                                  <div>
+                                    <ErrorText>
+                                      <TranslationsConsumer textKey="SIGN_BANKID_STANDARD_ERROR_MESSAGE">
+                                        {(errorText) => errorText}
+                                      </TranslationsConsumer>
+                                    </ErrorText>
+                                  </div>
+                                )
+                              }
+                              if (data) {
```
palmfjord: can we move these nested branching statements into a util functions?
a) nesting branching statements gets quite hard to read
b) We're mixing view and business logic here imo
c) I don't like branching in general, so if we could solve this in another approach i would be very happy :D

src/pages/Sign/sections/SignUp.tsx
```diff
@@ -137,62 +196,216 @@ export const SignUp: React.SFC = () => (
...
+                                      }
+                                    } else {
+                                      return null
+                                    }
+                                  case SIGNSTATE.COMPLETED: // TODO: backend sets this to null atm
+                                    if (
+                                      dataStatus.status ===
+                                      BANKIDSTATUS.COMPLETE
+                                    ) {
+                                      window.location.href = '/download'
```
palmfjord: this should probably be a `Redirect` component from `react-router-dom` so we don't loose what we have in memory

src/pages/Sign/sections/SignUp.tsx
```diff
@@ -137,62 +196,216 @@ export const SignUp: React.SFC = () => (
...
+                                    } else {
+                                      return null
+                                    }
+                                  case SIGNSTATE.FAILED:
+                                    if (
+                                      dataStatus.status === BANKIDSTATUS.FAILED
+                                    ) {
+                                      switch (dataStatus.code) {
+                                        case 'expiredTransaction':
```
palmfjord: why did we not make these into enums?

src/pages/Sign/sections/SignUp.tsx
```diff
@@ -121,6 +131,55 @@ const UserSchema = Yup.object().shape({
     .required('SIGN_PERSONAL_NUMBER_REQUIRED'),
```
palmfjord: Imo `UserSchema` -> `userSchema` as it is not static

src/pages/Sign/sections/SignUp.tsx
frapperino: They are strings in the BE so I wanted to be consistent

src/pages/Sign/sections/SignUp.tsx
```diff
@@ -3,9 +3,13 @@ import {
   TranslationsConsumer,
   TranslationsPlaceholderConsumer,
 } from '@hedviginsurance/textkeyfy'
+import { SessionContainer } from 'containers/SessionContainer'
 import { Field, Form, Formik } from 'formik'
+import gql from 'graphql-tag'
 import * as React from 'react'
+import { Mutation, Subscription } from 'react-apollo'
 import styled from 'react-emotion'
+import { Redirect } from 'react-router'
```
palmfjord: just fyi, this should be `react-router-dom` :) `react-router` works but only for legacy compatibility reasons and its deprecated

src/containers/OfferContainer.tsx
```diff
@@ -0,0 +1,61 @@
...
+export const OfferContainer: React.SFC<OfferContainerProps> = ({
+  children,
+}) => (
+  <Query<OfferData> query={OFFER_QUERY}>
+    {({ loading, error, data }) => {
+      if (loading) {
+        return <div>Loading</div>
+      }
+      if (error || !data) {
+        return <pre>{JSON.stringify(error, null, 2)}</pre>
```
hedvigoscar: We probably need to remove this 🤷‍♂️

src/containers/OfferContainer.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export const OfferContainer: React.SFC<OfferContainerProps> = ({
+  children,
+}) => (
+  <Query<OfferData> query={OFFER_QUERY}>
+    {({ loading, error, data }) => {
+      if (loading) {
+        return <div>Loading</div>
+      }
+      if (error || !data) {
```
hedvigoscar:
```suggestion
      if (error) {
```

src/containers/OfferContainer.tsx
```diff
@@ -0,0 +1,61 @@
...
+interface OfferContainerProps {
+  children: (offer: OfferData) => React.ReactNode
+}
+
+export const OfferContainer: React.SFC<OfferContainerProps> = ({
+  children,
+}) => (
+  <Query<OfferData> query={OFFER_QUERY}>
+    {({ loading, error, data }) => {
+      if (loading) {
```
hedvigoscar:
```suggestion
      if (loading || !data) {
```

src/pages/Sign/sections/SignSubscription.tsx
```diff
@@ -0,0 +1,161 @@
...
+        case SIGNSTATE.IN_PROGRESS:
+          if (dataStatus.status === BANKIDSTATUS.PENDING) {
+            return <BankidStatus message={dataStatus.code} />
+          } else {
+            return null
+          }
+        case SIGNSTATE.COMPLETED:
+          if (dataStatus.status === BANKIDSTATUS.COMPLETE) {
+            return <Redirect to="/download" />
+          } else {
```
hedvigoscar: Lets drop all these `else return null`-cases: if we have a `default: return null` we can rely on that case to improve conciseness and readability.

```suggestion

```

src/pages/Sign/sections/SignUp.tsx
```diff
@@ -121,79 +149,137 @@ const UserSchema = Yup.object().shape({
...
+                      address: offer.insurance.address,
+                    }}
+                  >
+                    {(title) => title}
+                  </TranslationsPlaceholderConsumer>
+                )}
+              </OfferContainer>
+            </Header>
+          </HeaderWrapper>
+          <SessionContainer>
```
hedvigoscar: We should remove this `<SessionContainer />` as it no longer adds any value, given our `<SessionTokenGuard />`at the top.

```suggestion

```

src/pages/Sign/sections/SignUp.tsx
```diff
@@ -121,79 +149,137 @@ const UserSchema = Yup.object().shape({
...
+                              <TranslationsConsumer textKey="SIGN_INPUT_ONE_TITLE">
+                                {(title) => title}
+                              </TranslationsConsumer>
+                            </InputTitle>
+
+                            <InputField
+                              name="email"
+                              id="email"
+                              style={{
+                                borderBottom: touched.email
```
palmfjord: this gets so hard to read... could we maybe move this out to a util function?

src/pages/Sign/sections/SignUp.tsx
```diff
@@ -121,79 +149,137 @@ const UserSchema = Yup.object().shape({
...
+                            <InputTitle>
+                              <TranslationsConsumer textKey="SIGN_INPUT_TWO_TITLE">
+                                {(title) => title}
+                              </TranslationsConsumer>
+                            </InputTitle>
+                            <InputField
+                              name="personalNumber"
+                              id="personalNumber"
+                              style={{
+                                borderBottom: touched.personalNumber
```
palmfjord: ...also we're kinda repeating ourselves. We should definitely move this out to a utils function

src/pages/Sign/sections/SignUp.tsx.orig
```diff
@@ -0,0 +1,571 @@
+import { colors, fonts } from '@hedviginsurance/brand'
```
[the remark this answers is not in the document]
frapperino: omg

hedvigoscar:
Lets get an internal launch rolling! :tada:

PR #35, opened 2018-10-18
Slide in messages so the chat doesn't jump as much

palmfjord:
smoooooooooth.... ✅

@HedvigInsurance/clients


src/components/animations/appearings.tsx
```diff
@@ -26,3 +26,26 @@ export const fadeUp = keyframes({
...
+  })
+
+export const SlideUp = styled('div')(
+  ({
+    maxHeight = 300,
+    initialMaxHeight = 0,
+  }: {
+    maxHeight?: number
+    initialMaxHeight?: number
+  }) => ({
```
Iamsamwhoami: add `will-change: max-height` here for some extra fps :)

src/components/animations/appearings.tsx
```diff
@@ -26,3 +26,26 @@ export const fadeUp = keyframes({
...
   animation: `${fadeUp} 300ms forwards`,
 })
+
+export const slideUp = (maxHeight: number, initialMaxHeight = 0) =>
+  keyframes({
+    from: {
+      maxHeight: initialMaxHeight,
+    },
+    to: {
+      maxHeight,
```
Iamsamwhoami: Just FYI the easing gets messed up if you don't use the exact height of the element as maxHeight, as it will interpolate in a curve going from say 0 to 300, so it might not look as good as it could if you would have the exact height. 

I would say that you should render the element off screen first to get the exact height :)

src/components/animations/appearings.tsx
```diff
@@ -26,3 +26,26 @@ export const fadeUp = keyframes({
...
+    },
+  })
+
+export const SlideUp = styled('div')(
```
Iamsamwhoami: Is this really `SlideUp` maybe more something like `AnimatedHeight` is more descriptive for its purpose? :)

src/components/animations/appearings.tsx
palmfjord: Good point, but after an irl discussion i think we will push this for later :)

PR #39, opened 2018-10-18
Feature/top bar pixels

frapperino:
Working on all window sizes and is consistent with hedvig.com navbar


src/components/TopBar.tsx
```diff
@@ -135,6 +161,13 @@ export const TopBar: React.SFC<Props> = (props) => (
             ))
           : null}
       </BarProgressContainer>
+      <CollapsedProgressContainer>
```
palmfjord: maybe we should actually hide the whole container if we don't have a progress?

frapperino: hmm, think the naming might be confusing, the CollapsedProgressContainer is only showing the current step in the process

frapperino: Oh I see what you mean!

src/components/TopBar.tsx
```diff
@@ -116,25 +139,32 @@ export const TopBar: React.SFC<Props> = (props) => (
   <Container>
     <Bar>
       <BarContainer>
-        <Logo src="/assets/offering/logo.png" />
+        <EscapeLink href="/hedvig">
```
palmfjord: should this be a link to `https://hedvig.com` rather?

src/components/buttons.tsx
```diff
@@ -36,3 +36,7 @@ export const Button = styled('button')(
     },
   }),
 )
+
+export const RefreshButton = styled('button')({})
```
palmfjord: why are these at the bottom or and why are they exported? :p

src/components/TopBar.tsx
frapperino: I'm also leaning towards that, but I guess it depends on what makes most sense in the context of web onboarding

src/components/buttons.tsx
frapperino: forgot to remove those, thx!

src/components/TopBar.tsx
```diff
@@ -65,36 +72,53 @@ const ProgressLabel = styled('div')({
...
+    flexDirection: 'column',
+    alignItems: 'center',
   },
 })
 
-const BarProgressContainer = styled('div')({
-  display: 'flex',
-  width: '60%',
+const CollapsedProgressWrapper = styled('div')({
+  display: 'none',
```
palmfjord: can we remove all these props if it's not gonna be visible anyways?

PR #42, opened 2018-10-18
Track things via Segment

hedvigoscar:


src/clientEntry.tsx
```diff
@@ -6,15 +6,20 @@ import { HelmetProvider } from 'react-helmet-async'
...
 import { createSession } from './utils/sessionStorage'
 
 ReactDOM.hydrate(
   <BrowserRouter>
-    <HelmetProvider>
-      <ApolloProvider client={apolloClient!.client!}>
-        <HotApp session={createSession(new CookieStorage({ expires: null }))} />
-      </ApolloProvider>
-    </HelmetProvider>
+    <PageTracker>
```
palmfjord: Don't we wanna add this for each route? Otherwise it's just gonna track the first page load right?

hedvigoscar: PageTracker actually hooks itself up to `history` and emits an `analytics.page()`-call each time the route changes. :+1:

palmfjord: right 😅too early to do PRs

PR #44, opened 2018-10-20
Enable dev tools for Constate in non-prod envs

hedvigoscar:
Setting `devtools=true` on the Constate `<Provider />` lets it interact
with `redux-devtools`, which makes for pleasant development.


hedvigoscar:
ikr? 😍

PR #45, opened 2018-10-24
Feature/top bar pixels

frapperino:


src/components/TopBar.tsx
```diff
@@ -62,45 +101,105 @@ interface Props {
...
     key: 0,
-    progressText: '1. Berätta om dig själv',
+    progressPercentage: '30%',
+    textKey: 'PROGRESS_TEXT_ONE',
   },
   {
     key: 1,
-    progressText: '2. Se din försäkring',
+    progressPercentage: '50%',
+    textKey: 'PROGRESS_TEXT_TWO',
```
palmfjord: i think these text keys should be what they represent over what order they come in (i.e. `PROGRESS_TEXT_OFFER`)

src/components/TopBar.tsx
```diff
@@ -62,45 +101,105 @@ interface Props {
...
-        {progress !== undefined
-          ? progressStrings.map((text) => (
-              <ProgressLabel
-                key={text.key}
+      <LogoWrapper>
+        <EscapeLink href="https://hedvig.com">
+          <Logo src="/assets/topbar/hedvig-wordmark-solid.svg" />
+        </EscapeLink>
+      </LogoWrapper>
+      {progress !== undefined ? (
```
palmfjord: wouldn't `progress !== undefined && (...` be sufficient?

palmfjord: applies to all of them obv

PR #46, opened 2018-10-25
Query sign state before starting subscription

palmfjord:
So we don't risk missing events

### DIFF
--- a/src/pages/Sign/sections/SignSubscription.tsx
+++ b/src/pages/Sign/sections/SignSubscription.tsx
@@ -1,9 +1,11 @@
 import { colors } from '@hedviginsurance/brand'
 import { TranslationsConsumer } from '@hedviginsurance/textkeyfy'
+import { ApolloError } from 'apollo-client'
 import gql from 'graphql-tag'
 import * as React from 'react'
-import { Subscription } from 'react-apollo'
+import { Query } from 'react-apollo'
 import styled from 'react-emotion'
+import { Mount } from 'react-lifecycle-components/dist'
 import { Redirect } from 'react-router-dom'
 import { trackEvent } from 'utils/tracking'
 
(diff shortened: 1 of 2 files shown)


src/pages/Sign/sections/SignSubscription.tsx
```diff
@@ -52,26 +54,37 @@ const SIGN_SUBSCRIPTION = gql`
...
+      collectStatus {
+        status
+        code
       }
+      signState
     }
   }
+`
+
+interface CollectStatus {
```
Iamsamwhoami: Just a tip for the future, I did this in the app: https://github.com/HedvigInsurance/app/pull/279

It handles all this crap for one so that you don't have to write exhaustive types for stuff that is in the introspection query anyways 😄

palmfjord: yeh i think that'd be good to have haha

palmfjord:
@Iamsamwhoami yes iz awsome, thanks!

src/pages/Sign/sections/SignSubscription.tsx
```diff
@@ -93,51 +106,88 @@ const handleMessage = (
   return textkeys[message]
 }
 
-export const SubscriptionComponent: React.SFC = () => (
-  <Subscription<SignStatusData> subscription={SIGN_SUBSCRIPTION}>
-    {({ data, loading, error }) => {
-      if (loading || !data) {
-        return null
+interface StateComponentProps {
+  signState?: SIGNSTATE
```
johantjellden: Sign state should never be null. https://github.com/HedvigInsurance/member-registry-service/blob/121314ae7cb9ecf3cac2eb694b517ef26520a43d/member-service/src/main/java/com/hedvig/memberservice/web/v2/dto/SignStatusResponse.java#L13

palmfjord: (after irl discussion) sign no, state can be null (undefined) b/c `data` can be null

palmfjord: if we haven't completed the query yet

PR #48, opened 2018-10-25
Fix: Make dropdown arrow clickable

palmfjord:
👆

this was way easier than i thought 🤷‍♀️


palmfjord:
@frapperino as it turns out 🤷‍♀️

PR #49, opened 2018-10-25
Add reset button to top bar so it looks decent

palmfjord:
also easier than i thought 🤷‍♀️


src/components/TopBar.tsx
```diff
@@ -17,7 +17,7 @@ const Bar = styled('div')({
   display: 'flex',
   flexDirection: 'row',
   alignItems: 'center',
-  justifyContent: 'end',
+  justifyContent: 'space-between',
```
frapperino: I think this change is the reason the content is skewed to the right when a button isn't present

src/pages/Chat/ResetButton.tsx
```diff
@@ -1,15 +1,31 @@
-import { TranslationsConsumer } from '@hedviginsurance/textkeyfy'
 import * as React from 'react'
+import styled from 'react-emotion'
 import { ChatContainer } from './state'
 
+const Wrapper = styled('div')({
+  paddingRight: 24,
```
frapperino: I would like this to be marginRight: 26 to be consistent with the other buttons:)

frapperino: and a width of 20%, breaking at 850px to 33%, and at 600px to 50%

palmfjord: first comment = 👍
2nd i don't understand. this is what it looks like;
screenshot 2018-10-25 at 13 52 52

frapperino: Ah, sry not so clear comment. Those are the breakpoints for when the progress steps gets replaced by a single step, and lastly not showing progress step at all for smallest size

palmfjord: but why do we need any width in the first place?

PR #51, opened 2018-10-25
Make Offer-page use more text keys

hedvigoscar:

PR #52, opened 2018-10-25
Feature/hedvig info images

frapperino:


assets/offering/anmal_skador.svg
```diff
@@ -0,0 +1,33 @@
+<svg width="270" height="258" viewBox="0 0 270 258" fill="none" xmlns="http://www.w3.org/2000/svg">
```
palmfjord: can we be more consistent in naming? Why are we a) naming this in swedish b) naming it in snake case

frapperino: good point, did not change from the download, and i think the majority are actually in swedish. Maybe save for later and make a separate ticket for all images to english and update all usages?

palmfjord: sure, but plz change to kebab case :)

src/pages/Offer/sections/HedvigInfo.tsx
```diff
@@ -57,36 +73,33 @@ const Title = styled('h3')({
...
 })
 
 const COLUMNS = [
   {
     key: 0,
-    title: 'Anmäl skador och få betalt på nolltid',
-    paragraph:
-      'Med vår smarta chattbot baserad på AI och maskininlärning kan vi snabbt göra bedömningar av skador och betala ut pengarna till dig utan krångliga formulär eller väntetider.',
-    image: '/assets/offering/placeholder.svg',
+    title: 'Enkel överblick',
```
hedvigoscar: These texts are in GraphCMS and should be edited there (see #51 )

frapperino: Nice!

PR #55, opened 2018-10-25
displaying address with a space, and correct interpunct

frapperino:


src/pages/Offer/sections/Offer.tsx
```diff
@@ -135,6 +135,10 @@ const COLUMNS = [
   },
 ]
 
+const addSpace = (postalNumber: string) => {
```
palmfjord: could be an implicit return right? :)

frapperino: :D

PR #61, opened 2018-10-26
Feature/offer get insured section

frapperino:


src/pages/Offer/sections/GetInsured.tsx
```diff
@@ -83,7 +130,7 @@ export const GetInsured: React.SFC<Props> = ({
             {() => (
               <TranslationsConsumer textKey="OFFER_SUMMARY_SIGN_CTA">
                 {(ctaText) => (
-                  <GetInsuredButton>
+                  <GetInsuredButton style={{ marginTop: '30px' }}>
```
palmfjord: inline styling :/

frapperino: I know, since the button is global:/ Any nicer way to do that?

palmfjord: you can pass props to the button component and pick it up in the styled component :) like this https://github.com/HedvigInsurance/web-onboarding/blob/master/src/components/buttons.tsx#L21

src/pages/Download/sections/DownloadApp.tsx
```diff
@@ -92,7 +92,7 @@ export const DownloadApp: React.SFC = () => (
           {(insuredText) => insuredText}
         </TranslationsConsumer>
       </InsuredText>
-      <GetInsuredButton>
+      <GetInsuredButton margin="0px">
```
palmfjord: we could just set a default to `0` and make it optional :D

src/components/buttons.tsx
```diff
@@ -3,11 +3,11 @@ import styled from 'react-emotion'
 import { Link } from 'react-router-dom'
 
 export const GetInsuredButton = styled('div')(
-  ({ margin }: { margin: string }) => ({
+  ({ margin }: { margin?: string }) => ({
     display: 'flex',
     flexDirection: 'row',
     justifyContent: 'center',
-    marginTop: margin,
+    marginTop: margin ? margin : 0,
```
palmfjord: sorry to complain so much but `margin || 0` would be sufficient really 😅

hedvigoscar:
```suggestion
    marginTop: margin || 0,
```

PR #64, opened 2018-10-26
Fix/redirect input to offer when has offer

palmfjord:
☝️


hedvigoscar:
@palmfjord we've got a failing test here that we need to fix 🤕

PR #65, opened 2018-10-26
Always format postal number with space

palmfjord:


src/pages/Chat/steps/AddressInput.tsx
```diff
@@ -116,19 +118,25 @@ export const AddressInput: React.SFC<AddressInputProps & Focusable> = ({
                     <TranslationsConsumer textKey="CHAT_INPUT_ADDRESS_POSTAL_NUMBER_PLACEHOLDER">
                       {(placeholder) => (
                         <UserTextInput
-                          type="number"
+                          type="text"
```
hedvigoscar: how about we use `type="tel"` here? Then we will get the nice number keyboard despite allowing other characters.

I know it's non-semantic and a bit hacky though 🙁

palmfjord: u sure? i don't think it does :/

hedvigoscar: it's possible that I'm just imagining it 😞

hedvigoscar:
```suggestion
                          type="text"
                          inputMode="numeric"
```

hedvigoscar: We'll save this for later.

PR #66, opened 2018-10-26
zebra insured amount and text keys

frapperino:


src/pages/Offer/sections/InsuredAmount.tsx
```diff
@@ -32,34 +32,58 @@ const Row = styled('div')({
...
+  display: 'inline',
+})
+
+const rows: ReadonlyArray<{
+  key: number
+  titleKey: string
+  amountKey: string
+}> = [
   {
+    key: 0,
```
hedvigoscar: Do we need this? I think `titleKey+amountKey` is unique enough 🤔

PR #71, opened 2018-10-26
sizes, fonts and color fixes

frapperino:


src/pages/Offer/sections/HedvigInfo.tsx
```diff
@@ -69,15 +68,14 @@ const Row = styled('div')({
   },
 })
 
-const Title = styled('h3')({
+const Title = styled('p')({
```
palmfjord: this looks weird to me. Why are we using a title as a `p`?

src/pages/Offer/sections/Offer.tsx
```diff
@@ -156,9 +152,12 @@ export const Offer: React.SFC<Props> = ({ signButtonVisibility, offer }) => (
           <HeaderBackground>
             <HeaderWrapper>
               <Header>
-                <TranslationsConsumer textKey="OFFER_HEADER">
+                <TranslationsPlaceholderConsumer
+                  textKey="OFFER_HEADER"
+                  replacements={{ name: offer.member.firstName }}
```
palmfjord: could we name this replacement `firstName`? i think that's what we use in other places

src/pages/Offer/sections/HedvigInfo.tsx
frapperino: It should be a p size wise, maybe title is the wrong naming though

palmfjord: i think if it is a title semantically we should just restyle it to be a P

frapperino: Hmm, not sure what you mean🤔 there is very little styling done here, so lets leave it as a `p`?

frapperino: or did you want a `div`?

src/pages/Offer/sections/Offer.tsx
```diff
@@ -133,7 +129,7 @@ interface Props {
 const COLUMNS = [
   {
     key: 0,
-    title: 'Lagenhetsskydd',
+    title: 'Lägenhetsskydd',
```
hedvigoscar: We should move this to text keys I think :wink:

src/pages/Offer/sections/InsuranceCoverage.tsx
```diff
@@ -339,6 +345,7 @@ export const InsuranceCoverage: React.SFC = () => (
                     </SwitcherItem>
                   ))}
                 </Switcher>
+                <PerilInfo>(Klicka på ikonerna för mer information)</PerilInfo>
```
palmfjord: text key?

PR #76, opened 2018-10-29
Split text key on download if currently insured

palmfjord:


src/pages/Sign/sections/SignUp.tsx
```diff
@@ -143,7 +143,7 @@ const inputBorderCheck = (
     ? errors
       ? '3px solid red'
       : [`3px solid ${colors.GREEN}`]
-    : [`3px solid ${colors.DARK_PURPLE}`]
+    : [`3px solid ${colors.PURPLE}`]
```
frapperino:
```suggestion
    : [`3px solid ${colors.DARK_PURPLE}`]
```

frapperino: As it is handled differently now with focus etc 😊

palmfjord: amaze, lets merge that before then!

PR #80, opened 2018-10-29
Feature/helmet

palmfjord:
⛑


src/serverEntry.tsx
```diff
@@ -43,6 +46,22 @@ const server = createKoaServer({
...
+    extendTypes: { json: ['application/csp-report'] },
+  }),
+)
+
+if (process.env.USE_HELMET === 'true') {
+  appLogger.info('Using helmet and strict CSP ⛑')
+  server.app.use(helmet)
+  server.router.use(helmet)
+} else if (process.env.NODE_ENV !== 'development') {
+  appLogger.warn(
```
hedvigoscar: I would even consider bailing at this point. Going without in production would not be great as the log statement says!

src/types/koa.d.ts
```diff
@@ -0,0 +1,8 @@
+import 'koa'
+
+declare module 'koa' {
```
hedvigoscar: 🙏

src/types/koa-bodyparser.d.ts
```diff
@@ -0,0 +1,18 @@
+declare module 'koa-bodyparser' {
```
hedvigoscar: :pray:

src/serverEntry.tsx
palmfjord: I don't agree. I.e. if we're putting up a PR instance or want to try something locally it's not obvious we care about CSPs.

PR #82, opened 2018-10-29
Add more csp allowed targets & fix 500 page

palmfjord:


palmfjord:
@hedvigoscar gaaah, screw gtm 🔩
nah jk, lemme know what we need

PR #89, opened 2018-10-29
Don't send postal numbers with spaces

palmfjord:
💌


palmfjord:
idk why travis is so slow but im merging anyways, it did pass

PR #90, opened 2018-10-30
Responsive fixes

palmfjord:
📱

This does in no way look pretty responsively, but at least its not broken hehe


src/components/hedvig/chat.tsx
```diff
@@ -1,6 +1,9 @@
 import { colors, fonts } from '@hedviginsurance/brand'
+import { Container } from 'constate'
 import * as React from 'react'
+import AnimateHeight from 'react-animate-height'
```
hedvigoscar: 💪

PR #97, opened 2018-10-31
Feature/track things more closely

hedvigoscar:


src/utils/tracking.ts
```diff
@@ -1,10 +1,32 @@
+import { CookieStorage } from 'cookie-storage'
+import * as merge from 'deepmerge'
+
+const cookie = new CookieStorage()
```
palmfjord: this'll only work client-side though. Is there any time we may want tracking on the server or no?

hedvigoscar: I see your point but I'm not sure if we ever want to do 'this kind' of tracking from SSR. What do you think?

palmfjord: lets leave it for now 👍 obviously it wont work at all so i think its ok

hedvigoscar:
@palmfjord I think this is ready for re-review 🙂

src/containers/SessionContainer.test.tsx
```diff
@@ -23,6 +23,7 @@ jest.mock('../client/apolloClient', () => ({
     },
   },
 }))
+jest.mock('uuid/v4', () => () => '1')
```
palmfjord: is this needed now when we did `__mocks__.*`?

hedvigoscar: Doubt it! Lets check
```suggestion
```

PR #102, opened 2018-11-05
Feature/legal text links old

frapperino:
Old as in the old safe version:)


src/utils/cdnFilesURL.ts
```diff
@@ -0,0 +1,31 @@
+import { InsuranceType } from 'utils/insuranceDomainUtils'
```
palmfjord: just checking, why not place this in `/src/utils/insuranceDomainUtils.ts`?

PR #103, opened 2018-11-05
Student offering label

palmfjord:
screenshot 2018-11-05 at 16 42 52


src/pages/Offer/components/StudentOfferBadge.tsx
```diff
@@ -0,0 +1,34 @@
...
+    width: 140,
+    height: 140,
+    textAlign: 'center',
+    fontWeight: 'bold',
+    backgroundColor: colors.PINK,
+    color: colors.WHITE,
+    borderRadius: '50%',
+
+    '@media (max-width: 560px)': {
+      display: 'none', // TODO fixme (this is a quick hack to get it working on desktop but not break mobile completely)
```
hedvigoscar: Something something no comments :wink:

No but when do we intend to fix this? 🙂

frapperino: Maybe we should discuss mobile on web onboarding as there are no designs for it, and no traffic is directed to it.

palmfjord: Haha if you have a suggestion on how to express this in code that is also easy to find if we're looking for todo/fixme comments i'm all ears ;)

On a more serious note i think its more about "if" over "when", i think it depends on how the web onboarding performs.

PR #104, opened 2018-11-06
Feature/input constraints

frapperino:


src/pages/Chat/steps/AgeInput.test.tsx
```diff
@@ -8,6 +8,35 @@ import { WithStorageProps } from 'utils/StorageContainer'
 import { ChatContainer } from '../state'
 import { AgeInput } from './AgeInput'
 
+it('only allows chars in age input', () => {
```
palmfjord: if you want to save some lines of code i think this could be expressed in `handles form changes`, it's just one assertion we need really right? I think this would be sufficient:
https://github.com/HedvigInsurance/web-onboarding/pull/104/files#diff-10a092b577aa0665e6f67f98a1b8a4dfR36

src/pages/Chat/steps/InsuranceTypeInput.test.tsx
```diff
@@ -24,6 +24,11 @@ it('handles form changes', () => {
     </Provider>,
   )
 
+  wrapper
+    .find('input#size')
+    .simulate('change', { target: { value: 'characters' } })
```
palmfjord: maybe we could add some numbers here just to be sure it doesn't swallow everything?

palmfjord: so something like `simulate 1blargh2blargh2` -> `expect 123`?

PR #111, opened 2018-11-15
Replace all tracking with `quepasa` and identify

hedvigoscar:
with memberId


.nvmrc
```diff
@@ -0,0 +1 @@
+10.11.0
```
palmfjord: ooh nice :D

hedvigoscar:
In test environments or while running unit tests? Because I think both should be mostly fixed, right? 🤔

palmfjord:
@hedvigoscar haha yes

hedvigoscar:
🙁

hedvigoscar:
@palmfjord ping 🙂

src/utils/tracking.ts
```diff
@@ -14,35 +11,26 @@ export interface UtmParams {
...
-    eventName,
-    merge(properties || {}, { context: { ...getUtmParamsFromCookie() } }),
-    options,
-  ]
-  logInDev('[ANALYTICS TRACK] ', event)
-  const castedWindow = window as any
-  if (castedWindow && castedWindow.analytics) {
-    castedWindow.analytics.track(...event)
-  }
+export enum CustomEvents {
```
palmfjord: i don't really like the naming `Custom`, can we make it something more like `ApplicationSpecific...` or so?

hedvigoscar: Sure!

hedvigoscar:
@palmfjord 🏓

PR #112, opened 2018-11-21
Feature/sign changes

frapperino:


src/pages/Offer/index.test.tsx
```diff
@@ -74,7 +74,6 @@ it('queries when it has a session', async () => {
   await mockNetworkWait()
   wrapper.update()
   expect(wrapper.find(Redirect)).toHaveLength(0)
-  expect(wrapper.find(Price).text()).toBe('99 kr/mån')
```
palmfjord: this may be the most important line in this test imo. We should still test this

src/pages/Offer/sections/InsuranceCoverage.tsx
```diff
@@ -335,57 +344,116 @@ export const InsuranceCoverage: React.SFC = () => (
...
-              ) : null}
-            </BigCol>
-          ))}
-          {state.showPerilNumber !== undefined &&
-          state.showIconNumber !== undefined ? (
-            <DropDownText>{state.textToShow}</DropDownText>
-          ) : null}
+
+          {PERILS.map(
+            (peril) =>
```
palmfjord: coming back to this i realise this is _quite_ heavy. Can we maybe make the peril into a separate component?

src/pages/Sign/index.tsx
```diff
@@ -1,20 +1,53 @@
...
+              <TranslationsConsumer textKey="SIGN_PAGE_TITLE">
+                {(t) => (
+                  <Helmet>
+                    <title>{t}</title>
+                  </Helmet>
+                )}
+              </TranslationsConsumer>
+              <TopBar progress={2} />
+              <div
+                style={{ backgroundColor: colors.OFF_WHITE, paddingBottom: 30 }}
```
palmfjord: no inline stylez plz

src/pages/Sign/index.tsx
```diff
@@ -1,20 +1,53 @@
...
-    <SignUp />
+    <SessionTokenGuard>
+      <OfferContainer>
+        {(offer) => {
+          if (!offer || !offer.insurance.type) {
+            return null
+          }
+          return (
+            <OuterWrapper>
+              {' '}
```
palmfjord: imo we shouldn't rely on whitespace for spacing, which i kind of assume this is

src/pages/Sign/sections/SignUp.test.tsx
```diff
@@ -85,7 +86,7 @@ it('queries when it has a session', async () => {
             }}
           >
             <MockTextKeyProvider textKeys={{ SIGN_HEADER_TITLE: '{address}' }}>
-              <SignUp />
+              <SignUp offer={mockOffer} />
```
palmfjord: if we pass the offer here, do wi still .need the `MeckedProvider`?

src/pages/Offer/sections/InsuranceCoverage.tsx
frapperino: yes maybe it is, but I don't think extracting everything but the header really changes much:/

src/pages/Offer/index.test.tsx
frapperino: done 👉

src/pages/Offer/sections/InsuranceCoverage.tsx
frapperino: I reverted to the old design, so nothing new here now

PR #115, opened 2018-12-03
Upgrade packages

palmfjord:
👮‍♂️


hedvigoscar:
Actually, it seems like the upgrade needs some lint fixes 🙂

PR #124, opened 2019-02-26
Change first message based on utm source

palmfjord:


palmfjord:
gonna close this for now, lets reopen if we care about it

PR #126, opened 2019-06-17
Referrals landing page

hedvigoscar:


src/pages/Referral/util.ts
```diff
@@ -0,0 +1,19 @@
+export const getFirebaseLinkDomain = () => {
+  if (
+    typeof window !== 'undefined' &&
+    (window as any).FIREBASE_LINK_DOMAIN !== undefined
+  ) {
+    return (window as any).FIREBASE_LINK_DOMAIN
```
palmenhq: we're making `return`s in virtually every branch here, do we really need all these `else`s n stuff? get's hard to read

src/routes.ts
```diff
@@ -34,5 +35,6 @@ export const reactPageRoutes = [
     exact: true,
   },
   { path: '/dont-panic/hedvig', Component: LazyDontPanic, exact: true },
+  { path: '/referrals/:code', Component: Referral, exact: true },
```
palmenhq: we need to remember to add this to CF but awesome 🙌

src/pages/Referral/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+  marginLeft: 16,
+  marginRight: 16,
+})
+
+const Centered = styled('div')({
+  marginTop: 65,
+  textAlign: 'center',
+})
+
+const LinkButton = styled(Button.withComponent('a'))({
```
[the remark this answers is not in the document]
hedvigoscar: ✅

src/pages/Referral/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+  fontWeight: 400,
+  padding: '16px 32px',
+})
+
+const PageContent: React.FunctionComponent<{ data: Data; code: string }> = ({
+  data: {
+    referralCampaignMemberInformation: {
+      incentive: { amount },
```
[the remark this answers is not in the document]
hedvigoscar: We had multiple values inside the `referralCampaignMemberInformation`-object before, but yes, it makes no sense now. I've refactored it around a bit

src/pages/Referral/index.tsx
```diff
@@ -0,0 +1,138 @@
...
+              return <div>Loading...</div>
+            }
+
+            return <PageContent data={data} code={code} />
+          }}
+        </Query>
+      )}
+    </Page>
+  ) : (
+    <Redirect to="/404" />
```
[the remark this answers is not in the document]
hedvigoscar: How would that look? Something like:

```tsx
...
<Mount on={() => (window.location = 'https://www.hedvig.com/404')} />
...
```
?

src/pages/Referral/util.ts
hedvigoscar: I tried to mirror the design found in `utils/apolloClient.ts::getGiraffeEndpoint`, but I'm open for suggestions on how to express it of course 🙂

src/routes.ts
hedvigoscar: Absolutely! 🙂

src/pages/Referral/index.tsx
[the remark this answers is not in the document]
palmenhq: yup look good. Maybe we could refactor it into a separate component to hide away the details of it

src/pages/Referral/util.ts
palmenhq: Blaming that file i feel a bit silly hehe 😬 imo not one of my strongest pieces of code. imo it should be refactored into something like

```ts
  if (
    typeof window !== 'undefined' &&
    (window as any).FIREBASE_LINK_DOMAIN !== undefined
  ) {
    return (window as any).FIREBASE_LINK_DOMAIN
 }

  if (process.env.NODE_ENV === 'development') {
    return process.env.FIREBASE_LINK_DOMAIN || 'https://hedvigtest.page.link'
  } 

  if (
    process.env.NODE_ENV !== 'development' &&
    process.env.FIREBASE_LINK_DOMAIN
  ) {
    return process.env.FIREBASE_LINK_DOMAIN
  }

  throw Error(
    'Unable to find firebase link domain - specify with envvar `FIREBASE_LINK_DOMAIN`',
  )
```

src/pages/Referral/index.tsx
```diff
@@ -0,0 +1,148 @@
...
+      >
+        <TranslationsConsumer textKey="REFERRAL_LANDINGPAGE_BTN_CTA">
+          {(text) => text}
+        </TranslationsConsumer>
+      </LinkButton>
+    </Centered>
+  </>
+)
+
+const MOCKED = true
```
palmenhq: this feels a bit suspicious also hehe

hedvigoscar: Interestingly enough it's going away completely - this view will no longer make API calls to giraffe.

So yeah, basically the entire reason I built in in web-onboarding has now gone away 😞

palmenhq: haha good 😅

src/pages/Referral/util.ts
hedvigoscar: Far prettier - I'm a huge fan of the early return over else's as well 😉

src/pages/Referral/index.tsx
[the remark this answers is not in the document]
hedvigoscar: Looks like the path never even matches without the parameter, so I'm just going to remove the code altogether

src/pages/Referral/index.tsx
```diff
@@ -0,0 +1,155 @@
...
+import { RouteComponentProps } from 'react-router'
+import { MobileContext } from 'utils/mobileContext'
+import { getFirebaseLinkDomain } from './util'
+
+type ReferralProps = RouteComponentProps<{
+  code: string
+}>
+
+const PageWrapper = styled('div')({
+  ['@media (min-width: 700px)']: {
```
palmenhq: no need for square brackets here, does prettier not remove this?

hedvigoscar: Nope 🤔 But I'll remove it!

palmenhq: haha good

PR #128, opened 2019-06-24
amends onboarding flow depending on  members current insurer

AlexSJ1990:


src/pages/Chat/state.tsx
```diff
@@ -28,9 +28,17 @@ export enum Insurer {
   ICA = 'ICA',
   GJENSIDIGE = 'Gjensidige',
   VARDIA = 'Vardia',
+  TRE_KRONOR = 'Tre Kronor',
   OTHER = 'OTHER',
 }
 
+export enum SwitchableInsurers {
```
palmenhq: Can't we make this a `ReadonlyArray<Insurer>` instead? It's so easy to make a typo or similar with this code duplication

src/pages/Offer/sections/HedvigSwitch.tsx
```diff
@@ -79,6 +81,28 @@ const ImageIcon = styled('img')({
...
+    paragraphKey: 'OFFER_NON_SWITCHABLE_COL_TWO_PARAGRAPH',
+  },
+  {
+    imageUrl: '/new-member-assets/offering/hedvig-dot-3.svg',
+    title: '3',
+    paragraphKey: 'OFFER_NON_SWITCHABLE_COL_THREE_PARAGRAPH',
+  },
+]
+
 const cols: ReadonlyArray<{
```
palmenhq: could we also rename this into `switchableCols`?

src/pages/Offer/sections/HedvigSwitch.tsx
```diff
@@ -101,20 +125,42 @@ const cols: ReadonlyArray<{
...
+
+const getOfferSwitchTitle = (currentInsuranceState: CurrentInsuranceState) => {
+  return Object.keys(SwitchableInsurers).includes(
+    currentInsuranceState.currentInsurer!,
+  )
+    ? 'OFFER_SWITCH_TITLE'
+    : 'OFFER_SWITCH_TITLE_NON_SWITCHABLE'
+}
+
+const getColumns = (currentInsuranceState: CurrentInsuranceState) => {
```
palmenhq: we could make an implicit return here :) Just something like `(...) => Object.keys(...)....` instead of declaring the full method body

src/pages/Chat/state.tsx
palmenhq: After some irl discussions lets keep it as is, this seems good enough 👍

palmenhq:
although plz update master be4 you merge

PR #137, opened 2019-08-07
Fix CSP for ad script

axelbacklund:
The online.adservicemedia.dk script from Google tag manager got blocked by the CSP directive 'script-src-elem'. 'Script-src-elem' should fallback to 'script-src' which we already have in the helmet, so the problem was the URL which had some bits and pieces ahead of its 'online.adservicemedia.dk'. At the top was '9270fc4b.id.opendns.com', so scripts with this url suffix are now allowed.


hedvigoscar:
This is probably not going to work as the ID is not static 😬

axelbacklund:
Will let this rest for a bit until we get an answer from the ad company

PR #140, opened 2019-08-15
Automatically consume discount codes provided through query parameters

hedvigoscar:
Turns out we already had infrastructure for this lmao


hedvigoscar:
Good point, will change

PR #143, opened 2019-09-02
Link header logo to start page + upgrade packages

palmenhq:


src/pages/Chat/ChatConversation.tsx
```diff
@@ -30,7 +30,11 @@ const getOfferQuestionTextKey = (
     return 'CHAT_HEDVIG_SHOW_OFFER_NO_INSURANCE'
   }
 
-  if (Object.values(SwitchableInsurers).includes(currentInsurer!)) {
+  if (
```
Iamsamwhoami: Maybe interesting for da future, but maybe we should add a query so that we can get the "SwitchableInsurers" from giraffe? I did that for the offering screens in-app here:  https://github.com/HedvigInsurance/giraffe/pull/465

palmenhq: ooh nice, didn't know this was a thing. I think yes in  the future but a bit out of scope for this pr. This is just a fix because typescript got a tiny bit more picky 😬

PR #144, opened 2019-09-03
Fix/heuristic fragment matcher crash

FredrikAreschoug:


FredrikAreschoug:
I've tested this and it works fine, we should probably investigate how we stitching Graphql schemas some more and maybe have a look at why this Union is acting up but this fixes the the issue we are having.

palmenhq:
@FredrikAreschoug what i ment was specficially if you reload the page (so you trigger the server-side rendering) on the sign page, the one that exploded before. I'm (almost) sure we will need to give the fragment matcher also to the server-side apollo client.

FredrikAreschoug:
Cool missed that that was a problem as well

palmenhq:
@FredrikAreschoug sorry, maybe i wasn't clear haha 😅

PR #147, opened 2019-09-11
Reduce initial loading time

oscarnylander:


src/pages/Offer/sections/offer/Discount.tsx
```diff
@@ -15,8 +15,9 @@ import { OfferData } from 'src/containers/OfferContainer'
 import * as Yup from 'yup'
 import { RedeemCodeMutation } from '../../containers/RedeemCodeMutation'
 
-const DiscountButtonLink = styled('button')({
+const DiscountButtonLink = styled('a')({
```
palmenhq: this seems odd to me, any particular reason?

palmenhq: Ah, i saw the commit message. When does it look terrible? I think that would be the case for virtually all buttons if that's the case no?

hedvigoscar: I mostly used this solution to get rid of the default button styling that web browsers sometimes ship. We decided that we wanted this button to instead be a link (https://trello.com/c/yVM1vUKl/191-web-replace-rabattkod-button-with-text), and this just makes it so that some browsers don't display it as a button 🙂

hedvigoscar: Firefox rendered it as a good old gray square which felt a bit off

hedvigoscar: The styling on the other buttons fully override the default button styling afaik, which makes this a non-issue for those buttons :slig

PR #149, opened 2019-09-25
Implement Adtraction

gustafguner:


src/utils/tracking.ts
```diff
@@ -43,3 +44,41 @@ export const { Track, TrackAction, Identify, IdentifyAction } = setupTrackers<
...
+  doEvent: () => void
+}
+
+export const adtraction = (
+  orderValue: number,
+  orderId: string,
+  emailAddress: string,
+  couponCode: string | null,
+) => {
+  // @ts-ignore
```
palmenhq: :( why?

src/utils/tracking.ts
```diff
@@ -43,3 +44,41 @@ export const { Track, TrackAction, Identify, IdentifyAction } = setupTrackers<
...
+  const adt: Adt = ADT
+  adt.Tag = adt.Tag || {}
+  adt.Tag.t = 3
+  adt.Tag.c = 'SEK'
+  adt.Tag.tp = 1412601108
+  adt.Tag.am = orderValue
+  adt.Tag.ti = orderId
+  adt.Tag.xd = md5(emailAddress)
```
palmenhq: Are we trying to scrub personal data or do they explicitly request us to md5 hash the emails?

src/server/page.tsx
```diff
@@ -60,6 +60,7 @@ const template = (
     Sentry.init(${JSON.stringify(sentryConfig())})
   </script>
   <script key="segment-snippet" nonce="${cspNonce}">${segmentSnippet}</script>
+  <script defer src="https://adtr.io/jsTag?ap=1412531808"></script>
```
palmenhq: this (and any other domains they might use) must be placed in the CSP headers

src/pages/Sign/sections/SignSubscription.tsx
```diff
@@ -188,16 +189,45 @@ const StateComponent: React.SFC<StateComponentProps> = ({
...
-              />
-            )}
-          </CurrentLanguage>
-        )
+
+        if (process.env.NODE_ENV !== 'test') {
+          return (
+            <OfferContainer>
+              {(offer) => {
+                adtraction(
```
palmenhq: this part bugs me. Why are we doing this on every render? Is that really what was intended?

src/utils/tracking.ts
gustafguner: They explicitly request the email addresses to be MD5-hashed...

src/utils/tracking.ts
gustafguner: I guess I could declare my own types for it, that's true :-)

src/server/page.tsx
gustafguner: Good catch! 💯

PR #155, opened 2019-10-23
Dreams ✨

gustafguner:


src/components/TopBar.tsx
```diff
@@ -106,15 +111,26 @@ const ButtonWrapper = styled('div')({
...
   '@media (max-width: 600px)': {
-    width: '50%',
+    width: '45%',
+    paddingRight: 20,
+    '& > *:last-of-type': {
+      marginRight: 10,
+    },
+    '& > *:last-child': {
+      marginRight: 0,
+    },
```
palmenhq: this seems a bit contradictory to me

hedvigoscar: The CSS selector is indeed a bit weirdly written. It's probably hard to write it better though given that the button-property is abstract enough that it takes any element.

It will also break when a third button is introduced into the mix.

Given the tight deadline on this feature, we're going ahead and merge it anyways.

PR #156, opened 2019-11-01
save partnership cookie

FredrikAreschoug:


src/server/page.tsx
```diff
@@ -106,6 +106,9 @@ export const getPage: Koa.Middleware = async (ctx) => {
       partner: serverCookieStorage.getItem('_hvpartner'),
     })
   }
+  if (ctx.query.partner) {
```
palmenhq: `ctx.query.code`? 🙈

palmenhq: haha too eager

palmenhq:
idk what typescript is doing but aparrently it needs some help to understand what we're trying to do. Lets fix that before we merge ofc

PR #157, opened 2019-11-14
New offer screen base

gustafguner:
- Perils section with information modal (the information of each peril is still WIP, Zac told me the design wasn't finished)
- Compare table

All data is mocked, but shouldn't be too hard to integrate the real data later.

Perils
Screenshot 2019-11-14 at 13 04 30

Peril modal
Screenshot 2019-11-14 at 13 04 39

Compare table
Screenshot 2019-11-14 at 13 04 19


src/components/ModalNew.tsx
```diff
@@ -0,0 +1,166 @@
...
+
+  svg {
+    width: 100%;
+    height: 100%;
+    fill: ${colorsV2.white};
+  }
+`
+
+export const Modal = (props: React.PropsWithChildren<ModalProps>) => {
+  const containerRef = React.useRef(null)
```
palmenhq: We should probably pass the type to `useRef` so that this becomes `React.useRef<HTMLDivElement>(null)`  - that way we don't need to disable ts in some places i think

src/components/ModalNew.tsx
```diff
@@ -0,0 +1,166 @@
...
+  width: 100%;
+  height: 100%;
+  left: 0;
+  top: 0;
+  bottom: 0;
+  right: 0;
+  background-color: ${hexToRgba(colorsV2.white, 0.75)};
+`
+
+const Container = styled(motion.div)`
```
palmenhq: We should probably name this `wrapper` just to not confuse it with  a state container :D also to be consistent with the rest of the code base

src/components/ModalNew.tsx
```diff
@@ -0,0 +1,166 @@
...
+              },
+            },
+            hidden: {
+              opacity: 0,
+              transform: 'translateX(-50%) translateY(50%) scale(0.9)',
+            },
+          }}
+        >
+          {props.children}
+          <CloseButton onClick={() => props.onClose()}>
```
palmenhq: `onClick={props.onClose}`? :)

src/components/ModalNew.tsx
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  svg {
+    width: 100%;
+    height: 100%;
+    fill: ${colorsV2.white};
+  }
+`
+
+export const Modal = (props: React.PropsWithChildren<ModalProps>) => {
```
palmenhq: I think its better (for consistency and react "best practice") to make this `const Modal: React.FunctionComponent<ModalProps> = (props) => { ...` instead

src/components/icons/DownArrow.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react'
+
+export const DownArrow = () => (
```
palmenhq: i think we should clean these svgs - protip: https://jakearchibald.github.io/svgomg/ is awesome

src/pages/OfferNew/Compare/CompareTable.tsx
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  :hover {
+    color: ${colorsV2.violet500};
+  }
+`
+
+const getProperty = (key: string, value: any): any => {
+  if (key === 'deductible') {
+    return <ColumnRowPrimaryContent>{value} kr</ColumnRowPrimaryContent>
```
palmenhq: could we utilize the `Intl` api instead? like so https://github.com/HedvigInsurance/hedgehog/blob/master/src/lib/intl.ts#L7-L10

src/pages/OfferNew/Compare/CompareTable.tsx
```diff
@@ -0,0 +1,314 @@
...
+      <InsurancePropertiesSection>
+        <ColumnHead>
+          <SubHeadingBlack>Skydd</SubHeadingBlack>
+        </ColumnHead>
+
+        <InsurancePropertyNames>
+          {Object.entries(props.insuranceProperties)
+            .filter(([key]) => key !== 'name')
+            .map(([_, property]) => (
+              <InsuranceProperty>
```
palmenhq: `key` prop plz :)

src/pages/OfferNew/Compare/CompareTable.tsx
```diff
@@ -0,0 +1,314 @@
...
+
+      <PrimaryCompanySection>
+        <PrimaryCompanyHead>
+          <HedvigSymbol />
+        </PrimaryCompanyHead>
+
+        {Object.entries(props.primaryCompany)
+          .filter(([key]) => key !== 'name')
+          .map(([key, property]) => (
+            <CompanyColumnRow>{getProperty(key, property)}</CompanyColumnRow>
```
palmenhq: same, `key` prop

src/pages/OfferNew/Perils/PerilModal.tsx
```diff
@@ -0,0 +1,262 @@
...
+  position: relative;
+  height: 88px;
+  display: flex;
+  ${(props) =>
+    props.transition && `transition: all ${TRANSITION_MS}ms ease-in-out;`}
+
+  ${(props) =>
+    `transform: translateX(${
+      props.currentPeril !== 0
+        ? `calc(-33.3333333333333% - ${(props.currentPeril - 12 - 1) *
```
palmenhq: we could also do `100%/3` i _think_

src/pages/OfferNew/Perils/PerilModal.tsx
```diff
@@ -0,0 +1,262 @@
...
+      transition: all 0.15s ease-in-out;
+      fill: ${colorsV2.gray};
+    }
+  }
+  :focus {
+    outline: none;
+  }
+  &:hover {
+    svg {
+      * {
```
palmenhq: we shouldn't need this

src/pages/OfferNew/Perils/PerilModal.tsx
```diff
@@ -0,0 +1,262 @@
...
+
+const createPerilItems = (
+  perils: Peril[],
+  setCurrentPeril: (index: number) => void,
+  actionsAllowed: boolean,
+) => {
+  const items = []
+  for (let i = 0; i < perils.length * 3; i++) {
+    const index = i % perils.length
+    items.push(
```
palmenhq: i don't like loops 😬 did we at least _try_ to do this with map/reduce?

src/pages/OfferNew/Perils/index.tsx
```diff
@@ -0,0 +1,153 @@
...
+              {'Säkerhet genom livets alla $%*!;€&-stunder'}
+            </HeadingBlack>
+            <Body>
+              Omfattande skydd för dig och din familj, ditt hus och dina prylar.
+            </Body>
+          </HeadingWrapper>
+          <PerilItemCollection>
+            {perils.map((peril, perilIndex) => (
+              <PerilItem
+                title={peril.title}
```
palmenhq: `key` prop :)

src/components/ModalNew.tsx
gustafguner: Yeah, in the other context we don't use `constate` so that wasn't a problem there 😄

src/components/ModalNew.tsx
```diff
@@ -0,0 +1,166 @@
...
+  position: absolute;
+  left: 50%;
+  top: 50%;
+  transform: translateX(-50%) translateY(-50%);
+  box-shadow: 0 0 14px rgba(0, 0, 0, 0.06);
+  box-sizing: border-box;
+  overflow-x: scroll;
+
+  @media (max-height: 900px) {
+    max-height: calc(100vh - 32px);
```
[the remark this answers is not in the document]
gustafguner: Agree, this came straight from another context where we didn't use `rem` so that's why.

src/components/icons/DownArrow.tsx
gustafguner: Nice tool!

src/pages/OfferNew/Perils/PerilModal.tsx
gustafguner: Reformatted the svg so that the fill property is set on the `svg` instead of on the `path`, nice catch!

src/pages/OfferNew/Perils/PerilModal.tsx
gustafguner: Managed to fix it now with a map, I don't like them either but I think it's okay in some occasions when it makes the logic easier to understand!

src/pages/OfferNew/Compare/CompareTable.tsx
gustafguner: Yeah maybe we could, but I think it's better to postpone this to the point when we get real data

PR #158, opened 2019-11-15
Track Studentkortet

gustafguner:


src/pages/Sign/sections/SignSubscription.tsx
```diff
@@ -301,6 +308,12 @@ export const SubscriptionComponent: React.SFC<SubscriptionComponentProps> = ({
                       currentOffer.insurance.type,
                     )
                   }}
+                  runStudentkortet={() =>
```
sampettersson: Maybe we should figure out if they are from studentkortet first? based on utm-tags? so that we don't load scripts for all users.

gustafguner: Oh, lol my bad 😆Now checking _hvcode

PR #160, opened 2019-11-20
Embark MVP

sampettersson:
Todos for later:
- Load data from GraphQL
- Types for data from angel
- Use some kind of browser history implementation
- Decide which file to load from angel


hedvigoscar:
Noice!

PR #169, opened 2019-12-02
External Redirect: Concrete Implementation

hedvigoscar:
Bit hard to test currently unfortunately, but we'll get there soon enough!

### DIFF
--- a/src/pages/Embark/index.tsx
+++ b/src/pages/Embark/index.tsx
@@ -152,6 +152,7 @@ interface EmbarkRootProps {
 }
 
 export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
+  const history = useHistory()
   const [data, setData] = React.useState<null | any>(null)
   const [initialStore, setInitialStore] = React.useState<null | {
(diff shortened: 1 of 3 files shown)


src/pages/Embark/index.tsx
```diff
@@ -196,6 +197,11 @@ export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
     <StorageContainer>
       {(storageState) => (
         <EmbarkProvider
+          externalRedirects={{
+            Offer: () => {
+              history.push('/new-member/offer')
```
sampettersson: use a constant for this maybe, so that we don't have to change it in two places :)

palmenhq: just checking does this work with `/en/new-member` also?

hedvigoscar: I don't know 😞 

I'll make sure to try it, there are some things I need to do before I can check this one off.

I'll mark it as do not merge!

hedvigoscar:
Mostly obsolete at this point

PR #171, opened 2019-12-03
Pre-chat

gustafguner:
A pre-chat page with links to the different onboarding chats

Screenshot 2019-12-03 at 14 21 29
Screenshot 2019-12-03 at 14 25 30


src/new-components/TopBar.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { colorsV2 } from '@hedviginsurance/brand'
+import { HedvigLogo } from 'components/icons/HedvigLogo'
+import * as React from 'react'
+import { CurrentLanguage } from '../components/utils/CurrentLanguage'
+
+interface Props {
+  transparent?: boolean
+}
+
+const Wrapper = styled.div<{ transparent?: boolean }>`
```
hedvigoscar: Lets extract this type 🙂

gustafguner: Yeah sure, I used the same `Props` interface instead 😸

PR #176, opened 2019-12-04
Add text keys

gustafguner:
Replaced a lot of hard coded copy with text keys. I left some stuff out, such as perils and insurance values as we haven't looked at the data source there yet. Oh, and I also added "remove discount code"-functionality to the sidebar and improved some other stuff.


src/new-components/buttons.tsx
```diff
@@ -60,11 +60,11 @@ LinkButton.defaultProps = defaultButtonProps
 export const PureLinkButton = Button.withComponent('a')
 PureLinkButton.defaultProps = defaultButtonProps
 
-export const TextButton = styled.button`
+export const TextButton = styled.button<{ color?: string }>`
```
hedvigoscar: Should we allow any color here or something like 'just one of the entries in `colorsv2`?

And also can we extract it? 😉

src/pages/OfferNew/Compare/CompareTable.tsx
```diff
@@ -310,7 +311,12 @@ export const CompareTable = (props: Props) => {
     <Container>
       <InsurancePropertiesSection>
         <ColumnHead>
-          <SubHeadingBlack>Skydd</SubHeadingBlack>
+          <SubHeadingBlack>
+            <TranslationsConsumer textKey="COMPARE_TABLE_TITLE">
```
hedvigoscar: Not necessarily something we need to do in the context of this PR but just a thought:

It would be super-nice if we could convert this thing from a render props-based API to a hooks based API 🙄

gustafguner: Yeah, I know right! Something I've been thinking about too, but I guess that's an issue for later 🙄

hedvigoscar: Yeah! Let's let it stew for a bit in the back of our heads and then do it later

src/new-components/buttons.tsx
gustafguner: I extracted it, but since the types of `colorsV2` are simply for example `violet700: string`, trying to constrain the color type like `typeof colorsV2[keyof typeof colorsV2]` simply becomes `string` so that's no help. 😞

hedvigoscar: Stupid typescript 😞

PR #197, opened 2019-12-10
Chat out of beta

hedvigoscar:


src/routes.tsx
```diff
@@ -70,7 +64,7 @@ export const reactPageRoutes = [
     exact: true,
   },
   {
-    path: '/beta/new-member/:name?/:id?',
+    path: '/new-member/:name?/:id?',
     render: ({ match }: RouteComponentProps<any>) => {
       const getProps = () => {
         switch (match.params.name) {
```
palmenhq: shall w eupdate these also? or no?

PR #205, opened 2019-12-12
Quote query, graphql-codegen, useTextKeys hook

gustafguner:
- Generating Apollo hooks with `graphql-codegen` and replaced old component based with hooks
- Querying `Quote` in new offer page
- Implemented a `useTextKeys` hook (props to @hedvigoscar for the help) with support for replacements

Using the `useTextKeys` hook is very straight forward and saves us a lot of component wrapping.

```javascript

<PreHeading>{textKeys.HERO_LABEL}</PreHeading>

<HeadingWhite>
  {textKeys.HERO_HEADLINE.fill({
    FIRST_NAME: offer.member.firstName,
  })}
</HeadingWhite>
```


src/utils/hooks/useTextKeys.tsx
```diff
@@ -0,0 +1,36 @@
+import { TranslationsContext } from '@hedviginsurance/textkeyfy'
+import * as React from 'react'
+
+const placeholderRegex = new RegExp('({[a-zA-Z0-9_]+})', 'g')
+const placeholderKeyRegex = new RegExp('([a-zA-Z0-9_]+)', 'g')
+
+String.prototype.fill = function(replacements) {
```
sampettersson: I didn't really understand what `fill` did at first, maybe call this `.withReplacements` or something? 😄

src/server/page.tsx
```diff
@@ -121,14 +125,16 @@ export const getPage: Koa.Middleware = async (ctx) => {
...
         <ApolloProvider client={apolloClient}>
-          <MobileContext.Provider
-            value={isMobile({
-              ua: ctx.req.headers['user-agent'],
-              tablet: true,
-            })}
-          >
-            <App session={session} dontPanicSession={dontPanicSession} />
-          </MobileContext.Provider>
+          <LegacyApolloProvider client={apolloClient}>
```
sampettersson: 👵

src/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
```diff
@@ -54,26 +54,21 @@ const TooltipWrapper = styled.div`
...
-    <Value>
-      {insuredAtOtherCompany ? 'När din gamla löper ut' : 'Idag'}
-      <TooltipWrapper>
-        <Tooltip size="lg" body="Info" />
-      </TooltipWrapper>
-    </Value>
-  </Wrapper>
-)
+      <Value>
+        {insuredAtOtherCompany ? 'När din gamla löper ut' : 'Idag'}
```
sampettersson: missing text keys 👀

src/utils/hooks/useTextKeys.tsx
palmenhq: This is awesome!!

__HOWEVER__ i think we shouldn't "bloat" the global string object (`fill` is quite generic and it could easily get overridden by something else, for example). I think we might as well do this for each text key we return or? I think something like 

```ts

function fill(replacements: Record<string, any>) { ... }

// ...

 return new Proxy(context.textKeys, {
      get: (textKeys, key: string) => {
        const actualKey = textKeys[key] || key
        actualKey.fill = fill
        return actualKey
      },
    })
```

src/pages/Embark/index.tsx
```diff
@@ -248,6 +248,13 @@ export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
                     externalRedirects={{
                       Offer: () => {
                         history.push('/new-member/offer')
+                        /*
+                        history.push(
+                          `/beta/new-member/offer/${
+                            storageState.session.getSession()!.quoteId
+                          }`,
+                        )
+                        */
```
palmenhq: 🙈

src/server/page.tsx
gustafguner: Some problem with using `react-apollo` and the `@apollo/` packages at the same time 🙄 https://github.com/apollographql/react-apollo/issues/3381#issuecomment-550330189

src/utils/hooks/useTextKeys.tsx
gustafguner: I agree 👍

src/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
gustafguner: ⚠️

src/utils/hooks/useTextKeys.tsx
hedvigoscar: I think TypeScript might hate us if we do this (string.fill is not a function or something like that)

I agree that not filling up the string prototype would be nice. We could perhaps return a function from the proxy so that you can do

```ts

textKeys.SOME_KEY_WITH_REPLACEMENTS({REPLACEMENT: 'VALUE'})
textKeys.SOME_REGULAR_KEY()
```

hedvigoscar: I tried to wing an implementation of this but it didnt turn out good enough so idk

hedvigoscar: Here is that implementation for reference @sampettersson @palmenhq 

```ts
import { Replacements, TranslationsContext } from '@hedviginsurance/textkeyfy'
const placeholderRegex = new RegExp('({[a-zA-Z0-9_]+})', 'g')
const placeholderKeyRegex = new RegExp('([a-zA-Z0-9_]+)', 'g')
export const useTextKeys = () => {
  const context = React.useContext(TranslationsContext)
  return React.useMemo(
    () =>
      new Proxy<{ [key: string]: () => string }>(
        {},
        {
          get: (target, key: string) => (
            replacements?: Replacements,
          )string => {
            const possiblyUnresolvedKey = context.textKeys[key] || key
            if (!replacements) {
              return possiblyUnresolvedKey
            }
            const matches = possiblyUnresolvedKey
              .split(placeholderRegex)
              .filter((value) => value)
            return matches
              .map((placeholder, _) => {
                if (!placeholderKeyRegex.test(placeholder)) {
                  return placeholder
                }
                const replacementKey = placeholder.match(
                  placeholderKeyRegex,
                )![0]
                const value = replacements[replacementKey]
                if (value) {
                  return value.toString()
                }
                return placeholder
              })
              .reduce((acc, curr) => acc.concat(curr), '')
          },
        },
      ),
    [context.textKeys],
  )
}
```

gustafguner: I tried this again, but typescript is puking on me... I will change the function name to `withReplacements` as @sampettersson suggested, making the name more unique and less likely to be overwritten

hedvigoscar: TypeScript actually puked on me with the current implementation: `Property 'fill' does not exist on type 'String'.ts(2339)`

gustafguner: Yep, same thing for me. Let's go with `withReplacements` for now, and if someone solves this, they are more than welcome to update the hook logic 😄

palmenhq: Can't we make a composite return type?

```
 new Proxy<{ [key: string]: (string & { fill: (..) => string }) }>(...)
```

or so?

hedvigoscar: If typescript accepts this it would be nice!

hedvigoscar: We messed with it a little bit and came back to the version with keys being functions that take an optional parameter.
Slightly less fun to use, but we avoid messing with the string prototype chain and typescript actually liked it this time which is nice

PR #211, opened 2019-12-12
Discount codes

gustafguner:
Now supporting all three `Incentive` types!

### DIFF
--- a/src/graphql/Offer.graphql
+++ b/src/graphql/Offer.graphql
@@ -48,9 +48,6 @@ query Offer($id: ID) {
           currency
         }
       }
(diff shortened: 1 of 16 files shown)


src/graphql/Offer.graphql
```diff
@@ -48,6 +48,9 @@ query Offer($id: ID) {
           currency
         }
       }
+      ... on NoDiscount {
```
hedvigoscar: I think it's sufficient to check `__typename`? 🙂

gustafguner: You're right, I just removed this now 👍

PR #213, opened 2019-12-13
Checkout UI

palmenhq:
Still no real functionality in here, but lets get one thing at a time, this is massive enough as-is i think.


palmenhq:
@sampettersson cool, thanks! Yes i think lets poke in these details if we have time left

PR #224, opened 2019-12-18
Offer details modal w. editQuote

gustafguner:
- Details modal
- Edit quote functionality

Apartment
Screenshot 2019-12-18 at 13 33 29

House
Screenshot 2019-12-18 at 13 34 27


src/new-components/inputs/index.tsx
```diff
@@ -0,0 +1,266 @@
...
+export const masks: { [key: string]: Mask } = {
+  zipCode: {
+    name: 'ZipCode',
+    mask: '999 99',
+    sanitize: (value) => value.replace(/\s+/, ''),
+  },
+  area: {
+    name: 'Area',
+    mask: '99999 m2',
+    sanitize: (value) => value.replace(/[\s\u200b]+m2/, ''),
```
palmenhq: what is `\u200b`? if you mean `²` that's technically `\u00b2` no?

src/new-components/inputs/index.tsx
```diff
@@ -0,0 +1,266 @@
...
+      <TextWrapper>
+        <Label>{label}</Label>
+        {mask ? (
+          <StyledField {...props}>
+            {({ field }: any) => (
+              <StyledInput
+                placeholder={props.placeholder}
+                mask={mask.mask}
+                alwaysShowMask={true}
+                maskChar={`\u200b`}
```
palmenhq: again what is `\u200b`?

src/new-components/inputs/index.tsx
```diff
@@ -0,0 +1,266 @@
...
+  color: ${colorsV2.coral700};
+  margin-top: 0.25rem;
+`
+
+interface CoreInputFieldOptions {
+  label: string
+  value: string
+}
+
+export const inputTypes: { [key: string]: string } = {
```
palmenhq:
```ts
const inputTypes: Record<string, string> = ...
```
maybe?

src/new-components/inputs/index.tsx
```diff
@@ -0,0 +1,266 @@
...
+}
+
+export interface TextInputProps extends CoreInputFieldProps {
+  showErrorMessage?: boolean
+  touched?: boolean
+  errors?: string
+}
+
+export const InputField: React.FC<TextInputProps &
+  GenericFieldHTMLAttributes> = ({
```
palmenhq: is this really how prettier likes it? 🤨

src/new-components/inputs/index.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+export const InputGroup = styled.div`
+  margin-bottom: 1.5rem;
+
+  > ${Wrapper}:first-child {
+    border-top-left-radius: 8px;
+    border-top-right-radius 8px;
+  }
+
+  > ${Wrapper}:last-child, ${InputGroupDeleteButton}:last-child {
```
palmenhq: i had no idea you could do this, amazing!

src/pages/Embark/index.tsx
```diff
@@ -259,7 +259,12 @@ export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
                   <EmbarkProvider
                     externalRedirects={{
                       Offer: () => {
-                        history.push('/new-member/offer')
+                        // history.push('/new-member/offer')
+                        history.push(
+                          `/beta/new-member/offer/${
+                            storageState.session.getSession()!.quoteId
+                          }`,
+                        )
```
palmenhq: oops

src/pages/OfferNew/Introduction/Sidebar/DetailsModal/DetailInput.tsx
```diff
@@ -0,0 +1,51 @@
...
+        const value = field.mask
+          ? field.mask.sanitize(e.target.value)
+          : e.target.value
+
+        if (field.type === 'number') {
+          formikProps.setFieldValue(formikName, parseInt(value, 10))
+        } else {
+          formikProps.setFieldValue(
+            formikName,
+            /^(true|false)$/.test(value) ? JSON.parse(value) : value,
```
palmenhq: maybe we should use `\A(true|false)\z` instead? it's a bit safer considering `^...$` matches any line whereas `\A...\z` matches start-end of the full string

src/pages/OfferNew/Introduction/Sidebar/DetailsModal/index.tsx
```diff
@@ -0,0 +1,425 @@
...
+const LoadingDimmer = styled.div<{ visible: boolean }>`
+  background: rgba(255, 255, 255, 0.5);
+  position: absolute;
+  top: 0;
+  left: 0;
+  right: 0;
+  bottom: 0;
+  transition: all 350ms;
+  opacity: ${(props) => (props.visible ? 1 : 0)};
+  visibility: ${(props) => (props.visible ? 'visible' : 'hidden')};
```
palmenhq: does `opacity` have _any_ effect here? considering we're using `visible`/`hidden` also?

src/pages/OfferNew/Introduction/Sidebar/DetailsModal/index.tsx
```diff
@@ -0,0 +1,425 @@
...
+                  return
+                }
+
+                await refetch()
+                onClose()
+                setIsUpdating(false)
+              })
+              .catch(() => {
+                setIsUpdating(false)
+              })
```
palmenhq: i think we should pick either async/await or `.then` and not mix them, imo that becomes a little messy.

src/pages/OfferNew/Introduction/Sidebar/DetailsModal/index.tsx
```diff
@@ -0,0 +1,425 @@
...
+                                  arrayHelpers.insert(0, defaultExtraBuilding)
+                                }}
+                              >
+                                {textKeys.DETAILS_MODULE_EXTRABUILDINGS_TABLE_BUTTON()}
+                              </ContentColumnTitleButton>
+                            </ContentColumnTitle>
+
+                            {formikProps.values.house?.extraBuildings?.map(
+                              (_, index) => (
+                                <InputGroup key={index}>
```
palmenhq: i believe this (`key=index`) might cause issues if you remove a house that isn't the last, considering React sees the `key` as the "definition" of which DOM element this is and the id in the middle remains the same. I guess since we're updating the name though (ie `extraBuildings.${index}.type`) this might not be an issue but it'll most likely cause strange and hard-to-track-down bugs in the future if we come back here. Maybe we could find a better `key` prop? just like a hash, id, anything except the array index

src/pages/OfferNew/Introduction/Sidebar/DetailsModal/index.tsx
```diff
@@ -0,0 +1,425 @@
...
+                                    nameRoot="house"
+                                    name={`extraBuildings.${index}.hasWaterConnected`}
+                                  />
+
+                                  <InputGroupDeleteButton
+                                    type="button"
+                                    onClick={() => {
+                                      const isLastItem =
+                                        formikProps.values.house?.extraBuildings
+                                          ?.length === 1
```
palmenhq: i think i don't really understand this logic, what if you have 3 extra buildings? then `isLastItem` doesn't make sense for `extraBulidings.length === 1` does it? Or do you mean that this is the only item in the list? Then i think we should specify that as a name instead

src/pages/OfferNew/Introduction/Sidebar/DetailsModal/types.ts
```diff
@@ -0,0 +1,37 @@
...
+  house: FieldSchemaBuilder<RequiredFields<EditHouseInput>>
+}
+
+export type FieldSchemaBuilder<T> = {
+  [P in keyof T]: T[P] extends string | number | boolean
+    ? RegularFieldType
+    : T[P] extends any[]
+    ? ArrayFieldType<T[P]>
+    : FieldSchemaBuilder<T[P]>
+}
```
palmenhq: 🤯

src/pages/OfferNew/Introduction/Sidebar/DetailsModal/utils.ts
```diff
@@ -0,0 +1,327 @@
...
+      label: 'DETAILS_MODULE_TABLE_ADDRESS_CELL_LABEL',
+      placeholder: 'DETAILS_MODULE_TABLE_ADDRESS_CELL_LABEL',
+      validation: Yup.string().required(),
+    },
+    zipCode: {
+      label: 'DETAILS_MODULE_TABLE_POSTALCODE_CELL_LABEL',
+      placeholder: 'DETAILS_MODULE_TABLE_POSTALCODE_CELL_LABEL',
+      mask: masks.zipCode,
+      type: inputTypes.string,
+      validation: Yup.string().matches(/^[0-9]{3}[0-9]{2}$/),
```
palmenhq: `\A...\z`? :)

src/new-components/inputs/index.tsx
gustafguner: `\u200b` the unicode character "Zero width space" 😎 https://www.fileformat.info/info/unicode/char/200b/index.htm

It was a hacky thing I did in order to get the masking to work... Otherwise, the `m2` was only added when you had filled it in if I remember correctly

src/new-components/inputs/index.tsx
gustafguner: See comment above 😄

src/new-components/inputs/index.tsx
gustafguner: Yes, much better!

src/new-components/inputs/index.tsx
gustafguner: Apparently 😅

src/pages/Embark/index.tsx
gustafguner: 🙈

src/pages/OfferNew/Introduction/Sidebar/DetailsModal/index.tsx
gustafguner: Very true, nice input

src/new-components/inputs/index.tsx
gustafguner: React-input-mask uses a `_` as mask char, so in order to remove it, but still keeping the `m2` I used an invisible character

src/pages/OfferNew/Introduction/Sidebar/DetailsModal/utils.ts
gustafguner: 👍

src/new-components/inputs/index.tsx
palmenhq: Really clever, i will just put it as a named constant so we remember this in the future

src/pages/OfferNew/Introduction/Sidebar/DetailsModal/index.tsx
palmenhq: honestly, this sucks but fixing it will suck even more 😂 i'll leave it for now

PR #246, opened 2020-01-13
Styling fixes for switcher onboarding

sampettersson:


src/pages/OfferNew/Introduction/ExternalInsuranceProvider/index.tsx
```diff
@@ -18,6 +18,12 @@ const ErrorBox = styled.div`
   padding: 1.25rem;
   background-color: white;
   border-radius: 0.5rem;
+
+  @media (max-width: 800px) {
+    padding: 1.25rem;
```
palmenhq: why do we apply the same padding again?

sampettersson: haha that was not intentional

PR #248, opened 2020-01-20
Show discount amount in percentage discount + don't remove cookie

palmenhq:


src/pages/OfferNew/common/price.tsx
```diff
@@ -105,7 +105,7 @@ export const Price: React.FC<{
         aria-hidden={!monthlyCostDeduction}
       >
         {textKeys.SIDEBAR_OLD_PRICE({
-          PRICE: Number(monthlyGross.amount),
+          PRICE: Math.round(Number(monthlyGross.amount)),
```
hedvigoscar: Are we getting non-rounded numbers from our backend? This seems a bit dangerous 🤔

palmenhq: Yes, in general i agree, but i spoke to some people in platform and we decided to go with this for now, since rounding half-even is always rounding half-even

PR #255, opened 2020-01-22
Allow choosing start date to be when previous insurance expires

hedvigoscar:
There's one outstanding issue in this PR: How will the backend respond if we send null on editQuote for startDate? Will it ignore it, or will it explicitly set the data to be null?

EDIT: I've updated this code to use `removeStartDate` 🙂

### DIFF
--- a/src/new-components/DateInput/index.tsx
+++ b/src/new-components/DateInput/index.tsx
@@ -8,6 +8,7 @@ import { addYears, subDays } from 'date-fns'
 import Dayzed, { RenderProps as DayzedCalendarProps } from 'dayzed'
 import { motion } from 'framer-motion'
 import * as React from 'react'
+import { TextButton } from '../buttons'
 import { AnimationDirection, Animator } from './Animator'
 import { useMeasure } from './useMeasure'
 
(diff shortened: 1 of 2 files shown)


palmenhq:
I think rather you should use the `removeStartDate` mutation @hedvigoscar

src/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
```diff
@@ -236,10 +236,11 @@ export const StartDate: React.FC<Props> = ({
           setStartDate({
             variables: {
               quoteId: offerId,
-              date: format(newDateValue, gqlDateFormat),
+              date: newDateValue ? format(newDateValue, gqlDateFormat) : null,
```
palmenhq: this'll most likely not work, we should use `removeStartDate` instead

hedvigoscar:
Fantastic! That answers a big question I had, thank you 😄

src/new-components/DateInput/index.tsx
```diff
@@ -345,6 +355,18 @@ export const DateInput: React.FC<DateInputProps> = ({
           }}
           render={(dayzedData) => <Calendar {...dayzedData} />}
         />
+        {hasCurrentInsurer ? (
```
palmenhq: `hasCurrentInsuerer && (...)` *

hedvigoscar:
@palmenhq Ready for re-review 🙂

src/new-components/DateInput/index.tsx
```diff
@@ -344,6 +354,18 @@ export const DateInput: React.FC<DateInputProps> = ({
...
         />
+        {hasCurrentInsurer && (
+          <AtStartDateContainer>
+            <TextButton
+              onClick={() => {
+                setDate(null)
+                setOpen(false)
+              }}
+            >
+              När min gamla försäkring löper ut
```
gustafguner: Text key police 🚓

hedvigoscar: 🕵

PR #256, opened 2020-01-22
Add favicons

hedvigoscar:
Note that this will only work when proxied through web-next.

That's fine though, right?


hedvigoscar:
I guess we could 🤔

PR #260, opened 2020-02-13
Fix moderna id

sampettersson:


src/pages/OfferNew/Compare/mock.ts
```diff
@@ -70,7 +70,7 @@ export const otherCompanies: CompanyProperties[] = [
   },
   {
     name: 'Moderna Försäkringar Grundskydd',
-    id: 'Moderna Försäkringar',
+    id: 'Moderna',
```
palmenhq: Ouff this feels dangerous. Has this casued any issues?

sampettersson: Yeah, we couldn't get the trustpilot score for moderna.

PR #265, opened 2020-03-03
Add new GraphQL resolvers

sampettersson:


src/pages/Embark/graphql.ts
```diff
@@ -0,0 +1,58 @@
...
+) => {
+  if (!apolloClient) {
+    throw new Error('Missing apollo client')
+  }
+
+  if (!storage.session.getSession().token) {
+    const sessionResult = await apolloClient.client.mutate({
+      mutation: CREATE_SESSION_TOKEN_MUTATION,
+    })
+    await afterTick(() => {
```
palmenhq: why not just something like
```js
await tick()
apolloClient!.subscriptionClient.close(true, true)
storage.setToken(sessionResult.data.createSessionV2.token)
```

PR #268, opened 2020-03-16
Send embark store changed events to segment

sampettersson:
this is a POC solution for embark tracking, sending the store each time it changes and trying out to see if we can make sense of this in mixpanel.

### DIFF
--- a/src/pages/Embark/index.tsx
+++ b/src/pages/Embark/index.tsx
@@ -297,6 +297,14 @@ export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
                     }}
                     initialStore={initialStore}
                     onStoreChange={(store) => {
+                      const castedWindow = window as any
+                      if (castedWindow && castedWindow.analytics) {
+                        castedWindow.analytics.track(
+                          `Embark Store Changed - ${props.name!}`,
+                          store,
+                        )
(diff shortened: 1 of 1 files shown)


src/pages/Embark/index.tsx
```diff
@@ -297,6 +297,14 @@ export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
...
                     initialStore={initialStore}
                     onStoreChange={(store) => {
+                      const castedWindow = window as any
+                      if (castedWindow && castedWindow.analytics) {
+                        castedWindow.analytics.track(
+                          `Embark Store Changed - ${props.name!}`,
+                          store,
+                        )
+                      }
+
```
palmenhq: i think we should do this after actually setting the data, in case this explodes for some reason

sampettersson:
Actually, this is IMO the preferred way to do things, the store is the source of truth of what action the user has taken and always has all the data regardless of what component has been used. This ease the implementation overhead on apps later on and makes the specification easier. 

We should solve the rest of the tracking with specific events where needed and not make specific tracking were not needed.

sampettersson:
Realised that this might not be needed, see: https://github.com/HedvigInsurance/embark/pull/198

closing for now 💃

PR #270, opened 2020-03-20
Norway offer page prep

palmenhq:
There's still plenty to do before we have anything e2e, but this is a first step to handling norwegian quotes without 💥 I would like to merge this, unless there's any issues with it, so we don't get a norway PR that grows and grows and grows :)


src/new-components/TopBar.tsx
```diff
@@ -15,7 +15,7 @@ const Wrapper = styled.div<Props>`
   background: ${(props) =>
     props.transparent ? `transparent` : colorsV2.white};
   position: absolute;
-  z-index: 1010;
+  z-index: 1009;
```
gustaveen: That's a very specific number :D 
A question, are we switching to using template literals for styling? I'm very pro that if we are :)

gustaveen (dismissed):
Good stuff! 🙌

src/new-components/TopBar.tsx
palmenhq: Haha yes, maybe we should name this number, now that you mention it 😬 its because the overlay of the checkout is `1010`

Honestly we just haven't been very consistent haha. I guess they have their perks and cons both of them, and maybe we should just decide? A theme for next week's brekke?

gustaveen: Sounds like a perfect topic! 🙌

gustaveen (approved):

PR #271, opened 2020-03-20
Norway/edit personnummer

palmenhq:
Add the personnummer in the checkout for norway, and you shouldn't be able to continue without filling it out (aka when the quote has it registered). There's some todos in here, but i think we'll iron them out in the future - i think they are non-critical for getting a sign flow up and running, but important enough to let us find them later.

Screenshot 2020-03-23 at 16 15 05

### DIFF
--- a/src/pages/OfferNew/Checkout/CheckoutContent.tsx
+++ b/src/pages/OfferNew/Checkout/CheckoutContent.tsx
@@ -1,6 +1,10 @@
 import styled from '@emotion/styled'
 import { colorsV2, fonts } from '@hedviginsurance/brand/dist'
-import { CompleteQuote, useRedeemedCampaignsQuery } from 'data/graphql'
+import {
+  CompleteQuote,
+  useEditQuoteMutation,
+  useRedeemedCampaignsQuery,
+} from 'data/graphql'
 import * as React from 'react'
 import { useTextKeys } from 'utils/hooks/useTextKeys'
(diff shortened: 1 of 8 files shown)


src/pages/OfferNew/Checkout/CheckoutContent.tsx
```diff
@@ -103,7 +111,24 @@ export const CheckoutContent: React.FC<Props> = ({
...
-        <UserDetailsForm email={email} onEmailChange={onEmailChange} />
+        <UserDetailsForm
+          email={email}
+          onEmailChange={onEmailChange}
+          ssn={firstQuote.ssn ?? ''}
+          onSsnChange={(ssn) => {
+            setFakeLoading(true)
+            setReallyLoading(true)
+            window.setTimeout(() => setFakeLoading(false), 1000)
+            // TODO we somehow need to compare the birth date to the ssn to check so they match. In case they don't, we should warn (as they might get a different price)
```
gustaveen: So we use the birthdate to calculate price now right?

gustaveen (approved):
👍

src/pages/OfferNew/Checkout/CheckoutContent.tsx
palmenhq: Yes, apparently norwegians are really careful with giving out their fødelsenummer, so we're pushing that step as late as possible

PR #273, opened 2020-03-23
Add Norwegian landing page

vonElfvin:


src/pages/Embark/Landing.tsx
```diff
@@ -147,26 +149,61 @@ export const Landing: React.FC<{ language?: string }> = ({ language }) => {
...
+                  <Paragraph>{textKeys.STARTPAGE_INSURED_BODY()}</Paragraph>
+                  <ProceedButton
+                    size="lg"
+                    to={(language ? '/' + language : '') + '/new-member/switch'}
+                  >
+                    {textKeys.STARTPAGE_INSURED_BUTTON()}
+                  </ProceedButton>
+                </Card>
+              </>
+            )}
```
palmenhq: What about moving this stuff to separate components, like a `LandingContentSe` or so? Just to slim this one down a bit and separate the responsibilities.

palmenhq (commented):
Awesome! Just one question really. And welcome to the web team! You may join @HedvigInsurance/www also, if you wanna ;)

src/pages/Embark/Landing.tsx
vonElfvin: I tried that but sending "textKeys" was a nightmare, I have no idea what type it is and don't want to implicitly cast it to any. I'll try it again though.

palmenhq: I think you could put `const textKeys = useTextKeys()` inside these components also :)

src/pages/Embark/Landing.tsx
```diff
@@ -150,63 +150,74 @@ export const Landing: React.FC<{ language?: string }> = ({ language }) => {
...
             )}
           </Container>
         </Wrapper>
       </LandingPageContainer>
     </Page>
   )
 }
+
+const LandingPageCardsSe: React.FC<{
+  textKeys: Record<string, any>
```
palmenhq: I think we should use `TextKeyMap` from https://github.com/HedvigInsurance/web-onboarding/blob/master/src/utils/hooks/useTextKeys.ts - or just use `useTextKeys` inside these components instead.

palmenhq (approved):
💥🔥! Lets 🛳 it!

PR #274, opened 2020-03-24
Fetch token provided by app

sampettersson:


src/AppTokenRetrival.tsx
```diff
@@ -0,0 +1,24 @@
...
+export const AppTokenRetrival = ({ children }) => (
+  <StorageContainer>
+    {(storageState) => (
+      <Mount
+        on={() => {
+          if (location.hash.includes('token=')) {
+            const token = location.hash.replace('#token=', '')
+            console.log('setting token to', token)
+            apolloClient!.subscriptionClient.close(true, true)
+            storageState.setToken(token)
```
palmenhq: shouldn't this happen in the reverse order?

src/AppTokenRetrival.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { StorageContainer } from './utils/StorageContainer'
+
+export const AppTokenRetrival = ({ children }) => (
+  <StorageContainer>
+    {(storageState) => (
+      <Mount
+        on={() => {
+          if (location.hash.includes('token=')) {
+            const token = location.hash.replace('#token=', '')
+            console.log('setting token to', token)
```
palmenhq: 🙈

src/AppTokenRetrival.tsx
```diff
@@ -0,0 +1,24 @@
...
+import { apolloClient } from './client/apolloClient'
+import { StorageContainer } from './utils/StorageContainer'
+
+export const AppTokenRetrival = ({ children }) => (
+  <StorageContainer>
+    {(storageState) => (
+      <Mount
+        on={() => {
+          if (location.hash.includes('token=')) {
+            const token = location.hash.replace('#token=', '')
```
palmenhq: Could we do this with react router instead? Would make me feel better

src/AppTokenRetrival.tsx
sampettersson: hmm, we do in this order in other places 🤔

sampettersson: but it makes more sense to set the token first, and that worked, so doing that instead 👨‍🔬

src/AppTokenRetrival.tsx
```diff
@@ -0,0 +1,38 @@
...
+            if (location.hash.includes('token')) {
+              var params = {}
+
+              location.hash
+                .substring(1)
+                .split('&')
+                .map((hk) => {
+                  let param = hk.split('=')
+                  params[param[0]] = param[1]
+                })
```
palmenhq: This is a bit nit but using `Array#reduce` would be prettier ;) but i guess this is _acceptable_ although i think `forEach` would be more appropriate if we wouldn't want to use reduce for some reason

src/AppTokenRetrival.tsx
```diff
@@ -0,0 +1,38 @@
...
+export const AppTokenRetrival = ({ children }) => {
+  const location = useLocation()
+
+  return (
+    <StorageContainer>
+      {(storageState) => (
+        <Mount
+          on={() => {
+            if (location.hash.includes('token')) {
+              var params = {}
```
palmenhq: `let`*

palmenhq: But most preferrably `Array#reduce` + `const`

src/AppTokenRetrival.tsx
sampettersson: hahah was to quick copying from stackoverflow, this is now a reduce hehe

src/AppTokenRetrival.tsx
sampettersson: haha muskelminne, in Swift a let is immutable but and a var is mutable 😭

PR #277, opened 2020-03-26
Get perils from content service

gustaveen:


src/graphql/Quote.graphql
```diff
@@ -30,6 +30,20 @@ query Quote($id: ID!) {
           currency
         }
       }
+      perils(locale: sv_SE) {
```
gustaveen: Some idea on how we should we set/get locale in the Quote query? @palmenhq

palmenhq: I think the best option we have is to pick it from the url. I think we could use it as an input (`$perilsLocale`) to the query, and then pick the locale from [`useCurrentLocale`](https://github.com/HedvigInsurance/web-onboarding/blob/master/src/components/utils/CurrentLocale.tsx). In there you even have a `getLocaleIsoCode` which i _think_ you can use right off the bat

src/pages/OfferNew/Perils/index.tsx
```diff
@@ -16,30 +15,25 @@ import {
...
+  perils: ReadonlyArray<PerilV2>
 }
 
 const Wrapper = styled.div`
   padding: 8.5rem 0 5rem 0;
   background-color: ${colorsV2.offwhite};
   display: flex;
 `
+export const getIconUrl = (iconPath: string) =>
+  `https://graphql.dev.hedvigit.com${iconPath}` // TODO where should we get the base url from?
```
palmenhq: I think we use `window.GIRAFFE_ENDPOINT` right now, but I do think that includes `/graphql`. Could we do something with that maybe?

src/pages/OfferNew/Perils/types.ts
```diff
@@ -8,3 +8,25 @@ export interface Peril {
...
+}
+
+export interface PerilV2 {
+  title: React.ReactNode
+  description: string
+  covered: string[]
+  exceptions: string[]
+  info?: string
+  icon: PerilIcon
+}
```
palmenhq: Correct me if i'm wrong but aren't these types in the graphql schema already? If so we could use them directly from `data/graphql` after running `yarn gen`

src/pages/OfferNew/Perils/types.ts
```diff
@@ -8,3 +8,25 @@ export interface Peril {
...
+type IconVariants = {
+  dark: IconVariant
+  light: IconVariant
+}
+
+export type PerilIcon = {
+  variants: IconVariants
+}
+
+export interface PerilV2 {
```
palmenhq: do we need `V2`? I think we could kill off the old ones right? And just make this Peril type the _the_ Peril? Assuming we can't use the peril types from graphql, that is

src/pages/OfferNew/Perils/PerilModal.tsx
```diff
@@ -351,7 +352,7 @@ export const PerilModal: React.FC<PerilModalProps & ModalProps> = (props) => {
                   props.setCurrentPeril(index % tripledPerils.length)
                 }
               >
-                {peril.icon}
+                <img src={getIconUrl(peril.icon.variants.light.svgUrl)} />
```
palmenhq: I had this idea that maybe we could use ajax calls to fetch the icons' contents, for example if we wanted to `fill`/tint them. But maybe they're in the right color already?

palmenhq (commented):
Looks awesome! BUT i think we could kill a lot of code here :)

src/pages/OfferNew/Perils/types.ts
gustaveen: Ah, of course! 😄

src/graphql/Quote.graphql
gustaveen: Let's do it!

src/pages/OfferNew/Perils/index.tsx
gustaveen: Updated now but it feels kinda hacky :D

palmenhq: Very 😬 What about instead adding another environment variable?

gustaveen: Hell yeah

src/pages/OfferNew/Perils/PerilModal.tsx
gustaveen: They are black, and they should be in the slightly grayer variant. So that solution can be good to add but I'm thinking in a separate PR to just get the other parts in place or WDYT?

palmenhq: Good thinking 👍 i like taking on this type of debt more than other kinds of shortcuts - it's extremely visible and will be easy to detect in a qa session, meaing we'll _actually_ fix it "later"

palmenhq (dismissed):
Awesome! Lets just fix CI :)

palmenhq (dismissed):
Lgtm! Although shall we maybe kill the old perils while we're poking around here?

gustaveen:
> Lgtm! Although shall we maybe kill the old perils while we're poking around here?

A yes! Took out the trash now :)

palmenhq (dismissed):
What a lovely view 😍
Screenshot 2020-03-30 at 16 48 52

Although i realize since content-service isn't in production we need to hold the merge until that happens :/ Awesome work, but lets leave it like this until we have content-service out there :)

gustaveen:
Ah yes, always nice with some spring cleaning! 🧹Good catch, yes let's wait for that!

palmenhq (approved):
lets ship this puppy!

PR #284, opened 2020-04-02
Add adtractionn for norway

vonElfvin:


vonElfvin:
Obviously have to fix some tests here 😄

PR #292, opened 2020-04-07
Asset fallback proxy

palmenhq:
Now that we proxy dev/prod to different apps we somehow need to be able to get the built js/assets from both the apps. This is a solution, where we try first the stuff on `/new-member-assets` and then whatever we set the `ASSET_FALLBACK_PROXY` to (for example, `https://hvg-web-onboarding-no-staging.herokuapp.com`


vonElfvin (approved):
neat

PR #294, opened 2020-04-07
Await bankid completed event

palmenhq:


src/components/LoadingPage.tsx
```diff
@@ -0,0 +1,41 @@
...
+  text-align: center;
+`
+
+const ContentWrapper = styled('div')`
+  padding-top: 2rem;
+  font-size: 1rem;
+  color: ${colorsV3.white};
+`
+
+export const LoadingPage: React.FC = ({ children }) => {
```
vonElfvin, line 29: Smort to extract this

src/pages/OfferNew/Checkout/CheckoutContent.tsx
```diff
@@ -119,9 +119,9 @@ export const CheckoutContent: React.FC<Props> = ({
           onEmailChange={(email) => {
             const onCompletion = new Promise<void>((resolve, reject) => {
               Promise.all(
-                quoteIds.map((quoteId) => {
-                  editQuote({ variables: { input: { id: quoteId, email } } })
-                }),
+                quoteIds.map((quoteId) =>
+                  editQuote({ variables: { input: { id: quoteId, email } } }),
+                ),
```
vonElfvin, line 124: The battle of the tslints

vonElfvin (approved):
Damn Daniel

src/pages/OfferNew/Checkout/CheckoutContent.tsx
palmenhq: Actually, this is a bug fix :D since we didn't return the promise `editQuote` returns the `refetch` that should happen after happened at the same time as `editQuote` making members unable to sign unless they edit their ssn twice

PR #295, opened 2020-04-07
Use hasContract when polling for moving to payments

vonElfvin:


src/pages/OfferNew/Checkout/index.tsx
```diff
@@ -209,7 +206,7 @@ export const Checkout: React.FC<Props> = ({
       offerData.person.ssn,
   )
 
-  if (signStatus?.signState === SignState.Completed) {
+  if (hasContract) {
```
vonElfvin, line 209: We could also do this to make it super-safe
```suggestion
  if (signStatus?.signState === SignState.Completed && hasContract) {
```

vonElfvin: Where we check both the first and last step of the sign process. Right now creating the contract is last in the chain so to speak.

vonElfvin:
> Cool! Just a general question, do we handle the state where sign is completed but you haven't received any contract yet? I haven't tried this but judging from the code it'd probably stop the loading state, but not redirect, right?

Very true, I now updated the loading state to also look at `hasContract`

src/pages/OfferNew/Checkout/index.tsx
```diff
@@ -274,6 +275,7 @@ export const Checkout: React.FC<Props> = ({
             signQuotesMutation.loading ||
             signUiState === SignUiState.STARTED ||
             signUiState === SignUiState.STARTED_WITH_REDIRECT ||
+            !hasContract ||
```
palmenhq, line 278: won't this cause the loading state to always be true until you signed?

vonElfvin: Really good catch! Fixed now

vonElfvin:
This will be solved by the backend instead. A similar outcome with a backend solution.

PR #298, opened 2020-04-08
Fix so peril images shrink in chrome

palmenhq:
So this fix makes no sense, but it does fix this issue:

# BEFORE
Screenshot 2020-04-08 at 09 36 15

# AFTER
Screenshot 2020-04-08 at 09 36 06

i love css

(and yes, we need to shorten the titles a bit but i guess this is a bit less sucky) (and ping @gustaveen btw you might want to fix this in web-next also)


AlexSJ1990 (approved):
🚢

PR #299, opened 2020-04-08
Fix so intercom doesn't pop up during adyen payments

palmenhq:
Also fix some minor ui bugs

### DIFF
--- a/src/utils/Intercom.tsx
+++ b/src/utils/Intercom.tsx
@@ -2,6 +2,23 @@ import React from 'react'
 import { useRouteMatch } from 'react-router'
 import { useVariation, Variation } from 'utils/hooks/useVariation'
 
+const createIntercom = () => {
+  const script = `
+      window.intercomSettings = {
+        app_id: "ziqa7goa"
+      };
+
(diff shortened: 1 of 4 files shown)


src/utils/Intercom.tsx
```diff
@@ -0,0 +1,45 @@
...
+      isIntercomMatch &&
+      typeof (window as any).Intercom !== 'undefined'
+    ) {
+      ;(window as any).Intercom('boot')
+    } else if (
+      !isIntercomMatch &&
+      typeof (window as any).Intercom !== 'undefined'
+    ) {
+      ;(window as any).Intercom('shutdown')
+    }
```
gustaveen, line 41: Not sure what's going on here haha, but I trust you do

gustaveen (dismissed):

src/utils/Intercom.tsx
palmenhq: Basically turning it on if we should display it and off if we shouldn't maybe that should be clarified though?

gustaveen (approved):
Final final!

PR #308, opened 2020-04-16
Favicons and pivotify some sections in offer page

palmenhq:


src/pages/OfferNew/SwitchSafetySection.tsx
```diff
@@ -146,16 +148,17 @@ const zoomIn = keyframes`
...
   }
   justify-content: center;
   align-items: center;
   flex-shrink: 0;
   width: 4.5rem;
   height: 4.5rem;
-  background: #fff;
-  border-radius: 18px;
+  background: ${colorsV3.gray900};
+  color: ${colorsV3.white};
```
gustaveen, line 160: In market web we use gray100 instead of white in all places, I'm guessing it's the same for onboarding. But we can change that later 
```suggestion
  color: ${colorsV3.gray100};
```

gustaveen (approved):

PR #316, opened 2020-05-05
Possibly fix utm tracking and add some missing tracking events

palmenhq:


src/pages/Download/index.tsx
```diff
@@ -3,9 +3,12 @@ import { colorsV3 } from '@hedviginsurance/brand/dist'
 import { TranslationsConsumer } from '@hedviginsurance/textkeyfy'
 import { TopBar } from 'components/TopBar'
 import { Page } from 'components/utils/Page'
+import { SessionTokenGuard } from 'containers/SessionTokenGuard'
+import { SemanticEvents } from 'quepasa'
```
gustaveen, line 7: Is this a package for sending events to Segments?

gustaveen (approved):
🙌

src/pages/Download/index.tsx
palmenhq: Yes! I think its made by us but a long time ago

PR #317, opened 2020-05-07
Update adtraction tag for norway

palmenhq:


src/server/page.tsx
```diff
@@ -170,9 +177,14 @@ export const getPage: Koa.Middleware = async (ctx) => {
     return
   }
 
+  const ADRTRACTION_NO = 'https://cdn.adt387.com/jsTag?ap=1492109567'
+  const ADTRACTION_SE = 'https://adtr.io/jsTag?ap=1412531808'
```
gustaveen, line 181: Why are they so different depending on market?

gustaveen (approved):

src/server/page.tsx
palmenhq: 🤷‍♀️🤷‍♀️ No clue

PR #320, opened 2020-05-13
Stop cleaning sessions after 30 minutes

palmenhq:
Idk why we do this anymore and maybe it causes some strange issues for some users


gustaveen (approved):

palmenhq:
I think we shouldn't use this atm, lets close it and see if we need it in the future

PR #324, opened 2020-06-08
Connect payment direct-linking

palmenhq:
Also introduces [snapshot testing](https://jestjs.io/docs/en/snapshot-testing) (here's a [good article](https://tsh.io/blog/pros-and-cons-of-jest-snapshot-tests/) that provides some opinions). I'd like some input on this! Is snapshoting something we should do or nah?


src/utils/ExchangeTokenRetrieval.tsx
```diff
@@ -0,0 +1,137 @@
...
+    setTokenExchangeState(actualLoadingState)
+  }, [actualLoadingState, hasFinishedFakeLoading])
+
+  React.useEffect(() => {
+    if (hasFinishedFakeLoading) {
+      return
+    }
+
+    setTimeout(() => {
+      setHasFinishedFakeLoading(true)
```
gustaveen, line 54: Why do we need this fake loading state?

gustaveen (approved):

src/utils/ExchangeTokenRetrieval.tsx
palmenhq: It's too fast 😂 looks like a content flash if we don't debounce the loading state a bit imo.

gustaveen: Ah, that's what I suspected 😄

PR #326, opened 2020-06-09
Add some og tags for sharing

palmenhq:
Since this page will be shared through email and sms and such we should probably have a proper page title and image at least 😬


gustaveen (approved):
👍

PR #327, opened 2020-06-09
Storybook setup and initial Forever component

gustaveen:
Also refactored inputs to enable dark and light variants

Screenshot 2020-06-11 at 17 34 41

inputs


palmenhq (commented):
awesome! just one thing

src/components/inputs/index.tsx
```diff
@@ -195,10 +227,11 @@ export const InputField: React.FC<TextInputProps &
   options,
   touched,
   errors,
+  variant = 'light',
```
vonElfvin, line 230: Maybe avoid default

src/components/inputs/inputs.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+      name="foo"
+      type="text"
+      autoComplete="off"
+      placeholder=""
+      touched={false}
+      errors={text('Error text', '')}
+      variant="dark"
+    />
+  </div>
+)
```
vonElfvin, line 47: Maybe create a generic one that takes the variant as input and then return that for Light/Dark

src/components/inputs/index.tsx
```diff
@@ -34,17 +34,45 @@ export const masks: Record<string, Mask> = {
...
+    background: colorsV3.gray900,
+    border: colorsV3.gray500,
+    error: colorsV3.red500,
+  },
+}
+
+const Wrapper = styled.div<{ errors?: string; variant: variantType }>`
   position: relative;
-  background: ${colorsV2.white};
+  background-color: ${(props) => inputVariants[props.variant].background};
```
vonElfvin, line 63: Nice clarification

src/components/inputs/index.tsx
```diff
@@ -171,18 +200,21 @@ const StyledRawInput = styled.input`
 export const RawInputField: React.FC<React.InputHTMLAttributes<
   HTMLInputElement
 > & {
-  label: string
+  label?: string
   errors?: string
-}> = ({ errors, label, className, ...props }) => (
+  variant?: variantType
+}> = ({ errors, label, className, variant = 'light', ...props }) => (
```
vonElfvin, line 206: Do we want to default the variant?

src/components/inputs/inputs.stories.tsx
vonElfvin: Although I can see now these are just test examples, I guess this is better then.

src/components/inputs/inputs.stories.tsx
```diff
@@ -0,0 +1,62 @@
...
+}
+
+const darkBackground = {
+  parameters: {
+    backgrounds: [{ name: 'gray900', value: colorsV3.gray900, default: true }],
+  },
+}
+
+Light.story = lightBackground
+Dark.story = darkBackground
```
vonElfvin, line 62: Even though the only thing in the "story" is background I'd prefer if the name of the variables are `lightStory` and `darkStore`. Or `storyWithLightBackground`. I thought these backgrounds were used someplace else with the current variables names.

src/pages/Forever/components/RedeemCode.tsx
```diff
@@ -0,0 +1,126 @@
...
+  width: 100%;
+  min-width: 12rem;
+  height: 3.5rem;
+
+  @media (min-width: 480px) {
+    width: auto;
+  }
+`
+
+export const RedeemCode: React.FC<RedeemCodeProps> = ({
```
vonElfvin, line 75: If this is to be generic maybe we should add "setCode" and "submitCode" inputs for external specification of where to send the code. If it is a single use component then having that logic here in the future is sufficient I guess.

src/pages/Forever/components/RedeemCode.tsx
```diff
@@ -0,0 +1,126 @@
...
+          <HedvigLogo width={94} />
+        </LogoLink>
+      </Header>
+
+      <Main>
+        <Paragraph>{textKeys.FOREVER_LANDINGPAGE_INPUT_TEXT()}</Paragraph>
+        <Formik
+          initialValues={{ code }}
+          onSubmit={() => {
+            // Submit
```
vonElfvin, line 93: Having external control of this would be cool

src/pages/Forever/components/RedeemCode.tsx
```diff
@@ -0,0 +1,126 @@
...
+          }}
+        >
+          {({ values }) => (
+            <Form>
```
vonElfvin, line 97: Similarly external validation could be considered.

src/pages/Forever/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+          content="https://www.hedvig.com/new-member-assets/social/hedvig-hemforsakring-2.jpg"
+        />
+      </Helmet>
+      <Page>
+        <Global
+          styles={css`
+            body {
+              background-color: ${colorsV3.gray900};
+            }
+          `}
```
vonElfvin, line 44: Do we like adding styles like this?

src/routes.tsx
```diff
@@ -170,6 +171,11 @@ export const reactPageRoutes: ReactPageRoute[] = [
     Component: Referral,
     exact: true,
   },
+  {
+    path: LOCALE_PATH_PATTERN + '/forever/:code',
```
vonElfvin, line 175: code is optional right?

vonElfvin (dismissed):
Added some thoughts, great work!

src/pages/Forever/components/RedeemCode.tsx
vonElfvin: Similarly, maybe session management should be external and thus the memberId/token could be handled outside as well. Making external implementation more reasonable.

vonElfvin: Or maybe you can access the token from the cache or something like that so maybe this is fine 🤷

src/components/inputs/index.tsx
gustaveen: I wanted to avoid breaking the existing implementations :) But we can remove this once we're happy with the new implementation

src/pages/Forever/components/RedeemCode.tsx
gustaveen: I agree, I might split this up when coming further in the implementation

src/pages/Forever/index.tsx
gustaveen: It's not optimal but works ok I think for setting a specific rule for a view. But body should be black on most of the views so will most likely move this further up

src/routes.tsx
gustaveen: Good catch! Will not merge this now I realise since we have to change the url i market-web first

PR #331, opened 2020-06-25
Add on submit

gustaveen:


src/pages/Forever/index.tsx
```diff
@@ -50,6 +54,36 @@ export const Forever: React.FC<ForeverProps> = ({
...
+      if (result.errors && result.errors.length > 0) {
+        // TODO handle errors
+        actions.setFieldError('code', 'FOREVER_ADD_CODE_ERROR')
+        return
+      }
+      // TODO redirect to success screens
+      history.push(`/${currentLocale}/new-member`)
+    } catch (e) {
+      // tslint:disable-next-line no-console
+      console.error(e)
```
palmenhq, line 81: i think rather we should capture a sentry exception here with this https://github.com/HedvigInsurance/web-onboarding/blob/a8d4c94357f733211a783851bd392c6f0f2f8ca6/src/utils/sentry-client.ts#L1

palmenhq (approved):
clean!

PR #333, opened 2020-06-29
Feature/forever error handling

gustaveen:
- Add loading state to submit button
- Add error messages to result 

loading

### DIFF
--- a/src/pages/Forever/useRedeemCode.ts
+++ b/src/pages/Forever/useRedeemCode.ts
@@ -1,0 +1,81 @@
+import { useCurrentLocale } from 'components/utils/CurrentLocale'
+import { useRedeemCodeV2Mutation } from 'data/graphql'
+import { FormikHelpers } from 'formik'
+import { useHistory } from 'react-router'
+import { captureSentryError } from 'utils/sentry-client'
+import { RedeemCodeFormValue } from './components/RedeemCode'
+
+export const useRedeemCode = () => {
+  const [redeemCode] = useRedeemCodeV2Mutation()
+  const locale = useCurrentLocale()
+  const history = useHistory()
+
+  const handleSubmit = async (
+    form: RedeemCodeFormValue,
+    actions: FormikHelpers<RedeemCodeFormValue>,
+  ) => {
+    try {
+      const result = await redeemCode({ variables: { code: form.code } })
+      if (!result) {
+        return
+      }
+
+      if (result.data?.redeemCodeV2.__typename === 'SuccessfulRedeemResult') {
+        // TODO redirect to correct success screens
+        history.push(`/${locale}/new-member`)
+        return
+      }
+      if (result.data?.redeemCodeV2.__typename === 'CannotRedeemOwnCampaign') {
+        actions.setErrors({ code: 'FOREVER_ERROR_CANNOT_REDEEM_OWN_CAMPAIGN' })
+        return
+      }
+      if (
+        result.data?.redeemCodeV2.__typename ===
+        'CampaignCannotBeCombinedWithExisting'
+      ) {
+        actions.setErrors({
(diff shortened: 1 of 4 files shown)


src/pages/Forever/useRedeemCode.ts
```diff
@@ -0,0 +1,81 @@
...
+        'MemberIsNotEligibleForCampaign'
+      ) {
+        actions.setErrors({
+          code: 'FOREVER_ERROR_CAMPAIGN_CANNOT_BE_COMBINED_WITH_EXISTING',
+        })
+        return
+      }
+
+      if (result.data?.redeemCodeV2.__typename === 'CampaignDoesNotExist') {
+        actions.setErrors({ code: 'FOREVER_ERROR_CAMPAIGN_DOES_NOT_EXIST' })
```
palmenhq, line 56: we could make this a switch statement instead, right?

src/pages/Forever/useRedeemCode.ts
```diff
@@ -0,0 +1,81 @@
...
+        result.data?.redeemCodeV2.__typename ===
+        'CannotRedeemCampaignFromDifferentMarket'
+      ) {
+        actions.setErrors({
+          code: 'FOREVER_ERROR_CANNOT_REDEEM_CAMPAIGN_FROM_DIFFERENT_MARKET',
+        })
+        return
+      }
+
+      if (result.errors && result.errors.length > 0) {
```
palmenhq, line 70: `result.errors?.length > 0`?

src/pages/Forever/useRedeemCode.ts
```diff
@@ -0,0 +1,81 @@
+import { useCurrentLocale } from 'components/utils/CurrentLocale'
+import { useRedeemCodeV2Mutation } from 'data/graphql'
+import { FormikHelpers } from 'formik'
+import { useHistory } from 'react-router'
+import { captureSentryError } from 'utils/sentry-client'
+import { RedeemCodeFormValue } from './components/RedeemCode'
+
+export const useRedeemCode = () => {
+  const [redeemCode] = useRedeemCodeV2Mutation()
```
palmenhq, line 9: what about also returning the loading state from here? You could use `const [redeemCode, { loading }] = useRedeemCodeV2Mutation()` or something. Instead of creating our own loading state

src/pages/Forever/useRedeemCode.ts
gustaveen: very much so! just made it simple and ugly to start with :P

src/pages/Forever/useRedeemCode.ts
gustaveen: this is something I copied from another file. will the respons ever be anythin other than `result.data`? should remove this otherwise right

src/pages/Forever/useRedeemCode.ts
gustaveen: I used the built in `isSubmitting` from Formik, but we could totally use that if it seems off

palmenhq: Aah sorry i missed that. I guess thats ok!

src/pages/Forever/useRedeemCode.ts
palmenhq: yes, there might be `errors` if something goes wrong server side or on the network, for example

gustaveen: Typescript still complains about `Object is possibly 'undefined'` if doing that 🤷‍♂️

palmenhq: even with the `?.` operator?

palmenhq (approved):
Awesome!

src/pages/Forever/useRedeemCode.ts
palmenhq: After having investigated this, yes. I guess it doesn't like comparing undefined as a number (which kinda makes sense to be fair)

PR #336, opened 2020-07-06
Track studentkortet if code is `stuk2` additionally to `studentkortet`

palmenhq:


src/utils/tracking.ts
```diff
@@ -145,7 +145,9 @@ export const useTrack = ({ offerData, signState }: TrackProps) => {
 
     if (
       redeemedCampaigns?.length > 0 &&
-      redeemedCampaigns[0].code.toLowerCase() === 'studentkortet'
+      ['studentkortet', 'stuk2'].includes(
+        redeemedCampaigns[0].code.toLowerCase(),
+      )
```
vonElfvin, line 150: For clarity maybe the array should be in a separate variable, or the check itself in a helper function

vonElfvin (approved):
Coolioo

src/utils/tracking.ts
palmenhq: If we'd do this in > 1 place i'd agree but imo this is ok

PR #338, opened 2020-07-06
Hedvig forever: Language picker

palmenhq:
Also adds some padding to the bottom footer in order for the language picker not to hide behind the bottom toolbar


rasmusguterstam (approved):
😎🌟🌟

PR #341, opened 2020-07-08
Hedvig forever: Smash bugs 🐞

palmenhq:
1) update picked locale when switching locale
  2) don't print undefined for empty code

(Caught before launch even, kinda)


rasmusguterstam (approved):
🐞💥

PR #342, opened 2020-07-08
Remove await refetch queries from redeemcodev2 temporarily

palmenhq:


vonElfvin (approved):
You could add a sleep for now just to ensure it is done while still giving an okayish ux

palmenhq:
@vonElfvin i think it should be ok because we have some loading state on the next page. Might result in a tiny content flash on slow connections but i think that's ok

PR #343, opened 2020-07-09
Add support to show text for visible no discount type

vonElfvin:


vonElfvin:
I have no idea why the snapshots fail in the CI, I haven't touched the language picker at all

PR #345, opened 2020-07-13
Fix landing boxes so they don't hide below fold or are too wide

palmenhq:
Screenshot 2020-07-13 at 13 29 52
Screenshot 2020-07-13 at 13 30 07


vonElfvin (approved):
Neat

PR #347, opened 2020-07-15
Fix segment identify call

sampettersson:
This fixes so that we call segments identify function with the userId, I need this to be able to see if people download the app later in Mixpanel so I took the liberty to fix it 😇 

It seems like we tried to do this before but the implementation was kinda wrong.


palmenhq (dismissed):
Snyggt! But ofc lets fix the tests :)

sampettersson:
@palmenhq nu är den glad 😄

palmenhq (approved):

PR #350, opened 2020-07-20
Disable SSR

palmenhq:
Finally


src/routes.tsx
```diff
@@ -1,26 +1,83 @@
...
+    path: LOCALE_PATH_PATTERN + '/new-member/connect-payment/retry',
+    titleTextKey: 'ONBOARDING_CONNECT_DD_PAGE_TITLE',
+  },
+  {
+    path: LOCALE_PATH_PATTERN + '/new-member/(offer|sign)',
+    titleTextKey: 'OFFER_PAGE_TITLE',
+  },
+  {
+    path: LOCALE_PATH_PATTERN + '/new-member/sign/success',
+    titleTextKey: '',
```
AlexSJ1990, line 62: should this be an empty string?

AlexSJ1990 (approved):
Looks good! 👍

src/routes.tsx
palmenhq: Nice catch! It is intentional though, because `/sign/success` is only used for iframing

PR #354, opened 2020-08-12
Fix broken storybook webpack config

gustaveen:


.storybook/main.js
```diff
@@ -10,7 +10,8 @@ module.exports = {
     'storybook-addon-paddings',
   ],
   webpackFinal: async (config) => {
-    config.resolve.modules.push(path.resolve(__dirname, '../src'))
+    config.resolve.modules.push(path.resolve(__dirname, '../src/client'))
+    config.resolve.alias = { shared: path.resolve(__dirname, '../src/shared') }
```
palmenhq, line 14:
```js
    config.resolve.alias = {
      ...config.resolve.alias,
      shared: path.resolve(__dirname, '../src/shared')
    }
```

maybe? :)

gustaveen: Ah! Makes sense 🙌

PR #355, opened 2020-08-13
Switch to referral code in data layer

gustaveen:


src/types/gtm.d.ts
```diff
@@ -6,7 +6,7 @@ declare global {
 
 interface OfferData {
   insurance_type: TypeOfContract
-  campaign_code: 'yes' | 'no'
+  referral_code: 'yes' | 'no'
```
palmenhq, line 9: unless this is an external dependency should we name this something like `has_referral_code` or so? I'd expect `referral_code` to be the actual code

palmenhq (approved):

src/types/gtm.d.ts
gustaveen: Yeah it sounds a bit confusing, it's just the way it's set up in google tag manager by precis digital 🙃

PR #358, opened 2020-08-20
Add combo tracking

gustaveen:
Broke out the gtm tracking and added tracking for norwegian combo quotes


src/client/utils/tracking/gtm.ts
```diff
@@ -0,0 +1,58 @@
...
+}
+
+export interface DataLayerObject {
+  event: string
+  offerData?: GTMOfferData
+}
+
+const pushToGTMDataLayer = (obj: DataLayerObject) => {
+  window.dataLayer = window.dataLayer || []
+  window.dataLayer.push(obj)
```
gustaveen, line 26: TS started to complain all of a sudden about `Property 'dataLayer' does not exist on type 'Window & typeof globalThis'` but I only moved the file into a folder? 🤷‍♂️ @palmenhq It is defined in gtm.d.ts

palmenhq: i'd just cast `window` to `any` in this particular case 🙈

gustaveen: yup let's just do that 😄

src/client/utils/tracking/gtm.ts
```diff
@@ -0,0 +1,57 @@
...
+  castedWindow.dataLayer = castedWindow.dataLayer || []
+  castedWindow.dataLayer.push(obj)
+}
+
+export const trackOfferGTM = (
+  offerData: OfferData,
+  referralCodeUsed: boolean,
+) => {
+  try {
+    const contractType = () => {
```
palmenhq, line 36: Maybe we cold rename this to `getContractType` or `mapContractType` or something? Just to imply this variable is a funciton

palmenhq (commented):
Sorry, two more thoughts 😅

src/client/pages/OfferNew/utils.ts
```diff
@@ -108,9 +108,8 @@ export const isBundle = (offerData: OfferData): boolean =>
 export const isYouth = (offerData: OfferData): boolean =>
   offerData.quotes.every(
     (quote) =>
-      isNorwegianHomeContents(quote.quoteDetails) &&
-      isNorwegianTravel(quote.quoteDetails) &&
-      quote.quoteDetails.isYouth,
+      isNorwegianHomeContents(quote.quoteDetails) ||
+      (isNorwegianTravel(quote.quoteDetails) && quote.quoteDetails.isYouth),
```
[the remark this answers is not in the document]
gustaveen: true, that order would make more sense. but the reason is `Property 'isYouth' does not exist on type 'SwedishApartmentQuoteDetails'` But it feels like there's a much simpler solution I can't think of right now haha

src/client/utils/tracking/gtm.ts
gustaveen: indeed!

palmenhq (approved):

PR #362, opened 2020-08-28
Add canonical tag to prevent google to index multiple urls

gustaveen:
We've seen that google indexed certain pages with discount codes, hopefully this will prevent that. 
And after speaking with Carl, we should only index the start of the onboarding.


src/server/page.ts
```diff
@@ -103,6 +103,7 @@ const template = (
           'https://www.hedvig.com/f/62762/1920x1080/a8d806bbbf/background.png'}"
      />
      <meta name="google-site-verification" content="AZ5rW7lm8fgkGEsSI8BbV4i45ylXAnGEicXf6HPQE-Q" />
+     <link rel="canonical" href="https://hedvig.com/${locale}/new-member" />
```
palmenhq, line 106: what about `/forever`? Maybe we should do something that is slightly smarter than this..? I feel also like this will be missed if we would ever do something not related to `/new-member` updating this will be forgotten. Please challenge this though :D

gustaveen: From what I understood we don't want to index `/forever` either, but as you said this might not be the best way to implement this. When I think of it we should move this to `/embark/index.tsx` instead, so it will only effect the "onboarding flow" pages. But I might be misunderstanding the use of canonical links :D WDYT? @palmenhq

gustaveen: Updated now, we should probably only have it on `Landing.tsx` :D

palmenhq (approved):
looks great! Assuming this is ok to have client-side only?

styrelius:
> looks great! Assuming this is ok to have client-side only?

What would be the alternative? Just curious 🤖

palmenhq:
@styrelius good question! Right now what will happen is that the server will serve the HTML template without this tag, and as react renders it’ll inject this meta tag in the header. This might be fine, idk, but we might want to consider including it in the HTML template sent from the server. Because spiders/crawlers don’t always run the JavaScript, sometimes they just inspect the HTML template and if that’s true it’d miss this canonical tag :)

styrelius:
@palmenhq Ah, of course! That's very true

gustaveen:
Let's see if this helps, otherwise we can implement something smarter 👍

PR #364, opened 2020-09-02
White list trustly's test environment as valid postmessage origin

palmenhq:
also put the modal on top of the header


AlexSJ1990 (approved):
🚢

PR #365, opened 2020-09-08
Add avy as variation and post messages

palmenhq:
Also disable no-frame options

### DIFF
--- a/src/client/pages/ConnectPayment/components/TrustlyModal.tsx
+++ b/src/client/pages/ConnectPayment/components/TrustlyModal.tsx
@@ -2,9 +2,8 @@ import styled from '@emotion/styled'
 import { colors } from '@hedviginsurance/brand'
 import { TranslationsConsumer } from '@hedviginsurance/textkeyfy'
 import Modal from 'components/Modal'
-import { CurrentLocale } from 'components/utils/CurrentLocale'
-import { Container } from 'constate'
-import React from 'react'
+import { useCurrentLocale } from 'components/utils/CurrentLocale'
+import React, { createRef, useEffect, useState } from 'react'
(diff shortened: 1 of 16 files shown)


src/client/pages/ConnectPayment/components/TrustlyModal.tsx
```diff
@@ -77,53 +80,34 @@ export const TrustlyModal: React.FC<Props> = ({
   }, [])
 
   return (
-    <Container<State, Actions>
```
gustaveen, line 80: Whoop whoop, nice refactoring! 🤩  Would be nice to get rid of constate overall. 
Just me being picky, but I think it's handy to keep the refactoring in a commit by it's own 🙃

gustaveen (dismissed):

src/client/utils/Intercom.tsx
```diff
@@ -29,7 +29,9 @@ export const Intercom: React.FC = () => {
     LOCALE_PATH_PATTERN + '/new-member/connect-payment/direct',
   )
   React.useEffect(() => {
-    if ([Variation.IOS, Variation.ANDROID].includes(variation!)) {
+    if (
+      [Variation.IOS, Variation.ANDROID, Variation.AVY].includes(variation!)
```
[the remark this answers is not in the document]
palmenhq: `?variation=avy`

src/client/pages/ConnectPayment/components/TrustlyModal.tsx
palmenhq: I'd rather argue it's better to do it in small chunks in case something breaks. Thoughts on that? :)

gustaveen (approved):

PR #370, opened 2020-09-22
Fix broken favion url and add query string

gustaveen:
See update to web-next: https://github.com/HedvigInsurance/web-next/pull/540


palmenhq (approved):
whoups

PR #371, opened 2020-09-22
Lokalise web onboarding

palmenhq:
Yeah, here we are :)

It looks worse than it is size-wise


gustaveen:
Sweet! Looks really good, great job @palmenhq  🙌 
Was just thinking if we should document the workflow more in detail. As I understand it now (tell me if that's correct):

- You add the new textkey in the component and in en.json (with or without a translation?)
- You run `yarn download-translations` to sync the new textkey to Lokalise (or is there another cmd to push it to Lokalise?)
- Someone adds the translations in Lokalise
- You run `yarn download-translations` to update the translations files
- You commit the changes and do a deploy to get the translations into production

hedvigoscar:
For reference, this is the workflow we use on the apps:

- Design creates screens in their tooling (Figma)
- Design uses the Lokalise Figma-plugin to export keys in the Figma document to Lokalise
- Translators fulfill translations in Lokalise
- We use the Lokalise integrations to export the translations to the apps

src/client/utils/textKeys.tsx
```diff
@@ -0,0 +1,165 @@
...
+export const TextKeyContext = React.createContext<TextKeys | null>(null)
+
+export const replacePlaceholders = (
+  resolvedKey: string,
+  replacements?: Replacements,
+) => {
+  if (!replacements) {
+    return resolvedKey
+  }
+  const matches = resolvedKey.split(placeholderRegex).filter((value) => value)
```
styrelius, line 26: Seems to me like the `filter` function doesn't really do anything here? 🤔 Or am I missing something?

styrelius:
Nice! Also, I agree with Gustav that it's probably a good idea to have more extensive documentation on the workflow 📜

src/client/utils/textKeys.tsx
hedvigoscar: It performs a truthiness-check on `value`, which means that all values that are `undefined`, `null` or `false` (and maybe some others, javascript is weird) get removed 🙂

styrelius: Aha, of course! 🧠

gustaveen (approved):
🥳

PR #373, opened 2020-09-29
Adjust CSP rules

gustaveen:
## What
Remove nonce value from inline scripts and add `unsafe-inline` rule to the CSP

## Why
This is due to a bug where we could not get any injected GTM scripts to run nor the GTM preview mode. Since this is suboptimal we should aim to restore this.


src/server/middleware/helmet.ts
```diff
@@ -22,6 +21,8 @@ const defaultSrc = [
   'https://tagmanager.google.com',
   'www.googleadservices.com',
   'www.gstatic.com',
+  'https://fonts.gstatic.com',
+  'https://www.google-analytics.com',
```
palmenhq, line 25: U sure about this one? We have it on L20 already..

src/server/middleware/helmet.ts
```diff
@@ -68,13 +69,9 @@ export const helmet = koaHelmet({
       defaultSrc,
       scriptSrc: [
         "'unsafe-eval'",
+        "'unsafe-inline'",
```
palmenhq, line 72: my security heart is crying but i know we have to atm

src/server/middleware/helmet.ts
gustaveen: ah my bad, missed that one. the font is needed though for the GTM preview, the icons is messed up otherwise

src/server/middleware/helmet.ts
gustaveen: I feel you! 😅

src/server/middleware/helmet.ts
gustaveen: also needed a few other for the preview mode in optimize apparently

palmenhq (approved):
😭

PR #379, opened 2020-10-06
[HVG-134] Fix iframe content window access issue

palmenhq:
This may (or may not, i'm unsure) fix the issue where trustly keeps spinning for some, but it will for sure fix the issue about illegal frame accessing.


palmenhq:
I'm just gonna close this since it doesn't acutally solve the problem.

PR #381, opened 2020-10-13
Chore / add eslint import order rule

styrelius:
### What?
Enforce ordered imports with ESLint:
1) Install `eslint-plugin-import` and `eslint-import-resolver-typescript.
2) Add `import/order` rule in `.eslintrc.js`, with the same order as in `web-next` repo.
3) Add `typescript` `import/resolver` in `.eslintrc.js` in order for the setup to work properly.

### Why?
Less mess is beneficial 😇


palmenhq (approved):
Nice!

PR #383, opened 2020-10-13
Feature/add quote debugger

gustaveen:
Web onboarding debugger is back! Updated it to take locale and market into account. 
Screenshot 2020-10-13 at 17 25 13


src/client/pages/Debugger/index.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+const ActualDebugger = React.lazy(() =>
+  import(/* webpackChunkName: "debugger" */ './ActualDebugger').then((m) => ({
+    default: m.ActualDebugger,
+  })),
+)
+
+export const Debugger: React.FC = () => {
+  const [mounted, setMounted] = React.useState(false)
+  React.useEffect(() => setMounted(true), [])
```
styrelius, line 11: Getting a bit confused by this `useEffect` 🤔 I thought the linter would complain about `setMounted` being absent from the dependency array? And won't `setMounted(true)` also be implicitly returned here, which may not be what we want since it will then be used as cleanup function? Or maybe that's not how it works 🤷‍♀️

styrelius: or maybe the linter wouldn't complain since I apparently didn't add `eslint-plugin-react-hooks` hehe 🤷‍♀️

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -0,0 +1,495 @@
...
+                    currentInsurer: '',
+                    birthDate: '1995-09-29',
+                    ssn: '',
+                    startDate: '',
+                    email: 'blargis@hedvig.com',
+                    norwegianHomeContents: {
+                      coInsured: 0,
+                      isYouth: false,
+                      livingSpace: 0,
+                      street: 'Gulebøjsveien 1',
```
palmenhq, line 141: I love how this keeps coming back 😂

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -0,0 +1,495 @@
...
+export const NorwegianTravel: React.FC<WithFormikProps> = ({ formik }) => {
+  return (
+    <>
+      <InputField
+        label="Co-insured"
+        placeholder="1"
+        type="number"
+        {...formik.getFieldProps('norwegianTravel.coInsured')}
+      />
+      <div>isYouth TODO</div>
```
palmenhq, line 452: shall we do this now or nah?

src/client/pages/Debugger/index.tsx
palmenhq, line 11: isn't react suspense doing this logic for us tho?

palmenhq (commented):
shall we do something more clever in the debugger re combo quotes and stuff?

src/client/pages/Debugger/components/Offer.tsx
gustaveen: never gets old! 😂

src/client/pages/Debugger/index.tsx
gustaveen: true, we can remove that all together

gustaveen:
> shall we do something more clever in the debugger re combo quotes and stuff?

Sure! But maybe do it in a new PR?

styrelius:
I've had some weird behaviours locally with this but maybe it doesn't really matter. I mean since it's not used in prod we could fix those things later on..? Or what do you think?

gustaveen:
That was a really weird bug but I agree, it's not super crucial since it just a tool for us to test at the moment

styrelius:
@palmenhq What's your opinion? It's still possible to use the debugger, the bug just some weirdness with the `storageState` `quoteIds`

palmenhq:
@styrelius @gustaveen How bad is it..? Like is it quirky in quirky cases or is it hard to use?

styrelius:
> @styrelius @gustaveen How bad is it..? Like is it quirky in quirky cases or is it hard to use?

It's not really that bad - it doesn't make it hard to use. And also I haven't been able to reproduce the bug since before lunch. I say we merge this and maybe I can continue to look into it since I also wanted to make another change to this in a new branch/PR (fetching a new UUID automatically so you don't have to copy-paste it)

gustaveen:
It worked fine for me so let's merge IMO

styrelius (approved):

PR #385, opened 2020-10-26
[HVG-247] Fix / slightly smarter quote debugger

styrelius:
### What?

Make some changes to the quote debugger to make it a bit easier to work with:

1) Use the `uuid` package to generate a new quote UUID every time the `Offer` component mounts.
2) Add more initial form values to you don't have to add anything to the form (unless you want to) in order to create a quote.
3) Move the market specific input fields to their own files: `QuoteFormSweden.tsx`, `QuoteFormNorway.tsx` and `QuoteFormDenmark.tsx`.
4) Move the initial form values for different markets into these new files and add `initialFormValue` as properties to the objects in `quotesByMarket` to make them easily accessible.
5) Add a `hasQuoteCreatingError` state in order to render a message when there's an error in the quote creating.
6) Add `handleSubmit` function and move logic there that was copy-pasted three times.

Also add `da_DK` and `en_DK` to `getPickedLocaleFromCurrentLocale` in order to get the right perils back when calling `getQuote` from `useQuoteLazyQuery`.

<br>

### Why?

Having to actually write stuff in the form when creating a quote gets tiring fast and the lack of response when something went wrong also made it very confusing for me. Of course you can go to the "Network" tab but with a web socket connection it isn't super easy to find the right errors. Also the `Offer.tsx` file was really big and in my opinion really hard to get an overview of, with a lot of logic being repeated three or four times.

In addition to that we didn't use the right locale when fetching perils, so that needed to be fixed to get the right perils.

_Referenced ticket [here](https://hedvig.atlassian.net/browse/HVG-247)_ 

<br>

### Screenshot

**Placeholder perils:**
2020-10-26-placeholder-perils

### GIF

**Creating a danish quote with the debugger:**

2020-10-26-quote-debugger

<hr>
<hr>

### Update
The error state is now the actual error message sent from the backend instead of just a boolean:
2020-10-26-quote-debugger-error


styrelius:
Of course there are still a lot that could be done better here, like for example we still don't have a solution for quote bundles but let's fix that in another PR, after we get the whole signing up flow working for Danish home insurance?

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -94,6 +164,7 @@ export const Offer: React.FC = () => {
             {() => (
               <InputField
                 label="Quote id"
+                name="Quote id"
```
gustaveen, line 167: Now that we generate the uuid, we might as well set it the input to disabled since there's no need to edit it right?
```suggestion
                name="Quote id"
                disabled
```

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -24,63 +38,119 @@ enum QuoteType {
...
     }
+
+    storageState.session.setSession({
+      ...storageState.session.getSession(),
+      quoteIds: [quoteId],
+    })
   }, [quoteId])
 
+  useEffect(() => {
+    if (quoteId?.length === 36 && sessionToken) {
```
gustaveen, line 149: If we disabled the uuid input field you can remove the length check

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -24,63 +38,119 @@ enum QuoteType {
...
+        localeIsoCode,
+      )({
+        input: {
+          ...values,
+          id: quoteId,
+          currentInsurer: values.currentInsurer || undefined,
+          // @ts-ignore
+          startDate: values.startDate || undefined,
+        },
+      }).catch(() => setHasQuoteCreatingError(true))
```
gustaveen, line 122: Now that we get a somewhat better error messages, an idea would be to print it out to give you some clue on the error
```suggestion
      }).catch((e) => {
         setQuoteCreatingError(e.message)
       })
```

gustaveen (commented):
Nice push Ellen! 🙌

src/client/pages/Debugger/components/Offer.tsx
styrelius: Yep, that might be a good idea. 👍 I think I tested it but I don't remember why I removed it haha!

src/client/pages/Debugger/components/Offer.tsx
styrelius: But we want to be able to paste in an existing id and get the quote maybe?

styrelius: Is that a reason to leave it as is @gustaveen ? What do you think @palmenhq ?

src/client/pages/Debugger/components/Offer.tsx
styrelius: Have added the error message!

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -124,371 +197,90 @@ export const Offer: React.FC = () => {
...
+              </Formik>
             </>
           )}
           {data?.quote && <pre>{JSON.stringify(data, null, 2)}</pre>}
+          {quoteCreatingError && (
+            <>
+              <h2>Something went wrong 😔</h2>
+              <p>
+                <u>Error message:</u> {quoteCreatingError}
+              </p>
```
gustaveen, line 278: Might be easier to read in a pre tag :) 
```suggestion
              <p>Error message:</p> 
               <pre>{quoteCreatingError}</pre>
```

styrelius: It actually gets worse then since the `<pre>` tag makes it use the line breaks from the original formatting - which means _no_ line breaks.

styrelius: image

gustaveen: I thought it was easier to read when you got a longer error massage actually 😁  But let's go with a paragraph 👍
Screenshot 2020-10-27 at 09 41 35

src/client/pages/Debugger/components/Offer.tsx
gustaveen: But right now we can't fetch an already created quote right (and I don't think we need it). If you edit the id you will end up creating a new quote

styrelius: Well actually 🤓 If you paste an id of a quote you already created in the quote id field the quote is fetched and the data for that quote is displayed. Since the fetching is triggered by change to the quote id. So we can fetch already created quotes! 
I don't see the use case for it either though, but I have a faint memory of @palmenhq saying that was the idea behind having an input field for the id in the first place.

src/client/pages/Debugger/components/Offer.tsx
styrelius: Haha WOW never gotten that error message 😅 What did you do to trigger that?

styrelius: And YES, you're of course right it's easier to read those really long error messages with pre tag - I've just been so focused on the errors from submitting wrong id so didn't check this 😳

styrelius: OK, the best solution was the `<pre>` tag with `whiteSpace: 'pre-wrap'` styling to make it not overflow:
image
Thank you mister for being so persistent! 🙏

src/client/pages/Debugger/components/QuoteFormSweden.tsx
```diff
@@ -0,0 +1,61 @@
...
+        {...formik.getFieldProps('swedishApartment.street')}
+      />
+      <InputField
+        label="Zip code"
+        placeholder="12345"
+        {...formik.getFieldProps('swedishApartment.zipCode')}
+      />
+      <InputField
+        label="Type"
+        placeholder=""
```
palmenhq, line 50: is this required?

src/client/pages/Debugger/components/QuoteFormNorway.tsx
```diff
@@ -0,0 +1,86 @@
...
+        {...formik.getFieldProps('norwegianHomeContents.street')}
+      />
+      <InputField
+        label="Zip code"
+        placeholder="1234"
+        {...formik.getFieldProps('norwegianHomeContents.zipCode')}
+      />
+      <InputField
+        label="Type"
+        placeholder=""
```
palmenhq, line 63: same do we need it?

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -124,371 +197,93 @@ export const Offer: React.FC = () => {
...
-                        // @ts-ignore
-                        startDate: values.startDate || undefined,
-                      },
-                    })
+              <Formik
+                initialValues={
+                  getCurrentAvailableQuoteData(currentMarket, quoteType)
+                    ?.initialFormValues || {}
+                }
+                onSubmit={async (values) => handleSubmit(values, storage)}
```
palmenhq, line 205: since we don't have any `await` this does not have to be async.

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -124,371 +197,93 @@ export const Offer: React.FC = () => {
...
+                    </Form>
+                  </>
+                )}
+              </Formik>
             </>
           )}
           {data?.quote && <pre>{JSON.stringify(data, null, 2)}</pre>}
+          {quoteCreatingError && (
+            <>
+              <h2>Something went wrong 😔</h2>
```
palmenhq, line 275: Aaw ❤️

palmenhq (dismissed):
looks good, just some nit

src/client/pages/Debugger/components/Offer.tsx
styrelius: OK, since we agreed on keeping the UUID input field as it is (for now, maybe adding a button to generate a UUID in the future) I will resolve this conversation! 😇

src/client/pages/Debugger/components/Offer.tsx
styrelius: OK, since we agreed on keeping this as it is (for now, maybe adding a button to generate a UUID in the future) I will resolve this conversation! 😇

src/client/pages/Debugger/components/QuoteFormSweden.tsx
styrelius: For Sweden it is required to pass the type in order to create a quote yes. Or maybe you mean we should actually add `required` to all input fields that are required?

src/client/pages/Debugger/components/QuoteFormNorway.tsx
styrelius: Yes :)
image

styrelius: But we actually don't need `ssn` so maybe I should remove that

src/client/pages/Debugger/components/Offer.tsx
styrelius: Ah, I believe I thought that since `handelSubmit` is an `async` function we needed the `async` here as well. 😁 But you are of course very right, that's not how it works. I will remove that!

src/client/pages/Debugger/components/QuoteFormSweden.tsx
styrelius: Or did you mean the `placeholder`? I don't remember, since those were already there when I started working one this I assumed it's something required by Formik. I'll look into it!

styrelius: OK so it's required in the `interface CoreInputFieldProps` in `/inputs/index.tsx`. Can I make that property optional maybe? Or will that mess up something else?

palmenhq: I think we should make it optional, empty string just feels silly

src/client/pages/Debugger/components/Offer.tsx
gustaveen: Snyggt! 👌  Haha, sorry for being a nag 😆

src/client/pages/Debugger/components/QuoteFormSweden.tsx
styrelius: 👍

palmenhq (dismissed):
Let’s just fix the typecheck and we’re good to go imo! Much awesomeness in here 😎

gustaveen (approved):
Wohoo! Awesome @styrelius 🥳
giphy

PR #386, opened 2020-10-26
[HVG-156] Fixed on submit not firing in Checkout

virrios:
Norway: Not possible to sign on Android

Added onSubmit function and hidden input field to trigger sign.

### DIFF
--- a/src/client/pages/OfferNew/Checkout/CheckoutContent.tsx
+++ b/src/client/pages/OfferNew/Checkout/CheckoutContent.tsx
@@ -59,6 +59,7 @@ interface Props {
   refetch: () => Promise<void>
   onEmailUpdate: (onCompletion: Promise<void>) => void
   onSsnUpdate: (onCompletion: Promise<void>) => void
+  onSubmit: () => void
 }
 
 export const CheckoutContent: React.FC<Props> = ({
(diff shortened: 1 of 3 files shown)


gustaveen (approved):
Great job! @virrios 🥳

gustaveen:
Unrelated to the issue, it got me thinking when testing this, if we should move the date picker to the top? So adding your ssn and submitting is the last input? Also the datepicker is missing it's label

palmenhq:
@gustaveen which date picker? 😅

gustaveen:
This one (where it says Idag) 🙂  
Might be a bit unclear what the date is for if we don't show a label @palmenhq 
Screenshot 2020-10-27 at 13 48 09

styrelius:
I agree with @gustaveen that the date picker should be moved to the top, since it has to do with the insurance you're signing up for it's in my mind reasonable to have it grouped with the type of insurance and price. Also,  yes, a label would be neat!

virrios:
I agree, I was confused what the date was for! Should this change be part of this pr or are we handling it as an unrelated issue?

styrelius:
It's probably good to have that as another issue right?

virrios:
I agree, I think that makes the most sense

PR #387, opened 2020-10-26
[HVG-248] Chore / add react hooks ESLint plugin

styrelius:
### What?

Add the `eslint-plugin-react-hooks` plugin in order for us to use hooks in the way they say you should 😇

1. Install the package and add it (plus some rule declarations) to `.eslintrc.js`.
2. Make minor linting fixes, like just adding dependencies to `useEffect` dependency arrays or moving functions inside of `useEffect`'s.
3. Make pretty major changes to some files where there wasn't an easy fix available. All of these changes are committed one file at a time, so it's easy to revert if it would be discovered that some file is completely mutilated and non-functioning.

**Please note** that I was unable to check if `AdyenCheckout.tsx` broke in some way from my changes, since I couldn't get the test signing with ZignSec to work! Nor did I succeed in rendering that component in some other way. Would be highly appreciated if someone could take a close look at that particular file (aslo we need to look into what's wrong with the test signing). 

Kind of the same problem with `TrustlyModal.tsx`. Regardless of which branch I'm in, after signing with BankID Trustly is just loading for eternity so I can't know whether I messed something up. Though in this case it should be fine since the change was so minor.

<br>

### Why?

It's probably not a good idea to use hooks in another way than they are intended to. 😅

### DIFF
--- a/src/client/pages/ConnectPayment/components/AdyenCheckout.tsx
+++ b/src/client/pages/ConnectPayment/components/AdyenCheckout.tsx
@@ -143,9 +143,31 @@ export const AdyenCheckout: React.FC<Props> = ({ onSuccess }) => {
       history,
       onSuccess,
     }).mount(adyenCheckoutRef.current)
-  }, [paymentMethodsResponse, adyenLoaded])
+  }, [
+    currentLocale,
+    paymentMethodsResponse,
+    tokenizePaymentMutation,
+    submitAdditionalPaymentDetails,
+    history,
+    onSuccess,
+    adyenLoaded,
+  ])
 
-  useEffect(mountAdyenJs(setAdyenLoaded), [])
+  useEffect(() => {
+    const script = document.createElement('script')
+
+    script.src =
+      'https://checkoutshopper-live.adyen.com/checkoutshopper/sdk/3.6.1/adyen.js'
+    script.integrity =
+      'sha384-hUb/CFxzLJZWUbDBmQfccbVjE3LFxAx3Wt4O37edYVLZmNhcmVUyYLgn6kWk3Hz+'
+    script.crossOrigin = 'anonymous'
+    script.id = 'adyen-script'
+    script.onload = () => setAdyenLoaded(true)
+    document.body.append(script)
+    return () => {
+      document.body.removeChild(script)
+    }
+  }, [])
 
   useEffect(mountAdyenCss, [])
 
@@ -314,21 +336,6 @@ const createAdyenCheckout = ({
   return adyenCheckout.create('dropin')
 }
 
-const mountAdyenJs = (setAdyenLoaded: (adyenLoaded: boolean) => void) => () => {
-  const script = document.createElement('script')
-
-  script.src =
-    'https://checkoutshopper-live.adyen.com/checkoutshopper/sdk/3.6.1/adyen.js'
-  script.integrity =
-    'sha384-hUb/CFxzLJZWUbDBmQfccbVjE3LFxAx3Wt4O37edYVLZmNhcmVUyYLgn6kWk3Hz+'
-  script.crossOrigin = 'anonymous'
-  script.id = 'adyen-script'
-  script.onload = () => setAdyenLoaded(true)
-  document.body.append(script)
-  return () => {
-    document.body.removeChild(script)
-  }
-}
 const mountAdyenCss = () => {
   const link = document.createElement('link')
 
--- a/src/client/pages/ConnectPayment/components/TrustlyModal.tsx
+++ b/src/client/pages/ConnectPayment/components/TrustlyModal.tsx
@@ -53,7 +53,7 @@ export const TrustlyModal: React.FC<Props> = ({
     if (isSuccess) {
       onSuccess()
     }
-  }, [isSuccess])
+  }, [isSuccess, onSuccess])
 
   useEffect(() => {
     const handler = (e: MessageEvent) => {
--- a/src/client/pages/Embark/index.tsx
+++ b/src/client/pages/Embark/index.tsx
@@ -72,20 +72,19 @@ const Embark: React.FunctionComponent<EmbarkProps> = (props) => {
     embarkPassageId: string
     embarkPassageHistory: any
   }>()
+
+  const locationState = history.location.state
+
   const {
     reducer: [state, dispatch],
     goTo,
   } = useEmbark(() => {
-    if (
-      history.location.state &&
-      props.name === history.location.state.embarkPassageName
-    ) {
+    if (locationState && props.name === locationState.embarkPassageName) {
       return {
-        history: history.location.state.embarkPassageHistory || [
+        history: locationState.embarkPassageHistory || [
           props.data.startPassage,
         ],
-        passageId:
-          history.location.state.embarkPassageId || props.data.startPassage,
+        passageId: locationState.embarkPassageId || props.data.startPassage,
         data: props.data,
       }
     }
@@ -121,7 +120,7 @@ const Embark: React.FunctionComponent<EmbarkProps> = (props) => {
         embarkPassageName: props.name,
       })
     }
-  }, [currentPassage.id])
+  }, [currentPassage, history, state, props.baseUrl, props.name])
 
   const variation = useVariation()
 
--- a/src/client/pages/Embark/index.tsx
+++ b/src/client/pages/Embark/index.tsx
@@ -120,7 +120,16 @@ const Embark: React.FunctionComponent<EmbarkProps> = (props) => {
         embarkPassageName: props.name,
       })
     }
-  }, [currentPassage, history, state, props.baseUrl, props.name])
+  }, [
+    currentPassage.id,
+    currentPassage.url,
+    currentPassage.api,
+    currentPassage.externalRedirect,
+    history,
+    state,
+    props.baseUrl,
+    props.name,
+  ])
 
   const variation = useVariation()
 
--- a/src/client/pages/Forever/components/Intro/IntroStories.tsx
+++ b/src/client/pages/Forever/components/Intro/IntroStories.tsx
@@ -124,7 +124,7 @@ interface IntroProps extends RouteComponentProps {
   referrerName: string
 }
 
-const incrementPage = (page: number, pages: JSX.Element[]) => {
+const getNextPageNumber = (page: number, pages: JSX.Element[]) => {
   if (page === pages.length) {
     return page
   }
@@ -163,7 +163,7 @@ const IntroStoriesComponent: React.FC<IntroProps> = ({
 
   useEffect(() => {
     const autoPaginationTimeout = window.setTimeout(() => {
-      setPage(incrementPage(page, pages))
+      setPage(getNextPageNumber(page, pages))
     }, 4000)
 
     return () => {
@@ -216,7 +216,7 @@ const IntroStoriesComponent: React.FC<IntroProps> = ({
         offset="30%"
         onClick={(e) => {
           e.preventDefault()
-          incrementPage(page, pages)
+          getNextPageNumber(page, pages)
         }}
       />
       <SkipButton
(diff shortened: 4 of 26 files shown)


src/client/pages/ConnectPayment/components/AdyenCheckout.tsx
```diff
@@ -143,9 +143,31 @@ export const AdyenCheckout: React.FC<Props> = ({ onSuccess }) => {
...
       onSuccess,
     }).mount(adyenCheckoutRef.current)
-  }, [paymentMethodsResponse, adyenLoaded])
+  }, [
+    currentLocale,
+    paymentMethodsResponse,
+    tokenizePaymentMutation,
+    submitAdditionalPaymentDetails,
+    history,
+    onSuccess,
```
palmenhq, line 152: i'm unsure how this'll work. Since `onSuccess` for example probably will be a new function on every render idk how we should  treat that. Because `useEffect` checks object identity as far as i can remember this'll probably be triggered on every render of the parent component then, right? Meaning adyen will mount multiple times. That is probably not desired behaviour. Maybe that's resolvable but unless i remember the hooks api incorrectly this might have to be an exception to the rule, right?

src/client/pages/ConnectPayment/components/TrustlyModal.tsx
```diff
@@ -53,7 +53,7 @@ export const TrustlyModal: React.FC<Props> = ({
     if (isSuccess) {
       onSuccess()
     }
-  }, [isSuccess])
+  }, [isSuccess, onSuccess])
```
palmenhq, line 56: same here, i'm not sure `onSuccess` should be retrigged if that function changes. I don't think we should trust eslint blindly here

src/client/pages/Embark/index.tsx
```diff
@@ -121,7 +120,7 @@ const Embark: React.FunctionComponent<EmbarkProps> = (props) => {
         embarkPassageName: props.name,
       })
     }
-  }, [currentPassage.id])
+  }, [currentPassage, history, state, props.baseUrl, props.name])
```
palmenhq, line 123: why do we remove `currentPassage.id` here though? What if the id changes but not the `currentPassage`? Then `useEffect` wont detect that change

src/client/pages/Forever/components/Intro/IntroStories.tsx
```diff
@@ -124,63 +124,61 @@ interface IntroProps extends RouteComponentProps {
   referrerName: string
 }
 
-export const IntroStoriesComponent: React.FC<IntroProps> = ({
+const incrementPage = (page: number, pages: JSX.Element[]) => {
```
palmenhq, line 127: This name imo indicates side effects of the function while it's actually pure. Maybe we should name it something like `getNextPageNumber` or something?

src/client/pages/OfferNew/Introduction/Sidebar/StickyBottomSidebar.tsx
```diff
@@ -41,32 +41,22 @@ const Cta = styled(Button)`
...
+    if (isVisible) {
+      setDisplayNone(false)
+      setTimeout(() => setReallyIsVisible(true), 20)
+    } else {
+      setReallyIsVisible(false)
+      setTimeout(() => {
+        setDisplayNone(true)
+      }, 500)
+    }
+  }, [isVisible])
```
palmenhq, line 59: I dislike removing named functions like this. Now i don't understand why we do this little dance, before i understood that this was some kind of hack to fix some css-specific issue probably.

src/client/pages/OfferNew/Introduction/Sidebar/index.tsx
```diff
@@ -193,7 +199,7 @@ export const Sidebar = React.forwardRef<HTMLDivElement, Props>(
           refetchAll(),
         )
       }
-    }, [])
+    }, [redeemCode, redeemedCampaigns, refetchAll])
```
palmenhq, line 202: won't this retrigger once we change the redeemed campaigns now? Meaning we might get some kind of infinite loop or?

palmenhq (changes requested):
If i read the code correctly i think we might be trusting this plugin too blindly and are changing behavior based on sometimes misleading linting rules (mostly very good but in some cases misleading). I think [the hooks faq](https://reactjs.org/docs/hooks-faq.html#is-it-safe-to-omit-functions-from-the-list-of-dependencies) might be a good read to begin with. For example, i think we should utilise `useCallback` more. Imo this might be harder than anticipated and i think we should leave this task to rest a bit until we can spend time fixing these things properly. What's your thoughts on that?

src/client/pages/Forever/components/Intro/IntroStories.tsx
styrelius: That's a very good point! Will change it!

gustaveen:
I think this is a good push for sure, but I alos see some risk with doing it in a big bang release. I remember we did this at Tink and ended up breaking some stuff. So we did it incrementally instead, switching on the lint rule temporarily and fixing one issue at the time to be able to test and deploy each change. 

Then again I did it all in once in web-next 😁 But the risk felt lower and wasn't affecting that much critical functionality.

src/client/pages/ConnectPayment/components/AdyenCheckout.tsx
styrelius: Since you feel unsure about this we should of course wait with doing anything with this component until I (or someone else) can actually test these changes. Thank you, that was my point with splitting the changes up in several commits and asking for a closer check on this particular file! 🙏

src/client/pages/Embark/index.tsx
styrelius: I changed it since the `useEffect` also uses `currentPassage.url`, `currentPassage.api` and `currentPassage.externalRedirect` and instead of adding all of them to the dependency array I thought I could just added `currentPassage`. That might be a mistake though since I guess React will think `currentPassage` is new on every render since it's an object, and run the effect all the time.

styrelius: With that said, I'm pretty sure I checked that the number of renders did not change with this modification. I mean, I've of course manually tested every component I've made changes to, to see that behaviour didn't change.

src/client/pages/OfferNew/Introduction/Sidebar/StickyBottomSidebar.tsx
styrelius: I'm not sure I understand what you mean - you don't understand what it does now that it doesn't have a name? (Because I didn't understand it with the name either)

src/client/pages/Embark/index.tsx
styrelius: I'm changing it to have `currentPassage.id`, `currentPassage.url`, `currentPassage.api` and `currentPassage.externalRedirect` in the dependency array instead of just `currentPassage` (though I'm not really able to detect any difference in behaviour either way, but I might of course have missed something).

src/client/pages/OfferNew/Introduction/Sidebar/index.tsx
styrelius: No infinite loops when I tested it but maybe I'm missing something? Maybe there's something I don't know about `redeemedCampaigns`?

src/client/pages/OfferNew/Introduction/Sidebar/StickyBottomSidebar.tsx
palmenhq: Yes exactly :) without reading the whole function you can still kind of make out that it has to do with some kind of styling issue right? The naming gives this functionality kind of an explanation/context/purpose and I think named functions should be preferred to for example comments - comments would give the same benefits ish but have a high risk of becoming stale or outdated once you start changing the functionality later (especially since theyre commonly not on the same line meaning they’re invisible in git diffs and PRs, for example. Or at least that’s my experience. Please challenge it if you want!

palmenhq:
(Coming back to this I realize the ☝️ sounded a bit harsh maybe, not my intention 😅 sorry if it came across like that.)  Really good work on this! I just worry whether making all these changes is right timing wise. I think Gustav has a valid point, rather we should make this in smaller batches. The risk of unintentionally breaking something is high with so many changes to so many moving parts all at once. That way we can also take the time to really consider and discuss around each case. I have a feeling (which might be totally off though) that these rules might not always “have” to be applied, depending on context. (after having read the part of the react docs posted earlier)

src/client/pages/OfferNew/Introduction/Sidebar/StickyBottomSidebar.tsx
styrelius: I generally don't like comments in code either and that's not what I'm suggesting. Maybe a better solution generally is to declare _and_ call functions inside `useEffect` if that's the only place where they're used (like they recommend in the docs) instead of just doing stuff inside the `useEffect` (which also works and is clear enough in many cases IMO unless there are very complicated things going on).
That being said - this is not a good example of a self-explanatory function name since it's still such a strange component and it's impossible at least for me to understand it's purpose from just the name (or the implementation).
Also, this was the original linting warning and I'm not sure what a better solution would have been than doing as the linter suggests:
image

styrelius:
I wouldn't really agree that this is too blindly trusting this plugin, I’ve also spent a fair amount of time in the hooks docs, and regarding `useCallback` this article made a good case for not using it unless it _really_ is necessary, at least in my view: [https://kentcdodds.com/blog/usememo-and-usecallback](https://kentcdodds.com/blog/usememo-and-usecallback). The way I read the faq on omitting functions from the list of dependencies is you should probably not do it.

That being said, I think Gustav has a point, it’s probably a much better idea to do this incrementally. And we should of course not deploy any changes to `AdyenCheckout.tsx` before it’s possible for me (or anyone else) to check that it doesn’t break.

I think we could benefit from a discussion on code style more generally, since a few of the comments here actually has more to do with that than whether things will break. I also think we should discuss a strategy for Pull Requests and how we do reviews.

And yes @palmenhq I felt it was kind of rude to imply that this is something I just carelessly slapped together without reading docs or checking for infinite loops, which is of course not true. But thank you for acknowledging that and many of your points are of course valid.

styrelius:
And I know I'm not always the best at sounding nice, especially not in text. I'm trying to do better though!

src/client/pages/OfferNew/Introduction/Sidebar/index.tsx
styrelius: Looked a bit closer at it now and yes, it could have caused an infinite loop if I hadn't wrapped it in `useMemo` like the nice linter suggested to me :)

src/client/pages/ConnectPayment/components/TrustlyModal.tsx
styrelius: Yeah I guess `onSuccess` is "new" on every render but since the initial state of `isSuccess` is false there won't be any calls to `onSuccess` on re-render right? Or am I thinking about it wrong? Anyways, `onSuccess` is never called when I run this locally since Trustly iframe is blocked 😕  But the tests for `TrustlyModal` are fine and onSuccess is only called once there.

palmenhq:
Just for the record, in case anyone reads these comments this in retrospect:
We've discussed this a bit over phone instead just to not have any more misunderstandings and this was very inelegant from my end and that's ofc not ok - I feel like I keep telling people to think about their tone in PRs and that text can come through in other ways than intended but apparently I fail to live that rule by myself sometimes 😅 @styrelius has done some damn fine work with this. But even though it's very well crafted and well tested tested there might be cases we missed or just didn't know of, since we don't have great automated test coverage. So let's use this branch but transfer parts of it in small chunks so in case something would break, so that in the event it breaks we can fix that particular isssue before deploying the next thing. Also - great of you Ellen to raise this so it surfaces and we can resolve it 😊

 <span style="font-size: 40px" >and<span>

styrelius:
(And for the record, I know I should probably not call people rude in PR's. Will not do that again, sorry 💐)

PR #389, opened 2020-10-28
Chore / react hooks ESLint plugin / install and setup

styrelius:
### What?

First step towards using the `eslint-plugin-react-hooks`, in order for us to use hooks the way they say you should 😇

1) Install `eslint-plugin-react-hooks` and add some config to `.eslintrc.js`.
2) Modify dependency array of `useEffect` in `textKeys.tsx` since it caused ESLint to crash completely.

<br>

### Why?

It's probably not a good idea to use hooks in another way than they are intended to. 😅

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -11,6 +11,7 @@ module.exports = {
     'plugin:@typescript-eslint/recommended',
     'prettier/@typescript-eslint',
     'plugin:import/typescript',
+    'plugin:react-hooks/recommended',
     'plugin:prettier/recommended',
   ],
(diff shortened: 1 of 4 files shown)


styrelius:
This is a way for us to start using `eslint-plugin-react-hooks` in a controlled step-by-step way. All changes will be added to the [chore/react-hooks-eslint-plugin/main branch](https://github.com/HedvigInsurance/web-onboarding/tree/chore/react-hooks-eslint-plugin/main) through small PR's, before that branch is merge with master.

palmenhq (approved):
amaze!

PR #390, opened 2020-10-29
Added ReactNativeWebView for avy integration.

virrios:
Added variation check that only if it's in AVY the postMessage will be sent, for both web and native.

Created a AVY window type that extends window, however was unsure were to place it. 
Feedback will be appreciated

### DIFF
--- a/src/client/pages/ConnectPayment/sections/ConnectPayment.tsx
+++ b/src/client/pages/ConnectPayment/sections/ConnectPayment.tsx
@@ -111,7 +111,7 @@ export const ConnectPaymentPage: React.FC = () => {
   const variation = useVariation()
 
   const onSuccess = () => {
-    if (Variation.AVY === variation) {
+    if (variation === Variation.AVY) {
       const message = JSON.stringify({ event: 'PaymentConnected' })
       const awyWindow = window as AWYWindow
       awyWindow.frames.parent.postMessage(message, '*')
--- a/src/client/pages/ConnectPayment/sections/ConnectPayment.tsx
+++ b/src/client/pages/ConnectPayment/sections/ConnectPayment.tsx
@@ -5,6 +5,8 @@ import { Market, useMarket } from 'components/utils/CurrentLocale'
 import { AdyenCheckout } from 'pages/ConnectPayment/components/AdyenCheckout'
 import { TrustlyCheckout } from 'pages/ConnectPayment/components/TrustlyCheckout'
 import { useTextKeys } from 'utils/textKeys'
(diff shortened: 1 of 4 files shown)


src/client/pages/ConnectPayment/sections/ConnectPayment.tsx
```diff
@@ -106,10 +108,17 @@ const ConnectPaymentImage = styled('img')({
 export const ConnectPaymentPage: React.FC = () => {
   const textKeys = useTextKeys()
   const market = useMarket()
+  const variation = useVariation()
 
   const onSuccess = () => {
-    const message = JSON.stringify({ event: 'PaymentConnected' })
-    window.frames.parent.postMessage(message, '*')
+    if (Variation.AVY === variation) {
```
palmenhq, line 114: Are these kind of "yoda conditions" intentional? (Yes, that's the name, see [wikipedia](https://en.wikipedia.org/wiki/Yoda_conditions) if it sounds unfamiliar 😄) Personally I'm not sure i'm a fan, especially in this flavor of JS since everything is kind of constant anyways.

src/client/utils/tracking/signing.ts
```diff
@@ -1,9 +1,18 @@
 import { MemberQuery } from 'data/graphql'
 
+export interface AWYWindow extends Window {
```
palmenhq, line 3: Typo in `Avy` here :)

src/client/utils/tracking/signing.ts
```diff
@@ -1,9 +1,18 @@
...
+  ReactNativeWebView?: {
+    postMessage: (message: string, targetOrigin: string) => void
```
palmenhq, line 5: I'm not sure this is still the case but judging from [this github issue comment](https://github.com/react-native-webview/react-native-webview/issues/323#issuecomment-466059732) we may want to omit `targetOrigin` here. Maybe we could reach out to Avy to see if that's true?

palmenhq (changes requested):
Awesome, just some nit :D

src/client/pages/ConnectPayment/sections/ConnectPayment.tsx
virrios: change this, I will!

src/client/utils/tracking/signing.ts
virrios: yes its seems like it should be omitted, good catch!

styrelius: 👍

palmenhq (approved):
awesome!

PR #391, opened 2020-10-29
HVG-98 Connecting zign-sec to web on-boarding

virrios:
Added DanishBankId fields to sign mutation and updated Danish initial form for debugger to use danish test user 2.
The reason for using test user 2 is that test user1 generated some codes not present in the  table for test codes.
Also added the correct date format for Denmark to fix error for Denmark offer.


palmenhq (approved):
BOOM! Let's ship it 🚢

styrelius (approved):
🥳 Great job with this one @virrios !

PR #392, opened 2020-10-29
Chore / React hooks ESLint plugin / easy fixes

styrelius:
### What?

Second step towards using the `eslint-plugin-react-hooks`, in order for us to use hooks the way they say you should 😇

1) Fix linting warnings in components where fixing them wasn't complicated and where it was easy to manually test the number of re-renders etc. Easy fixes means adding dependencies to `useEffect` dependency array (without causing a lot of new re-renders), or moving some logic inside of `useEffect`. 

<br>

### DIFF
--- a/src/client/pages/Embark/index.tsx
+++ b/src/client/pages/Embark/index.tsx
@@ -72,20 +72,19 @@ const Embark: React.FunctionComponent<EmbarkProps> = (props) => {
     embarkPassageId: string
     embarkPassageHistory: any
   }>()
+
+  const locationState = history.location.state
+
   const {
     reducer: [state, dispatch],
     goTo,
   } = useEmbark(() => {
-    if (
-      history.location.state &&
-      props.name === history.location.state.embarkPassageName
-    ) {
+    if (locationState && props.name === locationState.embarkPassageName) {
       return {
-        history: history.location.state.embarkPassageHistory || [
+        history: locationState.embarkPassageHistory || [
           props.data.startPassage,
         ],
-        passageId:
-          history.location.state.embarkPassageId || props.data.startPassage,
+        passageId: locationState.embarkPassageId || props.data.startPassage,
         data: props.data,
       }
     }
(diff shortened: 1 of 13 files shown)


src/client/utils/hooks/useDocumentScroll.tsx
```diff
@@ -8,25 +8,26 @@ export const useDocumentScroll = (
   }) => void,
 ) => {
   const [, setScrollTop] = React.useState(0)
```
gustaveen, line 10: Unrelated to this, but why isn't there a state variable here? 🙃

src/client/pages/Embark/index.tsx
```diff
@@ -121,7 +121,16 @@ const Embark: React.FunctionComponent<EmbarkProps> = (props) => {
...
-  }, [currentPassage.id])
+  }, [
+    currentPassage.id,
+    currentPassage.url,
+    currentPassage.api,
+    currentPassage.externalRedirect,
+    history,
+    state,
+    props.baseUrl,
+    props.name,
```
gustaveen, line 132: Feel free to omit in this PR, would be nice to deconstruct these 🤗

gustaveen (approved):
Nicely done! @styrelius 🥳

styrelius:
Hello! Please note that I've removed changes made to `PerilModal.tsx` since those changes was a bit messy due to destructuring props. These changes will instead be in a separate PR!

src/client/utils/hooks/useDocumentScroll.tsx
styrelius: Haha, I would also like to know that!

src/client/pages/Embark/index.tsx
styrelius: I agree with you but made the decision to not have those kind of changes in this PR, I think it makes it cleaner and easier to review if this is purely fixing linting stuff 😇

gustaveen: Word! 👌

PR #393, opened 2020-10-29
HVG-126 Enable adyen in Denmark

virrios:
The same origin key is used for both Norway and Denmark so no configuration was required, as we first thought.
Everything is solved in the backend.


palmenhq (approved):
amaze!

palmenhq:
Travis is just a pain atm but we know CI passes so lets merge anyways

PR #394, opened 2020-10-30
Fix / Add offer page link to debugger

styrelius:
### What?

Add "Go to offer page" `<LinkButton>` in `/Debugger/components/Offer.tsx`. 
This is rendered above the quote data if there's a created quote and it leads to the real offer page `.../new-member/offer`.

<br>

### Why?

It's annoying when you're using the debugger tool to create quotes that you have to go to the address bar and _type stuff_ in order to see your quote data on the real offer page.

<br>

### GIF

**Creating a quote with the debugger and then going to the `/offer` page in _one click_ 😀**

2020-10-30-go-to-offer-page-debugger

### DIFF
--- a/src/client/pages/Debugger/components/Offer.tsx
+++ b/src/client/pages/Debugger/components/Offer.tsx
@@ -1,5 +1,4 @@
 import React, { useEffect, useState } from 'react'
-import { useLocation } from 'react-router'
 import { Form, Formik, FormikProps } from 'formik'
 import { v4 as uuid } from 'uuid'
 import { colorsV3 } from '@hedviginsurance/brand'
@@ -102,8 +101,6 @@ export const Offer: React.FC<OfferProps> = ({ sessionToken }) => {
   const currentLocale = useCurrentLocale()
   const localeIsoCode = getLocaleIsoCode(currentLocale)
   const currentMarket = useMarket()
-  const location = useLocation()
-  console.log('🐐: location', location)
 
   const [quoteType, setQuoteType] = useState(
     quotesByMarket[currentMarket][0].value,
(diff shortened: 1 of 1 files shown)


src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -269,7 +273,20 @@ export const Offer: React.FC<OfferProps> = ({ sessionToken }) => {
               </Formik>
             </>
           )}
-          {data?.quote && <pre>{JSON.stringify(data, null, 2)}</pre>}
+          {data?.quote && (
+            <>
+              <LinkButton
+                background={colorsV3.gray100}
+                foreground={colorsV3.gray900}
+                to={location.pathname.replace('debugger', 'offer')}
```
palmenhq, line 281: clever! But imo we should "hard-code" the link target if we can. Something like ``to={`/${currentLocale}/new-member/offer`}``

i'm thinking for reasons such as
1) if we'd replace the route with something else, we can't just look for `*/new-member/offer`.
2) if i change the route of `debugger` i'd be surprised why  the link _to_ the offer page would stop working

palmenhq:
Amaze feature tho :D

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -100,6 +102,8 @@ export const Offer: React.FC<OfferProps> = ({ sessionToken }) => {
   const currentLocale = useCurrentLocale()
   const localeIsoCode = getLocaleIsoCode(currentLocale)
   const currentMarket = useMarket()
+  const location = useLocation()
+  console.log('🐐: location', location)
```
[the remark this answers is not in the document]
styrelius: Haha OOPS! Did this pretty fast 😳

src/client/pages/Debugger/components/Offer.tsx
styrelius: You're right. I'll change it!

palmenhq (approved):

PR #395, opened 2020-11-02
HVG-286 Update partner script before november 9

virrios:
Removed the old student tracking script and added the new one, using Tune tracking.
   Added the external urls to whitelist.


styrelius (dismissed):
Nice! Let's merge so we can check that it actually works! 🙃

src/client/utils/tracking/tracking.ts
```diff
@@ -116,14 +116,20 @@ export const adtraction = (
...
+  tdl.init("https://aff.addreax.com/")
+  tdl.convert(
+  {
+  'adv_sub': '${memberId}'
+  }
+  )`
+  const scriptTag = window.document.createElement('script')
+  scriptTag.nonce = window.document.querySelector<HTMLScriptElement>(
+    '[nonce]',
+  )?.nonce
```
styrelius, line 130: Just a question though - is this the way you should add the `nonce` attribute? Asking since I'm not really sure myself.

palmenhq: I think you have a point here ellen! This worked previously but a) it does not any more because of security updates to the browsers and b) we allow unsafe-inline nowadays because the tracking scripts are just too much a hassle otherwise :(

so to conclude, i think we don't need this line at all now

palmenhq (dismissed):
asum!

src/server/middleware/helmet.ts
```diff
@@ -60,6 +60,8 @@ const defaultSrc = [
   'https://api-js.mixpanel.com',
   'checkoutshopper-live.adyen.com',
   'checkoutshopper-test.adyen.com',
+  'aff.addreax.com',
+  'js.go2sdk.com/v2/tune.js',
```
palmenhq, line 64: Sorry i missed this before but these should only be domains and protocols. So `js.go2sdk.com` should be sufficient :)

gustaveen (approved):

styrelius (approved):
🎈

PR #397, opened 2020-11-02
HVG-310 Update the ssn after creating the quote

virrios:
Enable the editing of ssn for Denmark.

danskSsn
Added gif to verify if the date format is correct and this was the expected result of the ticket.


styrelius:
Looks good! ✨ The SSN placeholder is wrong though - but that is something that needs to be changed in Lokalise. I don't know if we should maybe do that before merging, so we don't forget about it?

virrios:
Yes! good catch Ellen, I forgot to ask this, but should it be the personal Identity or National Identity number?

styrelius:
> Yes! good catch Ellen, I forgot to ask this, but should it be the personal Identity or National Identity number?

Ah, you're talking about the label? I meant the placeholder "DDMMYYXXXXX" though - it's just supposed to be four "X" but it looks like we have five now, like in Norway.

Regarding the label - I'm not sure what that one should say. Maybe we should just call it "CPR number" since that seems to be what they call it. Or as Wikipedia says "Personal identification number". What do you think @palmenhq and @gustaveen ?

virrios:
ohh, I see, yes I think its using the en.json right now. And there does not exist en_DK right now, as you said. How do we go about creating the keys?

styrelius:
OK so it turns out I had misunderstood a bit how we use our Lokalise translations, but the SSN placeholder will be correct as soon as we start using the Danish translation (or English for the danish market) since the right format is now added in Lokalise. For now, it will be wrong though, since we're using the plain English one (the "default" values for the text keys). But that's fine in my opinion 😃

styrelius (approved):
👏 👏

palmenhq (approved):

PR #400, opened 2020-11-03
HVG-306 Hide Switcher flow from Danish web onboarding

virrios:
Just hide the switcher flow when the market is Denmark

### DIFF
--- a/src/client/pages/Debugger/components/Offer.tsx
+++ b/src/client/pages/Debugger/components/Offer.tsx
@@ -219,11 +219,6 @@ export const Offer: React.FC<OfferProps> = ({ sessionToken }) => {
                         {...props.getFieldProps('lastName')}
                       />
                       <InputField
-                        label="Current Insurer (optional)"
-                        placeholder=""
-                        {...props.getFieldProps('currentInsurer')}
-                      />
-                      <InputField
                         label="Birth date"
                         placeholder="2012-12-12"
                         {...props.getFieldProps('birthDate')}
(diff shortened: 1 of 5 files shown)


styrelius (commented):
Nice! 🌸 I guess we should also remove the "Current Insurer" input from the debugger though? I think that could be within the scope for this ticket.

virrios:
> Nice! 🌸 I guess we should also remove the "Current Insurer" input from the debugger though? I think that could be within the scope for this ticket.

virrios:
I saw that it was optional in the debugger. It was not obvious for me that it should be removed from the ticket description so If you think it should be removed I will do it :D

palmenhq (dismissed):
🥳

styrelius:
> I saw that it was optional in the debugger. It was not obvious for me that it should be removed from the ticket description so If you think it should be removed I will do it :D

No, I know - the ticket didn't say so! But since we don't support the switching for Danish quotes you will get an error if you add something to that input. So I think we should just get rid of it!

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -218,21 +218,18 @@ export const Offer: React.FC<OfferProps> = ({ sessionToken }) => {
...
-                        placeholder=""
-                        {...props.getFieldProps('ssn')}
-                      />
+                      {currentMarket === Market.Se && (
+                        <InputField
+                          label="ssn"
+                          placeholder=""
+                          {...props.getFieldProps('ssn')}
+                        />
+                      )}
```
styrelius, line 232: Wasn't the exact same change in a previous, already merged, PR though? 🤔

PR #402, opened 2020-11-04
HVG-321 Remove the ssn input field in the deubgger for DK and No

virrios:
Reason is to match the actual embark flow in the debugger.


palmenhq (approved):
👍

PR #403, opened 2020-11-04
Whitelist hotjar with wss protocol in CSPs

palmenhq:
Hopefully this'll resolve the issues with tracking in different browsers


styrelius (approved):
🤞

PR #404, opened 2020-11-04
Hvg 306 hide switcher flow from danish web onboarding

virrios:
Remove the currentInsurer option for denmark as well as some refactoring the quoteForms.


styrelius (approved):
Great! 🙌

PR #406, opened 2020-11-05
[HVG-324] Feature/onboarding USP

gustaveen:
## What
Adding USP and heading to onboarding startpage. Also added a new base component for svg icons that mimics how material-ui handles icons.

We should probably merge your translations first @styrelius :) 

Screenshot 2020-11-05 at 12 15 18


src/client/pages/Landing/Landing.tsx
```diff
@@ -13,31 +13,113 @@ import {
...
+
+  @media (min-width: 1020px) {
+    span {
+      margin-right: 3rem;
+    }
   }
 `
 
-const Container = styled.div`
+const CardContainer = styled.div`
```
palmenhq, line 120: this is nit, and I see this was the case before also but should we decide on either "container" or "wrapper" and not mix the two? Imo wrapper is better to not confuse it with a container of state/functionality?

palmenhq (commented):
awesome, just a question really

src/client/pages/Landing/Landing.tsx
gustaveen: I get your point :) Personally I'm not a fan of Wrapper and I prefer Container (as the name for a wrapping div for a component) and that's what I've seen and used in most projects. I see what you mean with react container components, however I don't feel that has been a source of confusion for me historically :) But I'm up for discussion!

styrelius: I actually also like the "container" more. I've seen people argue that container is more for when you have multiple things grouped and wrapper is to  wrap one or a few things. Personally I think of the wrapper as bigger than a container. Like if you need to add styling to an entire component or something. What I wanted to say with this is that I think it's fine to mix them if we do it consistently 😇

src/client/pages/Landing/Landing.tsx
```diff
@@ -181,8 +281,8 @@ const LandingPageCardsSe: React.FC<{
   return (
     <>
       <Card to={`/${language}/new-member/new`}>
-        <Headline>{textKeys.STARTPAGE_UNINSURED_HEADLINE()}</Headline>
-        <Paragraph>{textKeys.STARTPAGE_UNINSURED_BODY()}</Paragraph>
+        <CardHeadline>{textKeys.STARTPAGE_UNINSURED_HEADLINE()}</CardHeadline>
```
styrelius, line 284: Imo this file is a bit large and therefore hard to get an overview of - maybe some things, like for example the `LandingPageCardsSe` component could be in its own file? With that said, I don't think we have to do this now. Just something to maybe fix in the future.

styrelius (approved):
Looks really nice! ✨

src/client/pages/Landing/Landing.tsx
virrios: I think, I have used wrapper as a wrapper of one single element, and container to "contain" multiple things, hence I agree with Ellen that it's okay to use both but we should be consistent with how we used them.

src/client/pages/Landing/Landing.tsx
gustaveen: True that! 🙌 How do we feel about having separate files for styling btw? I'm used to it myself but either or works. And maybe just in cases where you end up with large files.
Pros: shorter files and clearer separation
Cons: more files and imports

virrios: I think in this case if we break out some of the components some styled components will follow, like for instance Headline, Preamle, Wrapper and UspContainer are only used for one part, that could be split out into a separate file.
 
If we split out the different LandingPageCards [se,no,dk] I think the common Styled components used by those should be in a separate file.

PR #411, opened 2020-11-09
Chore / React hooks ESLint plugin / SignLoading.tsx

styrelius:
### What?

Fix `eslint-plugin-react-hooks` warnings in `SignLoading.tsx` 😇

1. Fix `setTimeout` in `useEffect` so that `timer` state can be removed and hence is not needed as a dependency to this `useEffect`.
2. Add dependencies to other `useEffect` dependency array.

<br>

### DIFF
--- a/src/client/pages/SignLoading.tsx
+++ b/src/client/pages/SignLoading.tsx
@@ -32,7 +32,6 @@ const TextWrapper = styled('div')`
 `
 
 export const SignLoading: React.FC = () => {
-  const [timer, setTimer] = useState<number | null>(null)
   const [hasTakenLong, setHasTakenLong] = useState(false)
   const signStatusQuery = useSignStatusQuery({
     pollInterval: 1000,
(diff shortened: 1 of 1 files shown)


styrelius:
### Question
Everything works as before but I don't understand if this component is even mounted? Like no `console.log`'s are actually printed from it. And that's the case both with and without the changes suggested here 🤷‍♀️

palmenhq (approved):
Woah, so much better!

palmenhq:
@styrelius where do you put the console log?

styrelius:
> @styrelius where do you put the console log?

I just tried to console.log the `hasTakenLong` state. In the `SignLoading` component that is.

palmenhq:
@styrelius i have no clue actually 🤷  But if it works it works i guess..

styrelius:
@palmenhq Haha OK so we merge then? 😅

PR #412, opened 2020-11-10
Adding memberId to gtm script (only when eventName is signed_customer…

virrios:
… according to the spec)

  The memberId should only be sent when its signed_customer and memberId is defined.
  One suggestion is to send "MEMBER_ID_UNDEFINED" as member_id if the member_id is undefined
  when the event is signed_customer and not hide it.

### DIFF
--- a/src/client/utils/tracking/gtm.ts
+++ b/src/client/utils/tracking/gtm.ts
@@ -11,6 +11,7 @@ interface GTMOfferData {
   number_of_people: number
   insurance_price: number
   currency: string
+  member_id?: string
 }
 
 export interface DataLayerObject {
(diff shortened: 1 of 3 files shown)


src/client/utils/tracking/gtm.ts
```diff
@@ -38,6 +39,7 @@ export const trackOfferGTM = (
         number_of_people: offerData.person.householdSize,
         insurance_price: parseFloat(offerData.cost.monthlyNet.amount),
         currency: offerData.cost.monthlyNet.currency,
+        member_id: offerData.memberId,
```
gustaveen, line 42: Could we maybe do something like this, only passing memberId if it's defined? :)
```suggestion
        ...(offerData.memberId && { member_id: offerData.memberId }),
```

gustaveen (approved):
Looks good to me, let's try it!

gustaveen (approved):

PR #388, opened 2020-10-27
[HVG-88] Bypass dk intro screen and go to directly to embark for now

gustaveen:
Will ofc not show anything until the embark flow is merged

PR #414, opened 2020-11-12
DK Embark flow

palmenhq:
- Redirect from `/(dk|dk-en)/new-member` directly into the `contents` flow
- Fetch correct story from embark

We'll also need to update the giraffe schema to include an option to pass the `locale` so we can fetch either DK or EN, but let's do that in giraffe first and then here. Will do in a separate PR.

### DIFF
--- a/src/client/pages/Landing/Landing.tsx
+++ b/src/client/pages/Landing/Landing.tsx
@@ -3,6 +3,7 @@ import styled from '@emotion/styled'
 import { colorsV3 } from '@hedviginsurance/brand'
 import React from 'react'
 import Helmet from 'react-helmet-async'
+import { Redirect } from 'react-router'
 import { LinkButton } from 'components/buttons'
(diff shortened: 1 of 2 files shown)


gustaveen (approved):

styrelius:
I'm a bit confused - didn't @gustaveen already fix this with bypassing the intro screen and redirecting directly to the embark flow here: [https://github.com/HedvigInsurance/web-onboarding/pull/388](https://github.com/HedvigInsurance/web-onboarding/pull/388)

palmenhq:
@styrelius I thought so also but i think that PR only added the DK routes but left the intro screen empty rather than with a redirect 🙃

styrelius:
@palmenhq Ah, I see 👍

gustaveen:
@palmenhq is cleaning up my mess @styrelius  😄  ❤️

PR #415, opened 2020-11-12
Clean up environments

palmenhq:
✅ Make passing environment variables from the server to the client more smooth
✅ Add correct deps to `app.json` making it possible to create review apps again
✅ Disable DK flow for production
✅ Disable debugger for production

**🤔 but why not just change `process.env.NODE_ENV`?**

`NODE_ENV` is used during build time and is then substituted to its value in the build, so anywhere in the code `process.env.NODE_ENV` appears will actually say `"production"` once built. Since we use the same bundle for staging and production (and imo that's desirable, so the code we test is the code we deploy) and it's with the current setup not possible to make custom builds for staging and production we need to make that "decision" runtime instead. How we do this is we set environment variables on the server, and then forward them to the browser via the `window` object whenever the browser loads a page (you can see this little dance [here](https://github.com/HedvigInsurance/web-onboarding/pull/415/files#diff-e2488cd41c99b3beafe59ebbf87b9ac81dcbdd1e6f6e363cfd016905eea3e1cfR113-R116)).

### DIFF
--- a/src/client/pages/Embark/index.tsx
+++ b/src/client/pages/Embark/index.tsx
@@ -7,7 +7,7 @@ import {
   useEmbark,
 } from '@hedviginsurance/embark'
 import { AnimatePresence, motion } from 'framer-motion'
-import React from 'react'
+import React, { useEffect } from 'react'
 import { useHistory } from 'react-router'
 
 import { colorsV3 } from '@hedviginsurance/brand'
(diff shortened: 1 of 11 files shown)


src/server/page.ts
```diff
@@ -99,13 +110,10 @@ const template = (
...
     <script>
-      window.GIRAFFE_WS_ENDPOINT = ${JSON.stringify(GIRAFFE_WS_ENDPOINT)};
-      window.GIRAFFE_ENDPOINT = ${JSON.stringify(GIRAFFE_ENDPOINT)};
-      window.CONTENT_SERVICE_ENDPOINT = ${JSON.stringify(
-        CONTENT_SERVICE_ENDPOINT,
-      )};
-      window.ADYEN_ORIGIN_KEY = ${JSON.stringify(ADYEN_ORIGIN_KEY)};
-      window.ADYEN_ENVIRONMENT = ${JSON.stringify(ADYEN_ENVIRONMENT)};
+      Object.defineProperty(window, 'hedvigClientConfig', {
+        value: Object.freeze(${JSON.stringify(clientConfig)}),
```
gustaveen, line 114: Oh nice, didn't know about `freeze` 🧊

src/client/pages/Embark/index.tsx
```diff
@@ -251,6 +251,20 @@ export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
     }
   }, [props.name])
 
+  const isDenmarkInProduction =
```
gustaveen, line 254: Was a bit confused by the name at first, but can't come up with a better one myself 😆

gustaveen (approved):
awsm

PR #416, opened 2020-11-12
Webpack 5

palmenhq:
Because we need to upgrade at some point 👍  as an attempt to optimize the build performance i'm trying to upgrade webpack to see if we get a difference in codefresh. the webpack 5 release focuses on performance and while not shipping with many new features "many refactorings have been done to up-level the architecture and create a good base for future features (and current features)" (as per [the release note](https://webpack.js.org/blog/2020-10-10-webpack-5-release/)).

I don't se high risks with this upgrade but some features could break if i missed something while testing.


gustaveen (approved):
Looks good to me! 🙌

gustaveen:
IMO it would be nice with a short description on what potential changes version 5 comes with that have implications for us, could be as easy as a link to a changelog, a blog post or something like that. I mean in general for larger upgrades of packages etc. I think that could speed up reviews when you don't have to search for it yourself. Thoughts on that? @palmenhq @styrelius @virrios

palmenhq:
@gustaveen yes true, like the reason behind it and possible consquences. Good push! I'll improve for next time 👍

styrelius:
@gustaveen I agree, that would be 👌 !

palmenhq:
@gustaveen @styrelius description updated 👍

PR #423, opened 2020-11-18
Improve error reporting to sentry

palmenhq:
## What?

Add more places where we send errors to sentry

## Why?

So we can more easily find errors if they occur, also for example if it would be connected to the backend.


vonElfvin (approved):
Awesome

PR #425, opened 2020-11-19
Fix/hide intercom button in mobile checkout

gustaveen:
## What?
Hide Intercom button in mobile checkout (screens under 640px in width)

## Why?
It currently covers part of the sign button and GDPR content

Kapture 2020-11-19 at 15 48 09

[HVG-438]: https://hedvig.atlassian.net/browse/HVG-438

### DIFF
--- a/src/client/pages/OfferNew/index.tsx
+++ b/src/client/pages/OfferNew/index.tsx
@@ -2,6 +2,7 @@ import { History } from 'history'
 import { SemanticEvents } from 'quepasa'
 import React from 'react'
 import { Redirect, useHistory, useRouteMatch } from 'react-router'
+import { useMediaQuery } from 'react-responsive'
 import { LoadingPage } from 'components/LoadingPage'
 import { TopBar } from 'components/TopBar'
 import {
@@ -61,6 +62,7 @@ export const OfferNew: React.FC = () => {
     '/:locale(se-en|se|no-en|no|dk-en|dk)/new-member/sign',
   )
   const toggleCheckout = createToggleCheckout(history, currentLocale)
+  const isMobile = useMediaQuery({ maxWidth: 640 })
 
   if (quoteIds.length === 0) {
     return <Redirect to={`/${currentLocale}/new-member`} />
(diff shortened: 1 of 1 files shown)


gustaveen:
Worlds largest gif 😅

styrelius:
> Worlds largest gif 😅

Hahaha what happened? 😮

palmenhq (approved):
nice! Another way we could solve it Is also to place the sidebar with a higher z-index I think. But this looks good to me 👍

src/client/pages/OfferNew/index.tsx
```diff
@@ -61,6 +62,7 @@ export const OfferNew: React.FC = () => {
     '/:locale(se-en|se|no-en|no|dk-en|dk)/new-member/sign',
   )
   const toggleCheckout = createToggleCheckout(history, currentLocale)
+  const isMobile = useMediaQuery({ maxWidth: 640 })
```
styrelius, line 65: I'm not saying we should do this now, because I don't think we should, but imo we should probably have a hook or just some function that returns `isMobile` so we use the same measurements/breakpoints everywhere? Instead of passing `maxWidth` to this hook? Or?

src/client/pages/OfferNew/index.tsx
```diff
@@ -76,6 +78,13 @@ export const OfferNew: React.FC = () => {
     return <LoadingPage />
   }
 
+  const handleCheckoutToggle = (open: boolean) => {
```
styrelius, line 81:
```suggestion
  const handleCheckoutToggle = (isOpen: boolean) => {
```
🙃

gustaveen:
> nice! Another way we could solve it Is also to place the sidebar with a higher z-index I think. But this looks good to me 👍

Ah smart! I like using the most lo-fi solution👌  Only thing I wonder is how Intercom sets the z-index, it's currently set to `z-index: 2147483001;` So it might be a dynamic value 😅   And it might be that we want to trigger Intercom while the checkout is open (for example if you're unable to sign). So maybe start with this

styrelius:
And I know I was too slow but I had already written the comments lol

src/client/pages/OfferNew/index.tsx
gustaveen: I could rename this because I'm more interested in the breakpoint in this case, since we change the layout at 640px the checkout (and add more padding so the Intercom doesn't intersect), if it's mobile or not doesn't really matter :)

PR #427, opened 2020-11-23
Disable websocket transport for non-subscripion queries

palmenhq:
## What?

- Disable websocket transport for non-subscripion queries.
- To not have to deal with cors headers n stuff i created a small proxy inside the web-onboarding for proxying `/graphql` to giraffe.
- Additionally, increase the session timeout to 3 hours instead of 30 minutes.

## Why?

As per the [Apollo docs on subscriptions](https://www.apollographql.com/docs/react/data/subscriptions/#3-use-different-transports-for-different-operations) where they have clarified:

> Apollo Client should use your WebSocketLink for subscriptions, but it shouldn't use it for queries or mutations. For those operations, Apollo Client should use HTTP as usual

Apollo Client should use your WebSocketLink for subscriptions, but it shouldn't use it for queries or mutations. For those operations, Apollo Client should use HTTP as usual

#### Note to self
- set the `GIRAFFE_HOST` environment variable in heroku before merging this


src/shared/sessionStorage.ts
```diff
@@ -4,7 +4,7 @@ import { MinimalStorage } from 'utils/storage/MinimalStorage'
 
 export const SESSION_KEY = '_hvsession'
 export const KA_SESSION_KEY = '_hvsessionexpiry'
-const SESSION_EXPIRY_TIMEOUT = 30 * 60 * 1000
+const SESSION_EXPIRY_TIMEOUT = 3 * 60 * 60 * 1000
```
gustaveen, line 7: ⏳ 👌

gustaveen: How did we decide on 3 hours? (Just curious, feels like a good session length to me 👍 )

gustaveen (approved):
Makes sense to me, given that I'm an Apollo rookie 😄

src/shared/sessionStorage.ts
palmenhq: Haha, there's a very healthy process behind it 😅

1. "Wat, we only have 30 minutes? 🤨"
2. "We should probably change that 🤔"
3. "Lets 10x it and see what happens 🤷‍♂️"
3. "Lol that's not 10x, but 3 hours seems good so ima keep that 👊"

src/server/page.ts
```diff
@@ -33,7 +32,7 @@ const clientConfig: ClientConfig = {
   adyenEnvironment: ADYEN_ENVIRONMENT,
   adyenOriginKey: ADYEN_ORIGIN_KEY,
   contentServiceEndpoint: CONTENT_SERVICE_ENDPOINT,
-  giraffeEndpoint: GIRAFFE_ENDPOINT,
+  giraffeEndpoint: '/new-member/graphql',
```
virrios, line 35: Got stuck on this, but then I saw that you are going to update the heruko env. 👍

src/server/serverEntry.tsx
```diff
@@ -47,7 +48,17 @@ appLogger.info(
   }, with environment "${sentryConfig().environment}"`,
 )
 appLogger.info(
-  `Using giraffe at batchHttp:"${GIRAFFE_ENDPOINT}" ws:"${GIRAFFE_WS_ENDPOINT}" 🦒`,
+  `Using giraffe at http:"${GIRAFFE_HOST}" ws:"${GIRAFFE_WS_ENDPOINT}" 🦒`,
+)
+
+app.use(
+  proxy('/new-member/graphql', {
```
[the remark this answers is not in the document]
palmenhq: No, but otherwise the graphql server would live on a different domain = need proper cors headers n stuff

styrelius (approved):
SO nice! My GOD it's easier to navigate the requests in the Network tab now! 😅

PR #428, opened 2020-11-24
Fix remove double import from graphql/data in SwedishDetails

virrios:
## What?
Fix Removed double import from graphql/data in SwedishDetails

## Why?

Because someone had a quick trigger finger on merge..The linter should have caught this?

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -49,6 +49,7 @@ module.exports = {
         ],
       },
     ],
+    'import/no-duplicates': 'error',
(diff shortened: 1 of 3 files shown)


styrelius (dismissed):
Yes, the linter should catch these types of things but apparently we don't have a rule for that? You're _more than welcome_ to add such a rule! 😃

virrios:
ESLINT

styrelius (approved):
Wonderful! Ordning och reda pengar på freda osv 😀

PR #429, opened 2020-11-24
[HVG-439] Fix / Checkout sign button

styrelius:
## What?

Fix the design of the `Sign` component, mainly for it to no longer cover a big portion of the `Checkout` sidebar but also just to make it look nicer in general. 🎨 
Changes is this PR entails:
1) Moving the `SignDisclaimer` out of the `Sign` component and put it in its own file. Import it and render it as the last item in the sidebar.
2) Changing the design considerably, for several components, with Manges new designs for the `Checkout` sidebar in mind.
3) Modifying how the `Sign` component is rendered and positioned.
4) Adding logic to get the width of the rendered `CheckoutComponent` minus eventual scroll bar so that the width of `Sign` `Wrapper` can be adjusted to not cover said scrollbar. This is really only an issue with Edge and Explorer since Chrome actually updated this (i.e. removed their scrollbars) during the time I was working on this. If anyone thinks this adds unnecessary complexity for a limited case, please let me know and I will remove it! 
See `src/client/pages/OfferNew/Checkout/index.tsx` row 181-199 for this.
5) General refactoring of the components concerned - moving stuff where it's in my mind more logical to have it, renaming some things etc.

## Why?

This is part of a bigger remake of the signup process, which came about due to the realisation that many potential members/customers (mainly in Norway) are not able to sign up for Hedvig, sometimes because they might not understand how to do it since our UI might not be clear enough. The changes made here belong to those we concluded would be of the more low-hanging fruit type 🍎

_Referenced ticket: [HVG-439]_

## Screenshots and GIF's

**Norwegian checkout with bundled quotes**

2020-11-24-sign-no

2020-11-24-sign-no-combo

**Swedish checkout**

2020-11-24-sign-se

**Starting a BankID sign session in Swedish checkout**

2020-11-24-sign-se-sign-started

[HVG-439]: https://hedvig.atlassian.net/browse/HVG-439


src/client/pages/OfferNew/Checkout/Sign.tsx
```diff
@@ -1,69 +1,33 @@
...
+const Wrapper = styled('div')<WrapperProps>`
+  background: ${colorsV3.gray100};
+  box-shadow: 0 -1px 4px ${colorsV3.gray500};
   width: 100%;
-
-  @media (max-width: 600px) {
-    justify-content: center;
-  }
+  max-width: ${({ maxWidth }) => (maxWidth ? `${maxWidth}px` : '100%')};
+  padding: ${({ isDesktop }) => (isDesktop ? '1.5rem 2rem' : '1rem')};
```
gustaveen, line 22: Maybe it's just me, but can't we just use a good ol' media query instead of sending a js calculated value. *old man shakes fist at computer 😆

styrelius: Haha! 😍 No but my idea was that we'll _centralise_ the media queries later on so we always use the same values everywhere. Like we'll have a nice li'l hook that returns `isDesktop` for example. I guess I made it this way to make it easier to change that. And I actually like a named variable more than a regular media query - to me it's clearer what the media query is for then.

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -191,6 +183,26 @@ export const Checkout: React.FC<Props> = ({
...
+
+    setScrollWidth()
+
+    window.addEventListener('resize', setScrollWidth)
+
+    return () => {
+      window.removeEventListener('resize', setScrollWidth)
+    }
+  }, [outerWrapper.current?.scrollWidth])
+
```
gustaveen, line 205: Didn't fully understand what issue we solved with the scroll width 🤔

src/client/pages/OfferNew/index.tsx
```diff
@@ -80,9 +78,7 @@ export const OfferNew: React.FC = () => {
 
   const handleCheckoutToggle = (open: boolean) => {
     toggleCheckout(open)
-    if (isMobile) {
-      Intercom('update', { hide_default_launcher: open })
-    }
+    Intercom('update', { hide_default_launcher: open })
```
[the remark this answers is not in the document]
styrelius: I think you have a point with intercom being accessible from the checkout but unfortunately it's then placed on top of the sign button - since the button is now almost as wide as the sidebar. It looks really weird imo. But maybe we can fins a solution for it that looks better.

src/client/pages/OfferNew/Checkout/index.tsx
styrelius: I tried to explain that in point 4. in the description but I guess I didn't do such a good job then. If the browser has a fixed scrollbar (like Edge and Explorer have, and Google had until like last weekend lol) the `Sign` component will cover that too. Which is ugly. But as I wrote in the description, if you think is just unnecessary complexity that only solves a problem for the few users who use Edge or Explorer (I mean who does, really?) then I can remove it.

styrelius: OK after looking into it more I'm not even sure it matters for Edge so I'll remove it. It was a neat solution though imo, when it was needed for Chrome last week 😄 But I'll remove it!

src/client/pages/OfferNew/Checkout/Sign.tsx
gustaveen: I agree that is nice to have one value as reference. :) In this case we could do like we do in a few other places, set a const at the top of the file `const DESKTOP = 600` or and then we could use it both like so. Then we get the best of both worlds! 👯‍♂️
```
[`@media (min-width: ${DESKTOP}px)`]: {
...
}
....
const isDesktop = useMediaQuery({ minWidth: DESKTOP })
```

gustaveen: And I think it can be hard to use the same `isDesktop` for all components since the component layout will be more contextual. But I we could definitely go over the breakpoints we have and se if we can find a pattern to standardise a few of them 🙌

styrelius: But maybe you just meant the `maxWidth`? In that case, see comments below (TLDR: I'll remove it).

virrios: What I think could be a problem about useMediaqueries  is that you introduce styling into the component logic. I feel like you should not introduce it where it is maybe not needed (It absolutely has its place sometimes!).  I do however like that it becomes more declarative when the styles should be applied. 
In another project I think we went about it like this, declaring constants for MediaQueries pixelValues, but that might not really add anything to the readability more than just centralising pixelValues we use for mobile etc. 
`
...
@media (max-width: ${MOBILE_WIDTH}px){
`
This is more thoughts in general, I think the pr look super! 💪 


EDIT: Haha I am to slow @gustafguner ... 🐌

src/client/pages/OfferNew/Checkout/index.tsx
gustaveen: It was just med that couldn't visualise it my head 😁  It think it's awesome that you took that in to consideration, I'm all for not punishing users with old browsers. But if it's not needed for Edge I agree that we can remove it to not have that complexity 🥳

gustaveen:
Awesome work with this! 🥳

src/client/pages/OfferNew/Checkout/Sign.tsx
styrelius: No sure, you're right - this centralising can be done without using a hook, I was confusing things. I'm changing it to a media query (I still think we should use the same breakpoints everywhere, but this is something we should discuss later and not now).

styrelius: > EDIT: Haha I am to slow @gustafguner ... 🐌

@virrios  Wrong Gustav 😳  But I'm also too slow apparently haha! Anyways, YES valid points both of you!

src/client/pages/OfferNew/index.tsx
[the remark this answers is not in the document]
styrelius: OK so I guess we're going with just removing the Intercom button for now, since Mange said that's fine? @gustaveen

src/client/pages/OfferNew/Checkout/Sign.tsx
```diff
@@ -1,69 +1,35 @@
...
+  background: ${colorsV3.gray100};
+  box-shadow: 0 -1px 4px ${colorsV3.gray500};
   width: 100%;
-
-  @media (max-width: 600px) {
-    justify-content: center;
+  padding: 1rem;
+  position: sticky;
+  bottom: 0;
+  transition: visibility 3s;
```
palmenhq, line 25: i had no idea visibility was animatable! Nice one :D

src/client/pages/OfferNew/index.tsx
```diff
@@ -97,8 +93,6 @@ export const OfferNew: React.FC = () => {
     )
   }
 
-  // TODO: Add error component to render if there's no offerData
```
palmenhq, line 100: lol 😓

palmenhq (dismissed):
Looks awesome imo :) nice refactorings also

src/client/pages/OfferNew/Checkout/Sign.tsx
gustaveen: Me neither! But where do we set the visibility on this component though?

gustaveen:
I tried on my iPhone X now and the button is hidden and you can't scroll down to it 😬

ezgif com-gif-maker (1)

I tried quickly with this:
1. `Wrapper` in sign needs to be `position: fixed` to ensure it's visible
2. `Scrollwrapper` needs `-webkit-overflow-scrolling: touch;` to enable scroll on iPhone

It stills seem to scroll the page behind though, so might be something more we need to add back to the `Scrollwrapper` though

styrelius:
@gustaveen Yep I know, that's what I've been working on the entire day 😭
image

I also tried with `position: fixed` but as you say it will scroll the page behind the sidebar then. But hey! Since the sidebar covers the entire screen in mobile I can just set it to `fixed` when it's mobile! Maybe that will solve it

gustaveen:
I hear you @styrelius ! It's always a hustle to fix correct scroll behaviour in overlays and such on mobile so no worries 😁  

Regarding the scroll on the content below I think the safest way to fix that is to set `overflow: hidden` on body, that's how material-ui and similar ui libs usually fix it. So one way to add that (might ba a cleaner way but works in my book 🙂 ) is to do it add `document.body.style.overflow = 'hidden'` to `handleCheckoutToggle`

styrelius:
OMG I think I [fixed it](https://web-onboardi-fix-checko-fl6zq5.herokuapp.com/no/new-member)!!

@gustaveen Really clever (and kind of dumb in the absolutely best way) to set `overflow: hidden` on body to prevent scroll - I didn't think of that! Still, I didn't use it since we already have this `useScrollLock` hook but that might be a good refactoring for the future. The hook we have now is very specific to the `Checkout` sidebar case and we might want to do this for other components as well.

src/client/pages/OfferNew/index.tsx
styrelius: 😅

src/client/pages/OfferNew/Checkout/Sign.tsx
styrelius: Lol it's not - I just accidentally left it there after some experimenting that didn't pan out 😄

gustaveen (dismissed):
giphy (1)

gustaveen (approved):
Boom! 💥

PR #430, opened 2020-11-25
Set correct inputmode on ssn field

gustaveen:
## What?

Change input mode in ssn field to only display numbers in mobile.

`inputmode="numeric"` brings up the numeric keyboard on both Android and iOS. That covers most of our users but to be backwards compatible I added a pattern as well (I've got input from Sam and Oscar on this).

## Why?

Because it's currently not a good solution for mobile users.

### DIFF
--- a/src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
+++ b/src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
@@ -105,6 +105,8 @@ export const UserDetailsForm: React.FC<Props> = ({
           name="ssn"
           id="ssn"
           type="number"
+          inputMode="numeric"
+          pattern="[0-9]*"
           value={ssn}
           // errors={ssnError ? textKeys.SIGN_SSN_CHECK() : undefined} TODO error handling?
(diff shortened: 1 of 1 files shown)


src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
```diff
@@ -105,6 +105,7 @@ export const UserDetailsForm: React.FC<Props> = ({
           name="ssn"
           id="ssn"
           type="number"
+          inputMode="decimal"
```
styrelius, line 108: Just curious - why `"decimal"` and not `"numeric"`?

styrelius (approved):
🔢 📱

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
gustaveen: Updated now after checking a bit more, numeric and pattern in a combo 🙌

styrelius (approved):
🙌

PR #432, opened 2020-11-27
Omit credentials on graphql endpoint from apollo client

palmenhq:
## What?

Don't include  the `basic` authorization and cookie headers, as we set the `authorization` hedar directly from the client

## Why?

The browsers behave a bit differently apparently, but ios safari for example overrides the `authorization` header with the `basic` credentials, whereas chrome and firefox override the `basic` credentials with the manually set `authorization` header.


styrelius (approved):
🤖 🤖 🤖

PR #434, opened 2020-11-27
Hvg 101 edit bundle quotes norway

virrios:
[HVG-100] Modal for editing the bundle quotes for Norway

#### Summary
Added functionality to edit bundle quote in Details Modal.
* Change Prop in DetailsModal from offerQuote to offerData because we don't want a single but potentially multiple quotes.
* Added util functions to handle bundle quotes.
* Enabled details button and details modal in Norway
#### Details
* Get the main quote from offerData which is  NorwegianHomeContent| DanishHomeContent for quotes bundles, 
they control the underwriterHitLimit. 
* We need to make sure that the editQuote mutation for the main quote gets through without error and no underwriterHitLimit to do the rest of the quotes in the bundle.
* If the first mutation succeed, we can carry on doing the rest of the mutations.
* We also check that there are no errors or underwriterhitLimits in the subsequent editQuite mutations. (I found a backend bug this way)

(Sidenote think it will never get to the errors since we wrap everything in a try catch. But It was there before so I did not want to change that, without input. The underwriterhitLimits can occur since it is "successful" response and not an error )

#### OBS
(If there is an error in the subsequent calls it will potentially cause a missmatch of data points, however we decided to go about it this way and see if those errors occur in sentry)

## Why?
These changes were made to be able to edit quotes in norway and denmark in the future since they can have quote bundles.
The quotes in quote bundles share some attributes and they need to be updated in all quotes when a change is made.
This is not solved in the backend so we need to solve it here with multiple editQuote mutations.

_Referenced ticket [https://hedvig.atlassian.net/browse/HVG-101]()_

[HVG-100]: https://hedvig.atlassian.net/browse/HVG-100

### DIFF
--- a/src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/index.tsx
+++ b/src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/index.tsx
@@ -5,12 +5,20 @@ import React from 'react'
 import { Button } from 'components/buttons'
 import { Modal, ModalProps } from 'components/ModalNew'
 import { EditQuoteInput, useEditQuoteMutation } from 'data/graphql'
-import { OfferQuote } from 'pages/OfferNew/types'
+import { OfferData } from 'pages/OfferNew/types'
 import { useTextKeys } from 'utils/textKeys'
+import { isBundle } from '../../../utils'
 import {
   getFieldSchema,
   getInitialInputValues,
   getValidationSchema,
+  getUpdatedQuoteDetails,
+  getFormData,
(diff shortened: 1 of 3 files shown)


virrios:
#### Successful change in Details
SuccessfulDetailsHD

#### Success -> Send two requests
NetworkSuccessHD


#### Main mutation gets Limit hit so it should only do one mutation, that fails ( no subsequent mutations)
DetailsLIMITHD

palmenhq (approved):
gr8 work!

PR #435, opened 2020-11-27
Fix / Forever page success path

styrelius:
## What?

Make the url and redirect less flaky on /forever page, when the discount  code has been validated and redeemed.

This means adding a `getSuccessPath` function in `/Forever/useRedeemCode.ts` and writing some simple tests for that function, since it was actually pretty tricky to get it right and extremely annoying and undignified to test it manually.

## Why?

Currently if you add your discount code to the input instead of getting it straight from the url some pretty strange behaviour can occur, like the redirect straight out not working but also, if you try to refresh when nothing happens the word "intro" is added to the code input since the redirect is `.../[your-discount-code]/intro`. 

And no, this is not the most pressing issue right now but I made these changes two weeks ago when I had some extra time and it seems wasteful to not merge these changes now that they are already made I think. 👩‍💻

### Old strange behaviour that we will have no more of with this merge:
2020-11-27-forever-page-OLD


palmenhq (commented):
Awesome, and awesome with the test! Just thinking whether this is the direction to head in - i feel like we basically parse the url ourselves and i think that kinda logic should be done by libraries... what about using `matchPath` instead and `useCurrentLocale`? That way we could just construct the success url from the actual values, instead of patching and validating the base url we have. Thoughts?

styrelius:
> Awesome, and awesome with the test! Just thinking whether this is the direction to head in - i feel like we basically parse the url ourselves and i think that kinda logic should be done by libraries... what about using `matchPath` instead and `useCurrentLocale`? That way we could just construct the success url from the actual values, instead of patching and validating the base url we have. Thoughts?

Sure, you really have a point! I think I was maybe too focused on just patching up the thing we already had. 🩹

virrios:
I think this is a fix to the solution we had, so if we want to redo the base solution complete how we handle all this it might be a different ticket. Then we can take the time to think about how  we want to solve this, while having the problem fixed in prod.  Love the tests 🥇

palmenhq:
@virrios i don't agree here, i don't think we should merge code we already know is not a direction we want to head, unless the "right" solution is very big and we need to get a fix out faster than we can fix the it "the right way". I don't think that would be the case here. I'm happy to be challenged though :)

virrios:
I think you are 100% correct with:
"unless the "right" solution is very big and we need to get a fix out faster than we can fix the it "the right way". 
As I understood it we are already in the direction, I mean this code is refactoring and adding som additional checks to fix and issue. The severity of the issue might not warrant to merge the code if we decide that we want to redo how we solve this from the ground up.
I am not sure how much of an issue this is for the user so someone who knows that could maybe pitch in.

styrelius:
@virrios It's not a _huge_ issue for users I think, mostly it was annoying to me when I was working with fixing linting errors in these components. Still we shouldn't have this strange behaviour. And actually I won't have the time to refactor this now so maybe I'm of the opinion now that we should merge this since I think it's still a step in the right direction.

palmenhq (approved):
hm fair points. I still think we should take the time and do it properly, but choosing between having one less bug and not having one less bug i'd choose the former :)

PR #437, opened 2020-11-30
Added try catch to catch the iframe access origin error for trustly.

virrios:
Fix "Blocked a frame with origin  from accessing a cross-origin frame" for trustly,
showing up in Sentry

## What?
Bugbash Fix To stop showing "Blocked a frame with origin  from accessing a cross-origin frame" in Sentry
and instead catch it in the frontend.

Added try catch to wrap **actualHandleIframeLoad** 

## Why?
These changes were mad because the iframe shows up, and should not bloat the Sentry issue logs.

CatchError

### DIFF
--- a/src/client/pages/ConnectPayment/components/TrustlyModal.tsx
+++ b/src/client/pages/ConnectPayment/components/TrustlyModal.tsx
@@ -94,11 +94,16 @@ export const TrustlyModal: React.FC<Props> = ({
           ref={iframeRef}
           onLoad={async () => {
             const contentWindow = iframeRef.current?.contentWindow
-            await actualHandleIframeLoad(
-              setIsOpen,
-              setIsSuccess,
-              generateTrustlyUrl,
-            )(contentWindow)
+
+            try {
+              await actualHandleIframeLoad(
+                setIsOpen,
+                setIsSuccess,
+                generateTrustlyUrl,
+              )(contentWindow)
+            } catch (e) {
+              console.error(e.message)
+            }
           }}
         />
       )}


src/client/pages/ConnectPayment/components/TrustlyModal.tsx
```diff
@@ -94,11 +94,16 @@ export const TrustlyModal: React.FC<Props> = ({
...
-            )(contentWindow)
+
+            try {
+              await actualHandleIframeLoad(
+                setIsOpen,
+                setIsSuccess,
+                generateTrustlyUrl,
+              )(contentWindow)
+            } catch (e) {
+              console.error(e.message)
```
palmenhq, line 105: i guess we could a) use just `console.error(e)` right?, and b) won't sentry capture this anyways if we write it on `console.error`? maybe thats ok and we can ignore it in sentry instead

palmenhq (commented):
nice one!

src/client/pages/ConnectPayment/components/TrustlyModal.tsx
virrios: That's what I wanted to test Sentry locally for ;) . I did not receive and issue when catching the error. (I spammed that connect button like a pro). 
 But if I removed the try/catch i got an issue in Sentry instantly.

We can use console.error(e to  get the whole stacktrace as well but I though the idea was just using this as a reminder and if you just use console.error(e) the output is a bit messy.

palmenhq (approved):
Awesome!

PR #438, opened 2020-11-30
Fix / make Sign button visible in all browsers

styrelius:
## What?

Make the `Sign` component visible in _all_ browsers by moving it (back) into the `<ScrollWrapper>` in `/Checkout/index.tsx`. This might not look _perfect_  in Safari on iOS if you scroll very enthusiastically but it's still the best solution if we're not going to completely remake the way we disable scrolling of the background when the checkout sidebar is open.

## Why?

Apparently the `Sign` component wasn't positioned correctly in Firefox after the very last changes made to these components last week, which resulted in the CTA button actually being completely hidden. Very sad. 🤕 

### Firefox
2020-11-30-sign-firefox

### Chrome
2020-11-30-sign-chrome

### Safari
2020-11-30-sign-safari-mac

### Safari iOS
2020-11-30-sign-safari-ios

### DIFF
--- a/src/client/pages/OfferNew/Checkout/Sign.tsx
+++ b/src/client/pages/OfferNew/Checkout/Sign.tsx
@@ -20,7 +20,7 @@ const Wrapper = styled('div')<WrapperProps>`
   box-shadow: 0 -1px 4px ${colorsV3.gray500};
   width: 100%;
   padding: 1rem;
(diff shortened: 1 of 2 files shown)


gustaveen (approved):
🙌

virrios:
Nice! Really like the pictures 👌.

PR #440, opened 2020-12-01
Chore / Update packages

styrelius:
## What?

1) Upgrade a bunch of packages with `yarn upgrade`.

2) Install the latest version of `@sentry/node` to fix some vulnerability of high severity connected to older versions of that package.

3) Remove the `sentryEnvironment` fallback `'dev'` in `sentry-server.ts` in order for my local `SENTRY_ENVIRONMENT` env variable to be picked up. Otherwise the environment will be "dev" in Sentry when you test the error logging locally.

## Why?

🧠 

### DIFF
--- a/src/client/utils/sentry-server.ts
+++ b/src/client/utils/sentry-server.ts
@@ -1,5 +1,5 @@
 const sentryDsn = process.env.SENTRY_DSN
-const sentryEnvironment = process.env.SENTRY_ENVIRONMENT || 'dev'
+const sentryEnvironment = process.env.SENTRY_ENVIRONMENT
 
 export const sentryConfig = () => ({
   dsn: sentryDsn,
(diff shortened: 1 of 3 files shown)


src/client/utils/sentry-server.ts
```diff
@@ -1,5 +1,5 @@
 const sentryDsn = process.env.SENTRY_DSN
-const sentryEnvironment = process.env.SENTRY_ENVIRONMENT || 'dev'
+const sentryEnvironment = process.env.SENTRY_ENVIRONMENT
```
gustaveen, line 2: Should we add these to the `.env.example` ?

gustaveen:
My OCD likes an overview of updated packages, so here it is 😄
Screenshot 2020-12-02 at 11 33 26

gustaveen (approved):
🙌

src/client/utils/sentry-server.ts
styrelius: That's a good idea but I'm not going to do that now since that's not really in the scope of the PR I would argue. But yes, I totally agree!

PR #441, opened 2020-12-02
Update ssn input type and ad maxlength

gustaveen:
## What?

Change input type to `type=text` and set maxlength on input based on ssn length for the specific market

## Why?
Input `type=number` is not optimal to use for ssn since it will increment on scroll or arrow clicks and can't have a max length. So in order to only input numbers we filter out all non-digit characters

### DIFF
--- a/src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
+++ b/src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
@@ -4,7 +4,7 @@ import * as yup from 'yup'
 import { RawInputField } from 'components/inputs'
 import { Market, useMarket } from 'components/utils/CurrentLocale'
 import { WithEmailForm, WithSsnForm } from 'pages/OfferNew/types'
-import { createSsnValidator } from 'pages/OfferNew/utils'
+import { createSsnValidator, ssnLengthByMarket } from 'pages/OfferNew/utils'
 import { useTextKeys } from 'utils/textKeys'
 
 const BottomSpacedRawInputField = styled(RawInputField)`
@@ -41,6 +41,7 @@ export const UserDetailsForm: React.FC<Props> = ({
 
   // FIXME should we pick which ssn validation to use in a different way? or not really?
   const market = useMarket()
+  const ssnMaxLength = ssnLengthByMarket[market]
   const isValidSsn = createSsnValidator(market!)
 
   const handleSsnChange = (event: React.FormEvent<HTMLInputElement>) => {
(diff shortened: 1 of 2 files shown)


src/client/pages/OfferNew/utils.ts
```diff
@@ -234,6 +234,12 @@ export const maskAndFormatRawSsn = (ssn: string) => {
   return ssn.substr(CENTURY_LENGTH, DATE_LENGTH) + '-****'
 }
 
+export const ssnLengthByMarket: Record<Market, number> = {
+  SE: 12,
+  NO: 11,
+  DK: 10,
+}
```
vonElfvin, line 241: Maybe we could consider handling both 12 and 14 characters at some point. The backend can handle both and if the user wants to input it with 19 at the start, why not let them 🤷

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
```diff
@@ -41,9 +41,14 @@ export const UserDetailsForm: React.FC<Props> = ({
...
   // FIXME should we pick which ssn validation to use in a different way? or not really?
   const market = useMarket()
+  const ssnMaxLength = ssnLengthByMarket[market]
   const isValidSsn = createSsnValidator(market!)
 
-  const setSsn = (newSsn: string) => {
+  const handleSsnChange = (event: React.FormEvent<HTMLInputElement>) => {
+    const { value } = event.currentTarget
+    const nonNumbersRegex = /\D/
+    const newSsn = value.replace(nonNumbersRegex, '')
```
styrelius, line 50: This is a clean solution and I'm obviously approving it ✨
I would just like to raise a question here: wouldn't we rather check for what we _do_ want in the input rather than for what we _don't_ want in a case like this? To me that would probably be a tiny bit more clear 🙃  You know, something like this:
```suggestion
    const numbersRegex = /^[0-9]*$/
    const newSsn = numbersRegex.test(value) ? value : ''
```

styrelius (approved):
Very neat! 👌 👌

src/client/pages/OfferNew/utils.ts
styrelius: You mean both 10 and 12 characters? 🙊

gustaveen: Yup, that's allowed now @vonElfvin as you pointed out @styrelius  :)

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
gustaveen: I see what you mean, I had a solution like that at first :) but I actually prefer not having the ternary operator. Maybe @palmenhq has influenced me more than I know 😄

PR #442, opened 2020-12-02
Fix Norway Update translations to quick fix bug co-insured.

virrios:
## What?
We currently have a bug in Norway details view.
We show the value for _co-insured_ as the _total number of insured people_.
This pr fixes this by updating the textKeys for that input field for textkeys in Norwegian and Norwegian English.

Updated 
* src/translations/nb_NO.json
* src/translations/en_NO.json

##### OBS this is a quick fix for the problem in production. We are going to move away from using co-insured. So this pr is only changing Norway and not Denmark. Since we are going to use Number insured people for every market. When we launch Danmark this should be fixed. Thats why I omit changing Denmark.

The reason for the quickfix is that this is a pressing issue and The big solution to deprecate coInsured involves changes in many places/repos and involving different teams.

##### in English

image

##### in Norwegian
image


styrelius:
> src/translations/da_DK.json
> src/translations/en_DK.json

So you mean

- src/translations/nb_NO.json 
- src/translations/en_NO.json 

? 🙃

styrelius (approved):
👍

virrios:
Y

> > src/translations/da_DK.json
> > src/translations/en_DK.json
> 
> So you mean
> 
> * src/translations/nb_NO.json
> * src/translations/en_NO.json
> 
> ? 🙃

Yes! 😵

PR #444, opened 2020-12-07
Feat/add ssn error message

gustaveen:
# Note! ⚠️

This PR **doesn't** fix the following existing behaviour and will be addressed in separate PRs
- Display an error message if you don't input enough numbers
- We don't disable submit if the user removes a valid ssn and inputs a new, invalid ssn (edge case behaviour IMO)
- Fix that error push content downwards (I think it should)

## Updated screen recording

mobile

## What?

Add error message for ssn validation. The validation is handled on the backend, meaning we send a `editQuote` mutation when the user inputs a valid ssn in therms of length and a regex pattern. The possible error codes we can get is: 
```
INVALID_SSN
INVALID_SSN_LENGTH (we won't get this since we only do the mutation on correct ssn length)
SSN_DOES_NOT_MATCH_BIRTH_DATE
```

Also some ui updates like colors, error message and focus state.

## Why?

Why currently doesn't show any user feedback if the ssn is invalid. It would be good to add a frontend error as well if the format is invalid and  since we only try to update the quote if matches the regex pattern. But let's start with this since it's way better then no errors 😄


src/client/pages/OfferNew/Checkout/hooks.ts
```diff
@@ -45,3 +46,22 @@ export const useScrollLock = (
...
+export const useSsnError = (editQuoteResult: EditQuoteMutationResult) => {
+  const [ssnBackendError, setSsnBackendError] = useState('')
+  const editQuote = editQuoteResult.data?.editQuote
+
+  useEffect(() => {
+    if (
+      editQuote?.__typename === 'UnderwritingLimitsHit' &&
+      editQuote?.limits.length > 0
+    ) {
+      const { code } = editQuote.limits[editQuote.limits.length - 1]
```
gustaveen, line 59: Since we get an array of limits for certain `UnderwritingLimitHits`, for reference a non matching ssn and birthdate returns this. If you have a better idea, shoot 😊  
Screenshot 2020-12-07 at 15 20 17

styrelius: Hm, I guess if you would use this hook somewhere else, say in the modal for editing your details, you could potentially get more limits back and then we can't know that it's the last one that we want to set as  `ssnBackendError`. So that might be a concern - that this hook only works if you're not trying to edit more quote properties than the ssn. But on the other hand this file is in the `/Checkout` folder so maybe it's clear enough that the hooks here are not intended to be used anywhere else. 🤔

styrelius:
This will be such an improvement! ✨ 👌 🌻 
Question though: should we drop the ssn validation that we're doing with the `createSsnValidator` from `.../pages/OfferNew/utils`? Or, alternatively, add an error message for that as well? 
I mean otherwise we will still have this issue with nothing happening and the button being disabled if the ssn doesn't pass the frontend validation (e.g. the third character is _not_ 0 or 1 in the Norwegian checkout).

src/client/pages/OfferNew/Checkout/hooks.ts
gustaveen: I think it's fine as long as it's a local hook, in this specific folder. And since we can only get three different limits, or two really I think it covers the specific cases.

gustaveen:
> This will be such an improvement! ✨ 👌 🌻
> Question though: should we drop the ssn validation that we're doing with the `createSsnValidator` from `.../pages/OfferNew/utils`? Or, alternatively, add an error message for that as well?
> I mean otherwise we will still have this issue with nothing happening and the button being disabled if the ssn doesn't pass the frontend validation (e.g. the third character is _not_ 0 or 1 in the Norwegian checkout).

Yes, we should add validation for that for sure. 🙌 I just wanted to make this in batches 😊

styrelius:
> Yes, we should add validation for that for sure. 🙌 I just wanted to make this in batches 😊

@gustaveen  Nice! So you don't think we should remove it (meaning the validation we do on the frontend with the `createSsnValidator`) and just let the backend do that kind of validation?

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
```diff
@@ -114,15 +124,9 @@ export const UserDetailsForm: React.FC<Props> = ({
           pattern="[0-9]*"
           maxLength={ssnMaxLength}
           value={ssn}
-          // errors={ssnError ? textKeys.SIGN_SSN_CHECK() : undefined} TODO error handling?
```
palmenhq, line 117: lol, apparently _someone_ thought of this at some point in history 😬

src/client/pages/OfferNew/Checkout/hooks.ts
```diff
@@ -45,3 +46,22 @@ export const useScrollLock = (
       window.removeEventListener('touchmove', listener)
     }
   }, [outerWrapperRef, visibilityState])
+
+export const useSsnError = (editQuoteResult: EditQuoteMutationResult) => {
+  const [ssnBackendError, setSsnBackendError] = useState('')
```
palmenhq, line 51: what about using `null` for the no-error state?

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
```diff
@@ -114,15 +124,9 @@ export const UserDetailsForm: React.FC<Props> = ({
...
+          errors={ssnBackendError ? textKeys[ssnBackendError]() : undefined}
```
palmenhq, line 127: i'd like to challenge whether we should use the error code from backend and translate that into a text key.. Could we maybe have some kind of mapping, and maybe map unknown errors to some generic text key (for example, if backend starts giving an error we don't have a text key for yet)

palmenhq (commented):
Looks great! But with some thoughts.

src/client/pages/OfferNew/Checkout/hooks.ts
gustaveen: True! I fix! 🙌

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
gustaveen: It's so few in this case but as you say it can be hard to keep up if this changes, so good point! I'll have a look!

gustaveen:
> > Yes, we should add validation for that for sure. 🙌 I just wanted to make this in batches 😊
> 
> @gustaveen Nice! So you don't think we should remove it (meaning the validation we do on the frontend with the `createSsnValidator`) and just let the backend do that kind of validation?

I'm not sure what I think 😄 Frontend validation is way faster so from a user perspective it's better. And I guess it's somewhat good to avoid mutations we know will fail. But it's a bit of a hustle to handle both frontend and backend error messages 😅 So in that sense it's easier to handle all in the backend

src/client/pages/OfferNew/Checkout/hooks.ts
virrios: > > > Yes, we should add validation for that for sure. 🙌 I just wanted to make this in batches 😊
> > 
> > 
> > @gustaveen Nice! So you don't think we should remove it (meaning the validation we do on the frontend with the `createSsnValidator`) and just let the backend do that kind of validation?
> 
> I'm not sure what I think 😄 Frontend validation is way faster so from a user perspective it's better. And I guess it's somewhat good to avoid mutations we know will fail. But it's a bit of a hustle to handle both frontend and backend error messages simultaneously 😅 So in that sense it's easier to handle all in the backend

Side note thought
If we are going to reuse some/all of this in the details modal, which currently only fires the mutations when clicking update, a diskussion regarding reuse of input fields and validation would be great and how we want to go about it.

gustaveen:
I'm looking at adding frontend validation now, when the user have written too few characters for example. IMO we can merge this as a first step and I'll continue in a new PR. WDYT?

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
```diff
@@ -64,6 +63,14 @@ export const UserDetailsForm: React.FC<Props> = ({
     )
   }
 
+  const handleSsnBlur = () => {
```
styrelius, line 66: Is this `onBlur` handler really necessary? I mean the `onSsnChange` will run when `isValidSsn` is true anyway, thanks to `handleSsnChange`?

styrelius (commented):
Nice! 🌻 
One thing in addition to my comment below - should we do something about this thing that all content below the ssn input field will jump when the error message occurs and then jump again when it disappears? Like having a fixed height for the error message container or something? 🤔 

> I'm looking at adding frontend validation now, when the user have written too few characters for example. IMO we can merge this as a first step and I'll continue in a new PR. WDYT?

Also, I agree that we can start with this and add error messages from frontend validation as a next step.

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
styrelius: Or am I missing something?

gustaveen: Dunno really, just wanted remove it as an inline function to keep the component cleaner :) But I will use it when adding FE validation 🙌

gustaveen:
Some style updates like larger start date "input" and added label, decreased font size of price and such

virrios:
Looks good, awesome work! 🙌 Tested it on my samsung as well 📱 

Found a minor strange behaviour. If I type in a valid ssn-> it fires an edit mutation successfully-> if I remove or type in an invalid ssn after that-> the mutation fails and does not update the valid quote. I can then still  press the sign button and move forward in the flow, since the quote is valid. 
Maybe we should disable the sign button if we have ssn errors. 
somewhatStrange

gustaveen:
Yup, that is a weird behaviour we have right now 🤪 @virrios Ellen have added a ticket for that previously https://hedvig.atlassian.net/browse/HVG-604

gustaveen:
I can continue to work on that if we think we should fix that as well now, but we will probably not merge this until next week then. 🙂 I'm open for either options. 🙌

styrelius:
> I can continue to work on that if we think we should fix that as well now, but we will probably not merge this until next week then. 🙂 I'm open for either options. 🙌

I think we can do that in a separate branch! In my view it's not really in the scope of "add ssn error message" 😇

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
styrelius: Yes, I'm all for cleanliness! 👌 Just realised that it probably doesn't do anything right now. But I guess we can leave it then if you'll use it later on, although I think as a principle it's better to remove stuff that's not used in order to make the PR easier to grasp. And then you can add it again when you need it 😊

src/client/pages/OfferNew/Checkout/CheckoutContent.tsx
```diff
@@ -38,19 +39,25 @@ const StartDateWrapper = styled.div`
   margin-bottom: 2rem;
 `
 
+const StartDateLabel = styled.p`
```
styrelius, line 42: I also like this syntax better than the `styled('p')` (which I guess is older, or?) but maybe we should at least be consistent within files and use the same syntax everywhere, and in this file all other `styled` components are declared with the `('p')`.

styrelius (commented):
Awesome! 🌈  Would like to see how it looks now after the styling changes but since where having some BE issues that's not really possible right now. But I think I can trust that it looks good, I mean it usually does 😊

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
styrelius: Yes, I'm all for cleanliness! 👌 Just realised that it probably doesn't do anything right now. But I guess we can leave it then if you'll use it later on, although I think as a principle it's better to remove stuff that's not used in order to make the PR easier to grasp. And then you can add it again when you need it 😊

gustaveen:
Added the latest screenrecording and clarified some issues @styrelius @virrios :)

src/client/pages/OfferNew/Checkout/CheckoutContent.tsx
gustaveen: updated

gustaveen:
> Awesome! 🌈 Would like to see how it looks now after the styling changes but since where having some BE issues that's not really possible right now. But I think I can trust that it looks good, I mean it usually does 😊

Thank you! 🙌  Hopefully BE will be up tomorrow so we can test it before we deploy to prod

styrelius (dismissed):
🥳

styrelius (approved):
🥳  again!

PR #445, opened 2020-12-07
Feature/hvg 582 debugger accident travel denmark

virrios:
[HVG-582] Create accident & accident + travel via debugger

## What?
* Added the the capability to create bundle quotes for denmark, Home + Accident, Home + Accident + Travel,
as a selectable input.
* Updated from using quoteId to quoteIds.

## Why?
We need to be able to create quote bundles in Denmark

_Referenced ticket [https://hedvig.atlassian.net/browse/HVG-582]()_
#### Home 

Home

#### Home + Accident
Home + Accident

#### Home + Accident + Travel
Accident +Travel

[HVG-582]: https://hedvig.atlassian.net/browse/HVG-582


src/client/utils/tracking/tracking.ts
```diff
@@ -69,7 +69,19 @@ export const { TrackAction, IdentifyAction } = setupTrackers<
...
+  TypeOfContract.DkAccidentStudent,
+  TypeOfContract.DkTravel,
+  TypeOfContract.DkTravelStudent,
+] as const
+type NotValidContracts = typeof NOT_VALID_CONTRACTS[number]
+
+type TypeOfContractExcludedUnused = Exclude<TypeOfContract, NotValidContracts>
+
+const adtractionProductMap: {
+  [type in TypeOfContractExcludedUnused]: number
```
gustaveen, line 83: I guess we should remove line 96 -> 99 after this change right :)

src/client/utils/tracking/tracking.ts
```diff
@@ -69,7 +69,19 @@ export const { TrackAction, IdentifyAction } = setupTrackers<
   { debug: process.env.NODE_ENV === 'development' },
 )
 
-const adtractionProductMap: { [type in TypeOfContract]: number } = {
+const NOT_VALID_CONTRACTS = [
+  TypeOfContract.DkAccident,
+  TypeOfContract.DkAccidentStudent,
+  TypeOfContract.DkTravel,
+  TypeOfContract.DkTravelStudent,
+] as const
```
gustaveen, line 77: Oh actually we should add all DK contract types
```suggestion
const NOT_VALID_CONTRACTS = [
  TypeOfContract.DkHomeContentOwn,
  TypeOfContract.DkHomeContentRent,
  TypeOfContract.DkHomeContentStudentOwn,
  TypeOfContract.DkHomeContentStudentRent,
  TypeOfContract.DkAccident,
  TypeOfContract.DkAccidentStudent,
  TypeOfContract.DkTravel,
  TypeOfContract.DkTravelStudent,
] as const
```

src/client/utils/tracking/tracking.ts
```diff
@@ -69,7 +69,19 @@ export const { TrackAction, IdentifyAction } = setupTrackers<
...
+type NotValidContracts = typeof NOT_VALID_CONTRACTS[number]
+
+type TypeOfContractExcludedUnused = Exclude<TypeOfContract, NotValidContracts>
+
+const adtractionProductMap: {
+  [type in TypeOfContractExcludedUnused]: number
+} = {
```
gustaveen, line 84: Minor thing and this is optional, but I think it's easier to understand a Record :) 
```suggestion
const adtractionProductMap: Record<TypeOfContractExcludedUnused, number> = {
```

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -2,7 +2,13 @@ import React, { useEffect, useState } from 'react'
 import { Form, Formik, FormikProps } from 'formik'
 import { v4 as uuid } from 'uuid'
 import { colorsV3 } from '@hedviginsurance/brand'
-import { CreateQuoteInput, useQuoteLazyQuery } from 'data/graphql'
+import { CreateQuoteVariables } from '@hedviginsurance/embark'
+import {
+  CreateQuoteInput,
+  useCreateDanishHomeAccidentQuoteMutation,
+  useCreateDanishHomeAccidentTravelQuoteMutation,
```
styrelius, line 9: Actually, now when I think about it I'm wondering if we rally should create these mutations that will only be used in the debugger? The way I think it works with bundles for Norway in the real onboarding Embark flow is that the `createQuote` mutation is just used twice - once for each contract. And if that's the way we'll do it in Denmark (which I guess it is but please correct me if I misunderstood) then maybe the debugger should mirror that? In the way that we should use the same mutation that will be used in the real flow, I mean. That way we can also be more confident that it will work with Embark as well, or? What are your thoughts on that everyone?

src/client/utils/tracking/tracking.ts
```diff
@@ -110,9 +122,20 @@ export const adtraction = (
...
+      if (isBundle(offerData))
+        return getComboAdractionProductValue(isYouth(offerData))
+      const contractType = offerData.quotes[0].contractType
+      if (!(contractType in NOT_VALID_CONTRACTS)) {
+        return adtractionProductMap[
+          contractType as TypeOfContractExcludedUnused
+        ]
+      }
+      return null
+    }
```
gustaveen, line 135: Maybe we could do an early return here :) 
```suggestion
    const getTpTag = () => {
      const contractType = offerData.quotes[0].contractType
      if (contractType in NOT_VALID_CONTRACTS) {
        return null
      }

      if (isBundle(offerData)) {
        return getComboAdractionProductValue(isYouth(offerData))
      }

      return adtractionProductMap[contractType as TypeOfContractExcludedUnused]
    }
```

src/client/utils/tracking/tracking.ts
```diff
@@ -110,9 +122,20 @@ export const adtraction = (
...
+
+    const tpTag = getTpTag()
+    if (tpTag) adt.Tag.tp = tpTag
```
gustaveen, line 138: Do we want to trigger the event if it's a dk quote?

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -90,9 +108,26 @@ const getCurrentAvailableQuoteData = (
   return currentQuoteTypeData
 }
 
+const getDanishQuoteValues = (
+  values: Values,
+  quoteType: 'danishTravel' | 'danishAccident',
+) => {
+  const { danishHomeContents, ...filteredValues } = values!
+  const { type, livingSpace, ...QuoteTypeValues } = danishHomeContents!
```
styrelius, line 116: OK this might be a bigger discussion on how we use TypeScript but to me it feels like something's actually off with the typing, which should probably be properly adressed by creating new types or whatever, if one needs to use non-null assertion and type assertion as much as we do in this file now 🙈 But I can of course be wrong and this is the only way to do it.

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -109,94 +144,140 @@ export const Offer: React.FC<OfferProps> = ({ sessionToken }) => {
...
               <InputField
-                label="Quote id"
-                name="Quote id"
-                placeholder="d6c60432-dc7b-4405-840e-b4fd8164e310"
-                onChange={(e: React.ChangeEvent<HTMLInputElement>) => {
-                  setQuoteId(e.target.value)
-                }}
-                value={quoteId}
+                label="Type"
+                placeholder=""
```
styrelius, line 243: I think the `placeholder` prop can be removed 🧹

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -109,94 +144,140 @@ export const Offer: React.FC<OfferProps> = ({ sessionToken }) => {
...
+          <Formik
+            initialValues={{ quoteIds }}
+            onSubmit={() => {
+              /* noop */
+            }}
+          >
+            {() => {
+              return quoteIds.map((quoteId, index) => (
+                <InputField
+                  key={`quote_ids_${index}`}
```
styrelius, line 261: Neat way to name the `key` I think! 😀

styrelius (commented):
Great job! 👍  Some things though:

When I change type in the debugger form I get this GraphQL error:
image
Is it something you recognise? Maybe it's something changed on the backend causing it?

src/client/utils/tracking/tracking.ts
styrelius: I agree! ☝️

src/client/pages/Debugger/components/Offer.tsx
```diff
@@ -109,94 +144,140 @@ export const Offer: React.FC<OfferProps> = ({ sessionToken }) => {
     values: Values,
     storage: Record<string, unknown>,
   ) => {
-    if (sessionToken) {
+    const input = {
+      ...values,
+      id: quoteIds[0],
+      currentInsurer: values.currentInsurer || undefined,
+      // @ts-ignore
```
gustaveen, line 151: Is this ignore still needed, passed typecheck without it for me :)

src/client/pages/Debugger/components/Offer.tsx
gustaveen: I have the least insight I think on this so I might be completely off 😄, but didn't we say that DK should have separate mutations because the way we did bundles in NO was not the path we wanted to continue on?

gustaveen:
Good job with this! 🙌  It works nicely to create a bundle now and I think we can continue to enhance the debugger if we want when we pick up DK again. Offer page needs some fix as you highlighted @styrelius and tickets has been made for that 👌 So let's resolve the discussions and be pragmatic on what we fix now and what we can do later 🎉

src/client/utils/tracking/tracking.ts
virrios: okej we should remove the existing danish types as well?

src/client/utils/tracking/tracking.ts
virrios: If we can remove every danish quote then we can skip to trigger it. Did not realise that the already existing danish types should also be put in the invalid type of contracts.
Then if that is the case we can just check if it's  danish quote and we can remove the list that I compare to. 
I thought that we wanted to keep the contracts that we had in the adtractionProductMap.
That makes it easier

src/client/pages/Debugger/components/Offer.tsx
styrelius: Oh, that might be completely true, I just missed slash forgot that then, sorry about that! So does that mean that these mutations will be used in Embark as well?

virrios:
> When I change type in the debugger form I get this GraphQL error:
We get this error in dev as well, however the error message is more descriptive. Its because we have this useEffect(in master we use getQuote) which queries the quote(in master)/quotes(in this branch) that does not exist yet.

  useEffect(() => {
    if (quoteIds.every((quote) => quote?.length === 36) && sessionToken) {
      getQuotes({
        variables: { input: { ids: quoteIds }, locale: localeIsoCode },
      })
    }
  }, [quoteIds, sessionToken, localeIsoCode, getQuotes])

Not sure of the purpose of the useEffect.

src/client/pages/Debugger/components/Offer.tsx
virrios: I think it's more readable to to one mutation instead of multiple and more easily managed. But I agree 100% with your point of being consistent with how we handle it in Embark/Onboarding.

styrelius:
@virrios I think you might be mixing the errors up - one error is the one we get when we try to fetch a quote that might exist or not, and it doesn't exist. And yes that is what happens in that `useEffect`. The purpose of that one is that if you want to look at a specific quote for some reason that quote will be fetched if you just add the ID to the input. Also this means that would the generated quote ID by insane chance be the same as an existing one you would just get that data instead of failing to create a new quote. I don't know if it's _that_ useful but anyway, that's how we do it now. 🤷‍♀️ 

But the error I submitted a screenshot from is _another_ error that I've never seen before. And we don't have that in dev. It occurs specifically when you change type of contract.

virrios:
@styrelius  
Sry maybe I was a bit short in my explanation 😵. What I meant was the "corresponding" error for the new query, since we changed from query Quote to query QuoteBundle.  
I mean, the cause of the error is calling the query. 
When we update type now, we update the quoteIds causing the useEffect to fire and the in turn the query to fire.

_"The purpose of that one is that if you want to look at a specific quote for some reason that quote will be fetched if you just add the ID to the input._"
 -Okey i see ,  is this applicable to when we have bundles, maybe the useEffect should not fire then?

I think they update something because now I get another error Contract not created.

(**Edit**  now i get the same old error, are they updating the backend?)


> **update:**
> OK so it seems to be something strange going on in the backend because I get the same error when trying to create a Norwegian quote on another branch.

Is it editQuote or another query/mutation? 😮

styrelius:
@virrios I don't know, but I guess we can leave the `useEffect` as it is.

src/client/utils/tracking/tracking.ts
```diff
@@ -122,20 +118,13 @@ export const adtraction = (
...
-      if (isBundle(offerData))
-        return getComboAdractionProductValue(isYouth(offerData))
-      const contractType = offerData.quotes[0].contractType
-      if (!(contractType in NOT_VALID_CONTRACTS)) {
-        return adtractionProductMap[
-          contractType as TypeOfContractExcludedUnused
-        ]
-      }
-      return null
+    if (!isDanish(offerData)) {
```
gustaveen, line 121: Nice! 👌

gustaveen (approved):
🐛 🇩🇰  🥳

PR #447, opened 2020-12-15
[HVG-629] Add ability to edit the personnummer in SE checkout

palmenhq:
## What?

Add personnummer in SE checkout. Also separate the different personnummer formats in to different text keys.

## Why?

Re text key separation the text keys should be translations, not market-specific things. Hence i moved the different markets' SSN format into separate text keys. Also i figured it makes sense to have DK and NO separated even if they have the same format, just for consistency.

_Referenced ticket [here](https://hedvig.atlassian.net/browse/HVG-629)_

Screenshot 2020-12-15 at 10 49 21


gustaveen:
Awesome Palmen! 🥳 Just some thoughts. 

Good clarification between translations and market. 👌 But I see some downsides with this approach:

- We will end up with a lot of empty translations. It might be a bit more cumbersome to spot if there actually is missing translations in PRs for example (even though it should be quite clear with the market suffix)
- In Lokalise the stats will not show that the translations are completed. Maybe not a problem but if we would use translators it could be hard for them to look at a language to know what's left to translate (see screenshot). But this might just be as easy as marking them as verified, but I assume this will be forgotten most of the time.
- If we would like to use the sam workflow as the app team, where translations and keys are added in Figma and uploaded to Lokalise, it might mean some extra work. (I'm guessing here but worth checking with Zak)

So this is maybe a larger translation workflow discussion, just wanted to check that we are thoughtful of how we want to grow this :)

gustaveen:
Since we use the 10 digit SSN format for Sweden in all places in Embark, should we maybe use it in the checkout as well? But still allow 12 digits. (Looking at Klarnas login once again 😄 )

styrelius:
I agree with @gustaveen here, and I also feel that it can be tricky sometimes to define what a _translation_  actually is. So I think it would probably be a very good idea to have "this larger translation workflow discussion" in some other forum. I've also been working a lot with Lokalise these past days and have been a bit confused  about how I should think regarding this, despite the very clear message in Oscar's Lokalise guide that "the meaning of each of the translated variants should be the same - do not change the meaning based on language."

gustaveen:
This might be clearly for defined already how we want to work with translations and just me that don't get it :) . Feature wise my only question is what SSN format for Sweden we want to write in the placeholder and label 😃 We can always check with Carl and Mange

src/client/pages/OfferNew/utils.ts
```diff
@@ -262,13 +262,12 @@ export const ssnLengthByMarket: Record<Market, number> = {
...
+  switch (market) {
+    case Market.No:
+      return /^[0-9]{2}[0,1][0-9][0-9]{2}[ ]?[0-9]{5}$/.test(ssn)
+    case Market.Dk:
+      return /^[0-9]{2}[0,1][0-9][0-9]{2}[ ]?[0-9]{4}$/.test(ssn)
+    case Market.Se:
+      return /^([1-2][0-9])?[0-9]{2}[0-1][0-9][0-9]{2}[-+]?[0-9]{4}$/.test(ssn)
   }
-
-  throw new Error(`Expected market to be NO, DK or SE but was ${market}`)
```
virrios, line 273: You could throw the error in the default case but, maybe were not planning to add any new markets any time soon and update the enum.

virrios:
Looks super! Nice clarification from all of you regarding the translations. I did not know that there was a definition on how we work with this. Byt how do you define meaning. I think in this case I think It a bit of a gray zone. I think it's possible to  argue for both cases. 

**Downside**:
* More logic to handle different markets
* Empty Values in localize
* What is "meaning" ssn meaning can it be YYYYDDMMXXXX/ DDMMYYYYXXXXX? (the underlying meaning is the same i.e  identification?) #semantics? 

**Upside**
* Make clear distinction that it works in different ways in different markets. 
* The correct whay to interpret "meaning"?
* As I understood it, the App team does it this way? #consistency

src/client/pages/OfferNew/utils.ts
palmenhq: The beauty of typescript is that `switch`es become exhausitive, meaning that whenever we add a market to the enum this'll give a compile time type error which imo is a much better way of detecting errors. So i _think_ this should never happen

palmenhq:
Thanks all, good points. I think we need a discussion around translations if we don't like the current structure. It's already quite wide spread if we are afraid of empty translations, and imo now is not a good timing to redo how we treat case-specific translations anyways :) to me, this approach makes sense for now and the structure is already there with market-specific english translations and stuff. I'll look in to how we could mask the ssn.

gustaveen:
> Thanks all, good points. I think we need a discussion around translations if we don't like the current structure. It's already quite wide spread if we are afraid of empty translations, and imo now is not a good timing to redo how we treat case-specific translations anyways :) to me, this approach makes sense for now and the structure is already there with market-specific english translations and stuff. I'll look in to how we could mask the ssn.

For sure, let's not make a big switch right now! 🙌 Awesome!

styrelius:
I just want to add that I really agree with one of Victor's points here, the one about "meaning". To me the meaning of the SSN format is the same even though the formats differ. Saying that different formats gives different meanings is a _little_ like saying different spellings equals different meanings, i.e. the entire concept of translation is lost 🙃 

So to _me_ it makes more sense to use only one text key for the ssn placeholder and one for the label (or whatever the exact structure was before these changes), and don't have all this logic to detect market and then pick the right text key.

palmenhq:
@styrelius @virrios 
> [...] **To me the meaning of the SSN format is the same even though the formats differ.** Saying that different formats gives different meanings is a _little_ like saying different spellings equals different meanings, i.e. the entire concept of translation is lost 🙃
> 
> So to _me_ it makes more sense to use only one text key for the ssn placeholder and one for the label (or whatever the exact structure was before these changes), and don't have all this logic to detect market and then pick the right text key.

**tldr; I really don't think we should market-separate in the same text key**

Coming back to this after having slept on it I have formed a stronger opinion and i deffo do not agree on this reasoning on the semantics. The translators should work with the translations, not with market separation. It works for us (developers) to change SSN formats across markets now because we understand scandinavian languages -  but:

a) it is _not_ up to the translators to decide on which input format their particular language should have, so somebody else needs to care about the implicit market separation to adjust the input formats. Again this works for scandinavian languages, maybe also for other western languages, but imagine going through that process in greek or with the cyrillic alphabet. Very error prone.

b) There's a high risk translators will not translate input formats correctly if it is not very clear that the source language has a different input format than what we expect from the translation.

c) If we share languages across markets (ie we go to france and belgium, or finland and have swedish there) it is very easy to miss a couple of text keys with market-specific input formats, as we have no clear way of differentiating them from market-agnostic translations at the moment.

imo these are very clear and solid arguments but i'd be happy if you could shoot them down 😄

gustaveen:
Valid points and good clarification @palmenhq In the bigger picture I agree that translations should be just that, purely translations. 😀

styrelius:
@palmenhq 
> a) it is not up to the translators to decide on which input format their particular language should have

This is a valid point for sure - but on the other hand translators probably know which formats are the most common for that country so maybe we actually _should_ let them decide. Like how did we decide on SSN format for Denmark now? If it's just that we googled it then maybe it's better that someone who knows the language decides?

(Also the way we do it now I don't think it's at all clear which keys should be translated since we have inputs for all languages in every key (hence we get the empty translations). Especially when you get the English ones automatically it can look as if some keys should be translated although they shouldn't. Is it possible to remove that automation? And maybe pick which translations you want for a text key?)

But hey, this is not a hill I'm willing to die on so let's do it this way.

styrelius (approved):
🤝

PR #449, opened 2020-12-15
Chore/add story for sidebar

gustaveen:
## What?

Added story for Sidebar in Storybook

## Why?

Since it's awesome to work in storybook 🥳

sidebar


styrelius (approved):
Nice! 👌 

Question: would it be possible to add the ability to switch between quotes in the Storybook view? Since we have some other mock quotes as well. Or is that too complicated to implement to be worth it? 

Also, the name `Sidebar` is pretty confusing to me but I of course know it was already called that. Just wondering if we should maybe rename it to whatever we call it now when we talk about it (is it "CTA box"?)?

gustaveen:
@styrelius for sure, I was thinking of adding dk and no like we have on the story for Landing now when I start to updating the design 👍 Yeah Sidebar is a bit confusing haha

PR #451, opened 2020-12-18
[HVG-676] Feature/sidebar update

gustaveen:
## What?
Update design and layout in price overview on Offer page, see Figma file for reference: https://www.figma.com/file/eRR5ACHtbsPWDd4eUauWv1/Onboarding-Offer-page?node-id=1%3A648

Have done a lot of testing to make sure we can fit all titles, discounts and prices into this little box :) We will adjust some of the discount texts that are a bit too long.

I also altered the conditional for when we show the original price, we didn't do that for bundle discount before. Danish prices are not taken into consideration and is out of scope for this PR.

**Review app:** https://web-onboardi-feature-si-hxxnsk.herokuapp.com/se/new-member

[HVG-676]

### Desktop NO Combo - edge case double discounts
Screenshot 2020-12-21 at 15 48 09

### Desktop SE Rent - no discount
Screenshot 2020-12-21 at 16 02 24

### Mobile NO - EN Combo - edge case double discounts with long discount copy
If a discount is very long and breaks into a new line, make sure it looks ok

[HVG-676]: https://hedvig.atlassian.net/browse/HVG-676


src/client/pages/OfferNew/Checkout/CheckoutContent.tsx
```diff
@@ -98,27 +90,23 @@ export const CheckoutContent: React.FC<Props> = ({
...
-          </div>
+            </span>
+          </InsuranceHeading>
           <div>
             <Price
               loading={fakeLoading || reallyLoading}
               monthlyGross={offerData.cost.monthlyGross}
               monthlyNet={offerData.cost.monthlyNet}
-              monthlyCostDeduction={monthlyCostDeduction}
+              isDiscountPrice={monthlyCostDeduction}
```
virrios, line 109: nitpick, but maybe we could rename the monthlyCostDeduction as well to something with "is" as postfix

src/client/pages/OfferNew/Introduction/Sidebar/Sidebar.stories.tsx
```diff
@@ -35,3 +39,47 @@ export const Defualt = () => {
...
+    </MockedProvider>
+  )
+}
+
+export const DKHomeContent = () => {
+  const ref = useRef<HTMLDivElement>(null)
+  return (
+    <MockedProvider>
+      <MemoryRouter initialEntries={['/no/new-member/offer']}>
+        <TextKeyProvider locale="nb_NO">
```
virrios, line 70: Should be Dk? 🇩🇰  "nb_NO" and "no/new/member"

src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
```diff
@@ -30,28 +30,35 @@ import {
   isBundle,
 } from 'pages/OfferNew/utils'
 import { useTextKeys } from 'utils/textKeys'
+import { Size } from 'components/types'
 import { gqlDateFormat } from './utils'
 
 interface Props {
```
virrios, line 36: Nitpick but these props are use at line 355 
`export const StartDate: React.FC<Props> = ({
`
Maybe we can move it just above the Component definition, IMO it makes it easier to navigate the file.

virrios, line 36: Nitpick but these props are use at line 355
export const StartDate: React.FC<Props> = ({
Maybe we can move it just above the Component definition, IMO it makes it easier to navigate the file.

src/client/pages/OfferNew/Checkout/CheckoutContent.tsx
```diff
@@ -45,19 +44,12 @@ const StartDateLabel = styled.p`
...
+const InsuranceHeading = styled.h3`
   font-size: 1.5rem;
-  font-family: ${fonts.FAVORIT};
-  line-height: 1;
+  line-height: 2rem;
 
-  @media (min-width: 40rem) {
-    font-size: 2rem;
+  span {
+    display: block;
```
styrelius, line 52: Maybe just use a `div` since that has `display: block` as default?

src/client/pages/OfferNew/Introduction/Sidebar/Sidebar.stories.tsx
```diff
@@ -35,3 +39,47 @@ export const Defualt = () => {
     </MockedProvider>
   )
 }
+
```
styrelius, line 42: Awesome with the new mocked insurance types! 🙌 ✨

src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
```diff
@@ -185,7 +192,8 @@ const DateForm: React.FC<{
   setShowError: (showError: boolean) => void
   modal?: boolean
   refetch: () => Promise<void>
```
virrios, line 194: Nitpick, I think this is fine if one or two props, but when we have this many props I think it would be nice to have an Interface DateFormProps, what do you think?

src/client/pages/OfferNew/Introduction/Sidebar/index.tsx
```diff
@@ -18,13 +18,16 @@ import {
...
 import { useTextKeys } from 'utils/textKeys'
+import { Badge } from 'components/Badge/Badge'
 import { Price } from '../../components'
 import { insuranceTypeTextKeys, isBundle } from '../../utils'
 import { DetailsModal } from './DetailsModal'
 import { DiscountCodeModal } from './DiscountCodeModal'
 import { StartDate } from './StartDate'
 import { StickyBottomSidebar } from './StickyBottomSidebar'
 
+export const SIDEBAR_BP = '@media (min-width: 1020px)'
```
styrelius, line 29: To me `BP` indicates the breakpoint value - like just the number or the number + `px`. Maybe call it something like `SIDEBAR_DESKTOP_MEDIA_QUERY`? Then you also know that it's a media query for specifically desktop 😀

src/client/pages/OfferNew/Introduction/Sidebar/index.tsx
```diff
@@ -145,14 +129,8 @@ const FooterExtraActions = styled.div`
...
-    margin-right: 1.375rem;
-    :last-child {
-      margin-right: 0;
-    }
-  }
+  margin-top: 0.75rem;
+  margin-bottom: 0.25rem;
 `
 
 export const Sidebar = React.forwardRef<HTMLDivElement, Props>(
```
virrios, line 136: Forgot what we said if we should fix this things in components but, getting linting warnings on UseEffect,
That we should add to the dependency array.
It's a pretty easy fix.
two of them are functions and will probably not change, but
redeemedCampaigns, this one is data from a query so maybe this one should be in it?

styrelius (commented):
Fantastic! 🥳 Great work on this one! And it looks so good with the new design! A few small things though 👇

src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
styrelius: Or move the other component to its own file maybe? Or is that too much work now? 🤔

src/client/pages/OfferNew/Checkout/CheckoutContent.tsx
gustaveen: personally I think the name is fine, but is works as well. however the function to get that value on line 74 has that name already `const monthlyCostDeduction = isMonthlyCostDeduction(...`

src/client/pages/OfferNew/Introduction/Sidebar/Sidebar.stories.tsx
gustaveen: oops, good catch! 👌

src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
gustaveen: I totally agree, I'd rather have a prop type for that

src/client/pages/OfferNew/Checkout/CheckoutContent.tsx
virrios: yeah I saw that too! As I think you are implying I agree,  it would be too much to change the function name because that is a rabbit whole that every pr should fix all naming convetions.  I just thought for this component that the variable could be named isDiscountPrice.

virrios (commented):
Awesome work! 🔥  Just had some nitpicks and it was mostly about old code that was already there.

src/client/pages/OfferNew/Checkout/CheckoutContent.tsx
gustaveen: This is intentional, since a `div` is not allowed as a child of a heading element

src/client/pages/OfferNew/Checkout/CheckoutContent.tsx
```diff
@@ -98,27 +90,23 @@ export const CheckoutContent: React.FC<Props> = ({
...
         <Excerpt>
-          <div>
-            {market === Market.Se && (
-              <InsuranceTypeLabel>
-                {textKeys.SIDEBAR_LABEL()}
-              </InsuranceTypeLabel>
-            )}
-            <InsuranceType>
+          <InsuranceHeading>
+            {market === Market.Se && <span>{textKeys.SIDEBAR_LABEL()}</span>}
```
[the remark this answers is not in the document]
gustaveen: Good point! 👍

src/client/pages/OfferNew/Checkout/CheckoutContent.tsx
styrelius: Ah, I see! Didn't think about that. 🤕  And what about just wrapping each of the text element in an `<InsuranceHeading>` tag? Since they also have `display: block` as default?

src/client/pages/OfferNew/Introduction/Sidebar/index.tsx
gustaveen: I wanted to align this with the the naming in `market-web` where we have `const LAPTOP_BP_UP = '@media (min-width: 1024px)'` for example. I can change the name to desktop or tablet or something similar if we think it makes more sense 👍

src/client/components/Badge/Badge.tsx
```diff
@@ -0,0 +1,34 @@
...
+    size === 'lg' &&
+    `
+    @media (min-width: 1020px) {
+    font-size: 0.875rem;
+  }
+  `};
+`
+
+export const Badge: React.FC<BadgeProps> = ({ children, size = 'sm' }) => (
+  <BadgeContainer size={size}>{children}</BadgeContainer>
```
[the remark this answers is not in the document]
gustaveen: I think it's rather neat to just render children if we want to use the badge with icons and text for example :)

src/client/pages/OfferNew/Introduction/Sidebar/index.tsx
gustaveen: Added a new PR for that to isolate potential issues :) Still have to add all of them as dependencies in order to get rid of the lint warning https://github.com/HedvigInsurance/web-onboarding/pull/454

src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
gustaveen: I think it's a good idea to split it into two files 🙌 But it turned out it was a bit more than a minor fix so I'd rather do it outside this PR if it's ok with you :)

src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
gustaveen: Let's do this when splitting this into two files 👍

src/client/pages/OfferNew/Introduction/Sidebar/index.tsx
styrelius: Ah, OK, I see. Yes that would be amazing! 🙏

src/client/pages/OfferNew/Checkout/CheckoutContent.tsx
gustaveen: No worries! That would work for sure👌  But for the sake of semantics, it could be bad to split a title into multiple headings. In this particular page and with a h3 I assume it wouldn't matter much :D But let's say it was an h1, then it would be worse since you should only have on h1 per page

src/client/pages/OfferNew/Checkout/CheckoutContent.tsx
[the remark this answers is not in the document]
gustaveen: Updated now! 🙌

src/client/components/Badge/Badge.tsx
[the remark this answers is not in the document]
gustaveen: We can for sure discuss the `children` pattern and when to use it  👍

styrelius (approved):
🙌 💃

PR #452, opened 2020-12-21
Feature/hvg 637 details modal add all fields

virrios:
[HVG-637] Details > add all fields
## What?
Added common fields such as first name, last name, birthdate **(Birthdate not for Sweden)**.
Updated design:
* Added 3 more input fileds fields, stated above
* Removed Chat with us
* Not blocked by Hedvig header
*  Margin, padding update. 
* Had to change logic and update types in utils etc, to cope with the new fields

## Why?
* New design and we should be able to edit all quote fields.
* The user want to se the "update details button" 

_Referenced ticket [https://hedvig.atlassian.net/browse/HVG-637]()_

image
Detailssweden

**Mobile screen, background scroll is locked when modal is visible (Captured from chrome inspector and does not fully represent how it looks on the phone** 

MobielScreenScroll

[HVG-637]: https://hedvig.atlassian.net/browse/HVG-637

## Why?

_Referenced ticket [here]()_


src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/Details/components/DetailInput.tsx
```diff
@@ -9,6 +9,7 @@ interface DetailInputProps {
   field?: RegularFieldType
   formikProps: FormikProps<EditQuoteInput>
   nameRoot:
+    | ''
```
virrios, line 12: Went back and forth between having nameRoot optional, name optional but I thought this was the best solution since this is more in line with how its used. You should pass a root and you should pass a name. I think having this as optional makes it unclear how to use this component.

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/utils.ts
```diff
@@ -400,28 +428,32 @@ const getMarketFields = (fieldSchema: FieldSchema, offerQuote: OfferQuote) => {
...
       : (fieldSchema as SwedishHouseFieldSchema).swedishHouse,
     key: isSwedishAppartment ? 'swedishApartment' : 'swedishHouse',
   }
 }
 
+const validateFields = (fields: MarketFields | CommonFieldSchema) =>
+  Object.entries(fields).reduce(
+    (acc, t) => ({
+      ...acc,
+      ...getValidationSchemaHelper(t),
```
gustaveen, line 442: Not you that wrote this code, but what is `t` in this context 😄

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/utils.ts
```diff
@@ -600,7 +632,20 @@ const getInitialNorwegianTravelValues = (
   },
 })
 
-export const getInitialInputValues = (offerQuote: OfferQuote) =>
+export const getInitialInputValues = (
+  person: OfferPersonInfo,
+  offerQuote: OfferQuote,
+) => {
+  const { firstName, lastName, birthDate } = person
```
gustaveen, line 639: Should we maybe name this something like `personalInfo` or `memberDetails` ? Person sounds a bit weird to me :)

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/Details/components/DetailInput.tsx
gustaveen: I see your point :) But I think it's odd with passing an empty string as `nameRoot`. I think I would prefer having it as optional but not a big deal to me if we implement it like this

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/utils.ts
virrios: haha yes indeed what does t mean, good catch, I can  rename that one!

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/utils.ts
virrios: I agree! The reason i chose person was that the property of OfferData is named "person". That property is used in several files. Might be better to change that as well?

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/Details/components/DetailInput.tsx
virrios: I thought about again.  The point, that enforcing nameRoot helps in understanding the component, is maybe not that valid.... At least not worth it. Changing it to optional! 😆

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/utils.ts
gustaveen: aha! then let's not bother now 🙌

gustaveen:
Good job! 🎉 Works nicely for me on the iPhone 🙏

src/client/pages/OfferNew/utils.ts
```diff
@@ -34,6 +34,9 @@ export const getOfferData = (quoteBundle: QuoteBundle): OfferData => {
     },
     quotes: quoteBundle.quotes.map((bundleQuote) => {
       return {
+        firstName: bundleQuote.firstName,
+        lastName: bundleQuote.lastName,
+        birthDate: bundleQuote.birthDate,
```
styrelius, line 39: Do we need the `firstName`, `lastName` and the `birthDate` here when we have them on the `person` object? I mean they won't differ between the quotes? Or what did I miss? 😄

styrelius (commented):
Nice work with this! 👌  But you need to download translations again! 🚨 A lot of the translations that you have in this PR are *not* up to date and should _really_ not be merged!

Also two minor things 👇

src/client/pages/OfferNew/utils.ts
virrios: You are correct!

gustaveen (approved):
🎉

styrelius (approved):
🥳 💥

PR #455, opened 2020-12-23
[HVG-712] Fix / Offer page CTA button text

styrelius:
## What?

1) New copy for CTA button on /offer page.
2) Same `font-size` (`1rem`) for all `Button`'s, regardless of `size` prop, which means a larger font for `sm`/default `Button`'s and the same as before for `lg` `Button`'s.

## Why?

1) We've agreed that the current copy is a bit confusing - it can easily be interpreted as if you would sign by clicking that button (which you of course _don't_, so it's pretty misleading).
2) As for the `font-size` all new designs use `1rem` (16 px) font size for all buttons.

_Referenced ticket [here](https://hedvig.atlassian.net/browse/HVG-712)_

## Screenshots

**/se-en (with Hedvig Forever code activated though apparently it's called "Hedvig Zero" in English? 🤔)**

2020-12-23-offer-page-cta-se-en

**/se  (with Hedvig Forever code activated)**

2020-12-23-offer-page-cta-se

**/no**

2020-12-23-offer-page-cta-no


styrelius:
Everyone except me is on vacation but I did get the mandate to merge and deploy some changes on my own. So _hold on to your hat_, here we go! 😀

PR #460, opened 2021-01-11
[HVG-675] Feature / New Offer page hero section

styrelius:
## What?

The main change here is the re-designed hero section of `/offer` page:
1) Background image instead of just a black background.
2) Some information from the quote - name, number of co-insured, address (if there is one) - instead of Hedvig USP's.
3) Decreased height of the hero.
4) New text key `HERO_OFFER_DETAILS_HEADER` and translations.

_New design in Figma [here](https://www.figma.com/file/eRR5ACHtbsPWDd4eUauWv1/Onboarding-Offer-page?node-id=0%3A1&viewport=-431%2C505%2C0.9561541080474854), on the "Header module Non-switcher" page._

Other changes are some refactoring of the files I've been working with, most important being:
1) Removing the whole `useDocumentScroll` implementation to make the `Sidebar` component (the "CTA box") fixed on desktop but not mobile (see commit [bcd0e04](https://github.com/HedvigInsurance/web-onboarding/commit/bcd0e04a0fcd0ac9c6e95343bc4d0f7a74bbd5f0) for more details on these changes).
2) Aligning breakpoints/media queries for all the components involved since they need to be exactly the same if stuff's not to break:
  Create `mediaQueries.ts` in the root `/utils` folder with the media query for larger screens (`1000px` and up) as a string.
  Use this variable everywhere where this exact media query was used (all the components I've been working on here, that is).

## Why?

Mange redesigned this section with mainly the following problems in mind: a lot of users don't understand that the page is actually scrollable, and it's probably not clear enough that the quote is specifically created for that particular user. Also he simply made it prettier 😀 

The reason behind the refactoring is that I had to move stuff around in those files anyway, which made it a pretty convenient opportunity to remove some (in my view) unnecessary complexity and make other things a bit clearer (hopefully not only to me).

_Referenced ticket [here](https://hedvig.atlassian.net/browse/HVG-675)_

## Screenshots and GIF's

**Desktop:**

2021-01-11-offer-page-hero-desktop

2021-01-11-offer-page-hero

**A real actual iPhone 11:**

2021-01-11-offer-page-hero-iphone11

**"iPhone 6/7/8" according to Chrome dev tools:**

2021-01-11-offer-page-hero-iphone678


src/client/pages/OfferNew/Introduction/index.tsx
```diff
@@ -98,7 +96,7 @@ export const Introduction: React.FC<Props> = ({
         </Hero>
         <HeroContentWrapper>
           <ContentContainer>
-            <OfferDetails />
+            <HeroOfferDetails {...{ offerData }} />
```
virrios, line 99: offerData={offerData}?

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const textKeys = useTextKeys()
+  return (
+    <Container>
+      <Header>{textKeys.HERO_OFFER_DETAILS_HEADER()}</Header>
+      <OfferInfo>
+        {person.firstName}
+        {!!numberCoInsured && ` +${numberCoInsured}`}
+      </OfferInfo>
+      {'street' in quotes[0].quoteDetails && (
```
virrios, line 60: In a previous ticket I asked palmen if we could assume that the first quote in a quote bundle is the "main" one. He said that it might not be the case. This is a something that maybe not likely to happen but if you have a Norwegian quote bundle and NorwegianTravel is the first one then it will not display the information.

src/client/pages/OfferNew/Introduction/index.tsx
styrelius: You're right, that's how we write it! 🧠  Just an old habit of mine slipping through I guess (or however one puts it)

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
styrelius: Hm, that's a very good point! 💡 Thank you! Will fix that.

src/client/pages/OfferNew/Introduction/index.tsx
```diff
@@ -59,55 +77,33 @@ export const Introduction: React.FC<Props> = ({
...
+        <Hero>
+          <HeroImage
+            alt="laptop grip"
+            onLoad={() => setHasHeroImageLoaded(true)}
+            hasLoaded={hasHeroImageLoaded}
+            src="/new-member-assets/landing/laptop_grip_small.jpg"
+            sizes="100vw"
+            srcSet="
+          /new-member-assets/landing/laptop_grip_small.jpg 1600w,
+          /new-member-assets/landing/laptop_grip_medium.jpg 2200w"
```
gustaveen, line 94: In my head when picturing a Hero it makes most sense when having the content of the Hero inside it as well. In this case it would mean renaming the `Wrapper` to `Hero` instead. Not a big deal and the Wrapper was there from before, just something I got stuck on 😄

styrelius (commented):
Haha what? I didn't start a review I tried to answer Gustav's comment and now I can't close this 🙈

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
```diff
@@ -0,0 +1,65 @@
...
+}
+
+const Container = styled.div`
+  width: 100%;
+  padding: 2rem 0;
+  color: ${colorsV3.white};
+
+  ${LARGE_SCREEN_MEDIA_QUERY} {
+    width: auto;
+    padding: 0;
```
gustaveen, line 20: Is there a reason we set width here? Couldn't see any layout change without it :)

src/client/pages/OfferNew/Introduction/index.tsx
```diff
@@ -1,56 +1,74 @@
...
+  overflow: hidden;
+  position: absolute;
+`
+
+const HeroImage = styled.img<HeroImageProps>`
+  height: 1000px;
+  object-fit: cover;
+  object-position: -550px -100px;
+  opacity: ${({ hasLoaded }) => (hasLoaded ? 0.5 : 0)};
+  transition: opacity 0.8s;
```
gustaveen, line 43: Should we have some kind of easing, like `ease-out`?

gustaveen (commented):
Wohoo, nicely done Ellen! 🥂 Just some small comments and questions :)

src/client/pages/OfferNew/Introduction/index.tsx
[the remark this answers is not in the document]
styrelius: Sure, I'll change it

src/client/pages/OfferNew/components.tsx
```diff
@@ -1,10 +1,12 @@
 import styled from '@emotion/styled'
 import { colorsV2, fonts } from '@hedviginsurance/brand'
 import React from 'react'
+import { LARGE_SCREEN_MEDIA_QUERY } from 'utils/mediaQueries'
+import { SIDEBAR_WIDTH } from './Introduction/Sidebar/index'
 
 export * from './common/price'
 
-export const Heading = styled('h1')`
+export const Heading = styled('h2')`
```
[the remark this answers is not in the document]
styrelius: 😊

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
```diff
@@ -0,0 +1,65 @@
...
+  }
+`
+
+const Header = styled.div`
```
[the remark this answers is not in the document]
styrelius: I agree it should probably be called `Heading` yes! 👍 

Regarding the semantics I made a decision to _not_ use semantic HTML since we don't have a strategy on that and in my mind it was better to just not use it at all until we actually have that. But maybe that goes against the overall style in the code base.

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
```diff
@@ -0,0 +1,65 @@
...
+  text-transform: uppercase;
+  font-size: 1rem;
+
+  ${LARGE_SCREEN_MEDIA_QUERY} {
+    font-size: 1rem;
+  }
+`
+
+const OfferInfo = styled.div`
```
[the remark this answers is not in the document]
styrelius: See previous comment ☝️

src/client/pages/OfferNew/Introduction/index.tsx
```diff
@@ -1,56 +1,74 @@
...
 
-  @media (max-width: 600px) {
-    padding: 7rem 0 5rem 0;
+  ${LARGE_SCREEN_MEDIA_QUERY} {
+    object-position: right -88px;
```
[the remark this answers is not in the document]
styrelius: Ah! Nice catch - I don't have  a monitor at home so couldn't easily check on a very large screen. So this 👇  will fix it?
```javascript
 ${LARGE_SCREEN_MEDIA_QUERY} {
    height: auto;
    object-position: right -5vw;
```

src/client/pages/OfferNew/Introduction/index.tsx
styrelius: Sure, I can add that. Believe it or not - I don't have a strong opinion here! 😄

src/client/pages/OfferNew/Introduction/index.tsx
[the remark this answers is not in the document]
gustaveen: No worries, hard to notice without it! Yes indeed 🙌 And this might be one of the few times safari might be of use :D you can check with their "Responsive design mode"
Screenshot 2021-01-12 at 16 43 11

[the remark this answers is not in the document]
styrelius: What??!! I had no idea 🤯  Thank you! 👑

src/client/pages/OfferNew/Introduction/index.tsx
styrelius: But interestingly the default is apparently `ease` so it's not linear as it is now. Would you rather have `ease-out` than `ease` though?

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
styrelius: Yes, the reason is that if you make the screen smaller this element pushes the CTA box outside of the screen if it has `width: 100%`. Like this:
image

gustaveen: Definitely! I meant that it looks the same when removing `width: 100%` as well on line 14, in smaller screens that is. But if there's a reason, omit this comment 🙌

src/client/pages/OfferNew/Introduction/index.tsx
gustaveen: Aha nice, my gut feeling was that it was `linear`, that makes sense though! I tend to use `ease-out` for fading in stuff. I usually have a sneak peek on how material design resonates about easings which I find helpful 🤗https://material.io/archive/guidelines/motion/duration-easing.html#duration-easing-natural-easing-curves

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
styrelius: Ah yes, you're right! Fixing that. Thank you! ✨

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  useEffect(() => {
+    const householdSize = getHouseholdSize(quotes[0].quoteDetails)
+    setNumberCoInsured(householdSize - 1)
+
+    const quoteWithStreet = quotes.find((quote) => {
+      return 'street' in quote.quoteDetails
+    })
+    if (quoteWithStreet) {
+      setStreet((quoteWithStreet as QuoteWithStreet).quoteDetails.street)
```
virrios, line 60: Since Norwegian travel does not have street in quote details I think you have to  have an assertion or a type guard.
I think this is a nice solution! Could not think of a more compact or easier way to do it! 👍

styrelius: Yes I know, still think TypeScript _should_ understand that the quote we're dealing with at this point actually has `street` in it since I do that check on line 58. But I guess that's having too high expectations 🤷‍♀️ Anyway, tank you! 😊

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
```diff
@@ -0,0 +1,76 @@
...
+  word-break: break-all;
+
+  ${LARGE_SCREEN_MEDIA_QUERY} {
+    font-size: 3rem;
+  }
+`
+
+export const HeroOfferDetails: React.FC<Props> = ({ offerData }) => {
+  const [numberCoInsured, setNumberCoInsured] = useState<number | null>(null)
+  const [street, setStreet] = useState<Address['street'] | null>(null)
```
virrios, line 48: I see the point of having Address['street'] but might it not be sufficient with string? Just a thought, not strongly opinionated

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
gustaveen: I agree, this does the job for sure! 👌

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
gustaveen: Ah, I see your point, `Address['street']` is equal to a string. Either or works for me, having just string would remove one import :)

styrelius: I wanted it to be clear that it's not just any string. I guess my thinking is that everything that we get from the back-end should also be typed accordingly, with the types generated from the GraphQL schema.

styrelius: So I'll leave it as is 😄

gustaveen (approved):
Amazing, this will be such a nice improvement! 🎉

PR #461, opened 2021-01-13
Check if Intercom is defined

gustaveen:
## What?
Potential fix for bug where Intercom was undefined in variations for iOS, Android and Avy

## Why?
Offerpage in Hedvig app (norwegian onboarding) and Avy onboarding was broken


styrelius (approved):
🚒

PR #463, opened 2021-01-15
Put back price comparison for switcher

gustaveen:
## What?

Add price comparison for switchers

## Why?

Intermediate fix until we have de new design in place for price comparison. Not the most beautiful thing but it should do the trick over the weekend 😄

Commit for reference: https://github.com/HedvigInsurance/web-onboarding/pull/460/commits/ff4d1d852c7db7297446cadf74a1ffbc8f4ed151
Unlike the previous implementation, we don't need to shift content in mobile and desktop.

Will need someone to verify this that can compare insurances, setting up a review app. I mocked the data locally

Screenshot 2021-01-15 at 15 22 49

Screenshot 2021-01-15 at 15 26 06

Screenshot 2021-01-15 at 15 27 33

Screenshot 2021-01-15 at 15 27 45

### DIFF
--- a/src/client/pages/OfferNew/Introduction/index.tsx
+++ b/src/client/pages/OfferNew/Introduction/index.tsx
@@ -23,16 +23,20 @@ const HERO_HEIGHT = '400px'
 
 const Hero = styled.div`
   width: 100%;
-  height: ${HERO_HEIGHT};
   display: flex;
   flex-direction: column;
   align-items: center;
+  background-color: ${colorsV3.gray100};
+
+  ${LARGE_SCREEN_MEDIA_QUERY} {
+    height: ${HERO_HEIGHT};
+  }
 `
 
 const HeroImageWrapper = styled.div`
   width: 100vw;
   height: ${HERO_HEIGHT};
-  background: ${colorsV3.gray900};
+  background-color: ${colorsV3.gray900};
   overflow: hidden;
   position: absolute;
 `
@@ -53,7 +57,7 @@ const HeroImage = styled.img<HeroImageProps>`
 const HeroContentWrapper = styled.div`
   width: 100%;
   max-width: 80rem;
-  position: absolute;
+  position: relative;
   padding-top: 3rem;
 
   ${LARGE_SCREEN_MEDIA_QUERY} {
--- a/src/client/pages/OfferNew/Introduction/index.tsx
+++ b/src/client/pages/OfferNew/Introduction/index.tsx
@@ -4,8 +4,10 @@ import { colorsV3 } from '@hedviginsurance/brand'
 import { Section } from 'pages/OfferNew/components'
 import { OfferData } from 'pages/OfferNew/types'
 import { LARGE_SCREEN_MEDIA_QUERY } from 'utils/mediaQueries'
+import { isBundle } from '../utils'
 import { HeroOfferDetails } from './HeroOfferDetails'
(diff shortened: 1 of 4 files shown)


src/client/pages/OfferNew/Introduction/index.tsx
```diff
@@ -97,7 +106,15 @@ export const Introduction: React.FC<Props> = ({
         </HeroImageWrapper>
         <HeroContentWrapper>
           <ContentContainer>
-            <HeroOfferDetails offerData={offerData} />
+            <HeroOfferDetailsContainer>
+              <HeroOfferDetails offerData={offerData} />
+              {hasDataCollection && (
+                <ExternalInsuranceProvider
+                  dataCollectionId={offerData.quotes[0].dataCollectionId || ''}
```
virrios, line 113: I know that this is the old stuff back and its a fix. However, just have to ask: Is it okay to just take the first one? It just needs an Id it does not matter if its a Home or travel?

virrios (commented):
Nice! 🥇  just have a question

src/client/pages/OfferNew/Introduction/index.tsx
styrelius: We only have Insurely implemented in Sweden, where you always only create one quote. So in this case it should be fine.

styrelius (commented):
Nice! And thank you! And sorry 🙈 

The only thing was this with the CTA box covering the page section following the header in mobile, which I sent you a screenshot of.

gustaveen:
Update
- Removed Trustpilot scores in compare boxes
- Make the content height of Hero responsive to make sure CTA box does not cover any content on the page


https://user-images.githubusercontent.com/6661511/105012049-ff708080-5a3d-11eb-81de-7e5e4ffec7bf.mov

Screenshot 2021-01-19 at 09 57 44

src/client/pages/OfferNew/Introduction/index.tsx
```diff
@@ -21,16 +23,20 @@ const HERO_HEIGHT = '400px'
...
+  ${LARGE_SCREEN_MEDIA_QUERY} {
+    height: ${HERO_HEIGHT};
+  }
 `
 
 const HeroImageWrapper = styled.div`
   width: 100vw;
   height: ${HERO_HEIGHT};
-  background: ${colorsV3.gray900};
+  background-color: ${colorsV3.gray900};
```
styrelius, line 39: What's the point of using the `background-color` property instead of the shorthand `background`? Also, we do use `background` for setting the background color in like 60 places in the repo.

styrelius (commented):
Nicely done! 🌟  Just one very minor question

styrelius (approved):

src/client/pages/OfferNew/Introduction/index.tsx
gustaveen: Good question! 😃 This is a force of habit for me, since the shorthand sets a few other properties to it's default values if you don't specify them, which can cause unwanted behaviour. In most cases it will be fine, but I have found myself having to change this in previous projects when finding bugs, so I try to stick to setting the specific value :) This is of course just an opinion. Being the nerd that I am about this I think this post points out the pros and cons 🤓
https://csswizardry.com/2016/12/css-shorthand-syntax-considered-an-anti-pattern/

PR #464, opened 2021-01-19
[HVG-804] Fix / GTM scripts

styrelius:
## What?

1. Attempt to fix some tracking script from `aff.addreax.com` by adding `https://` to it in the CSP `scriptSrc` in `server/middleware/helmet.ts`.
2. Add two new url's to be whitelisted: `'https://bat.bing.com/bat.js'` and `'https://analytics.tiktok.com'`.

## Why?

Apparently scripts from `aff.addreax.com` are not doing the tracking they should. This probably won't fix it but it's worth testing.

And yes, we're on TikTok now. 🙃

_Referenced ticket [here](https://hedvig.atlassian.net/browse/HVG-804)_


gustaveen:
Oh wow, that's the closest I come to TikTok so far! 👴🏻

src/server/middleware/helmet.ts
```diff
@@ -61,10 +61,12 @@ const defaultSrc = [
   'https://api-js.mixpanel.com',
   'checkoutshopper-live.adyen.com',
   'checkoutshopper-test.adyen.com',
-  'aff.addreax.com',
+  'https://aff.addreax.com',
   'js.go2sdk.com',
   'secure.adnxs.com',
   'snap.licdn.com',
+  'https://bat.bing.com/bat.js',
```
gustaveen, line 68: When trying out this locally, this still got blocked, it only wants the domain :)
```suggestion
  'https://bat.bing.com',
```

gustaveen:
You might already know, but you can test this locally by setting the env variable `USE_HELMET=true` which is neat :) We also need to add `'https://bat.bing.com'` to `imgSrc`, since the script it's adding a tracking pixel as well 🥴

src/server/middleware/helmet.ts
styrelius: Ah, of course! 🧠

styrelius:
No I did not know that! Thank you! 🙏  And I'll fix the `imgSrc`

gustaveen (approved):
Gotta love CSP work! 🥴

PR #465, opened 2021-01-20
Intercom add to connect payment and fix linter warnings

virrios:
[HVG-765] Connect Payment>Add Intercom 

## What?
Adding intercom to the connect-screen

## Why?
Its useful for the user be able to contact IEX From connect payment view if for instance they have questions 
or something went wrong

_Referenced ticket [https://hedvig.atlassian.net/browse/HVG-765]()_

### DIFF
--- a/src/client/utils/Intercom.tsx
+++ b/src/client/utils/Intercom.tsx
@@ -23,7 +23,8 @@ const createIntercom = () => {
 export const Intercom: React.FC = () => {
   const variation = useVariation()
   const isIntercomMatch = useRouteMatch(
-    LOCALE_PATH_PATTERN + '/new-member/:place(offer|sign|download)',
+    LOCALE_PATH_PATTERN +
+      '/new-member/:place(offer|sign|download|connect-payment)',
   )
   const isDirectConnectPaymentIntercomMatch = useRouteMatch(
     LOCALE_PATH_PATTERN + '/new-member/connect-payment/direct',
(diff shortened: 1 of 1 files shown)


src/client/utils/Intercom.tsx
```diff
@@ -23,7 +23,8 @@ const createIntercom = () => {
 export const Intercom: React.FC = () => {
   const variation = useVariation()
   const isIntercomMatch = useRouteMatch(
-    LOCALE_PATH_PATTERN + '/new-member/:place(offer|sign|download)',
+    LOCALE_PATH_PATTERN +
+      '/new-member/:place(offer|sign|download|connect-payment)',
   )
   const isDirectConnectPaymentIntercomMatch = useRouteMatch(
     LOCALE_PATH_PATTERN + '/new-member/connect-payment/direct',
```
styrelius, line 30: Is this thing needed anymore? Seems when I tested it locally like it's not, since this case is covered by the addition `connect-payment` to the `LOCALE_PATH_PATTERN`?

styrelius (commented):
Awesome! 🙌  And you are correct - the position of the Intercom "launcher" (button) should be the same as on all other pages.

But I think we can actually remove some code 👇

src/client/utils/Intercom.tsx
virrios: You are correct! 💯

styrelius (approved):
🌻

PR #466, opened 2021-01-20
Update copy for checkout button simple sign

gustaveen:
## What?

Update copy on sign button depending on sign method

## Why?

Clarify for the user if it's sign with bank id or not

[HVG-768]

https://user-images.githubusercontent.com/6661511/105160966-2350c680-5b11-11eb-9cbc-c8b314d26088.mov

[HVG-768]: https://hedvig.atlassian.net/browse/HVG-768


virrios:
Niiice! 🥇  
Tested Sweden and Norway but I have forgotten how to test for danish users

src/client/pages/ConnectPayment/sections/ConnectPayment.tsx
```diff
@@ -95,10 +95,11 @@ const HeaderPart = styled.span<HeaderPartProps>`
 `
 const ConnectText = styled.div`
   font-size: 1rem;
+  line-height: 1.5rem;
   color: ${colorsV3.white};
 
   ${MOBILE_SCREEN_MEDIA_QUERY} {
-    font-size: 1.25rem;
+    font-size: 1.125rem;
```
virrios, line 102: I designen är de 24px, tror jag körde 20 efter lite med Mange, jag tycker det ser bra ut med 18px också så det är väl lite hur vi vill göra

styrelius: Neither `18px` nor `20px` is among the font sizes that are defined in the "Web UI Kit" Figma so maybe that should also be looked over then. 
That being said I think 18px / `1.125rem` looks good here.

virrios: Exactly! I think I got the size 20px from when me and Mange sat and experimented live. But I also agree that 18px looks niice

gustaveen: Vaild points, I pointed that out to Mange that it doesn't follow the Figma design :D We have 18px as size for mobile but so it can be discussed if we should use it here, but I think it's fine since the gap from 16px to 24px can be a bit too much.

gustaveen: Or we can revert, either or we can change this later 👍

virrios: I think it looks good with 18px, as long as Mange and everyone is happy 😄

virrios (dismissed):
Super! 🚀

PR #467, opened 2021-01-20
Fixed height by adding height: 100%  property

virrios:
Fix Image looking strange on Iphone simulator

**Please test staging with you iphone if this problem exists!**

## What?
Added height property

## Why?

**Without height:**
image

**With:** 

image


gustaveen (approved):
Nice catch! 🙌

PR #468, opened 2021-01-25
[HVG-718] [HVG-720] Feature / New /download page design

styrelius:
## What?

New design of the `/download` page! 😀 

1) All components making up the page are replaced with new ones, including the App Store and Google Play icons, and the image. So this means a lot of deleted files, as well as a bunch of completely new ones.
2) New copy - which means new translations and `textKeys`.
3) Conditional rendering based on if the user has a Hedvig Forever referral code is removed since the new design is the same for everyone.
4) New hook `useBreakpoint` in `src/client/utils/hooks/` which listens to the `'resize'` event on Window in order to return whether we're dealing with a large screen or not. It uses the breakpoints declared in `mediaQueries.ts` and can easily be extended with new breakpoints and return values. 
This hook is meant to replace the `is-mobile` and `react-responsive` packages (yes we use two different packages to check for screen size/type of device with JS and no they don't do _exactly_ the same thing but we still don't seem to need both. Or any) that we're currently using since I think this hook is easier to use (and I trust you to tell me if you don't agree 😄).
5) Merge conflict fixes and variable name changes in `mediaQueries.ts`. This meant also changing some imported variabels in `ConnectPayment.tsx` and `TrustlyCHeckout.tsx`.

## Why?

New fresh design from Mange etc!

Also, in the case of the `useBreakpoint` hook I think stuff you can _easily_ write yourself _might_ be beneficial to do that way instead of having yet another dependency to maintain.

_Referenced tickets: [HVG-718](https://hedvig.atlassian.net/browse/HVG-718) and [HVG-720](https://hedvig.atlassian.net/browse/HVG-720)_
_[Design in Figma](https://www.figma.com/file/eRR5ACHtbsPWDd4eUauWv1/Onboarding-Offer-page?node-id=255%3A4253&viewport=-182%2C320%2C0.6568618416786194), on page "Success Screens ✅ "_

## Screenshots

**/se-en desktop**

2021-01-25-download-page-desktop-se-en

**/no desktop**

2021-01-25-download-page-desktop-no

**/se iPhone 11**

2021-01-25-download-page-mobile-se-1
2021-01-25-download-page-mobile-se-2


src/client/pages/Download/components/DownloadHeadline.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export const DownloadHeadline: React.FC = () => {
+  const textKeys = useTextKeys()
+
+  return (
+    <Headline>
+      <MainHeadline>
+        {textKeys.ONBOARDING_DOWNLOAD_HEADLINE_PART_1() + ' '}
+      </MainHeadline>
+      <span>{textKeys.ONBOARDING_DOWNLOAD_HEADLINE_PART_2()}</span>
```
virrios, line 39: Nitpick, but should this not be part of the h1?

styrelius: I did it this way to get the headline to line break in the right place in desktop but not line break at all in mobile. I thought other solutions to achieve this looked... weirder, so if it's not a deal breaker for you guys I'm going to keep it this way. 😀

src/client/utils/hooks/useBreakpoint.ts
```diff
@@ -0,0 +1,20 @@
...
+      setWindowWidth(window.innerWidth)
+    }
+
+    window.addEventListener('resize', onWindowResize)
+
+    return () => window.removeEventListener('resize', onWindowResize)
+  })
+
+  return {
+    isLargeScreen: windowWidth >= BREAKPOINTS.largeScreen,
```
virrios, line 18: nice! So we add more attributes here 👌 👍 when we need them

styrelius: Yes that was the idea! 😊

styrelius: Then it's more clearly tightly connected to the breakpoints we define in the `mediaQueries` file. Or that's at least what my thought was.

src/client/pages/Download/components/DownloadPageContent.tsx
```diff
@@ -0,0 +1,115 @@
...
+`
+
+const Paragraph = styled.div`
+  font-size: 1.125rem;
+  line-height: 24px;
+  color: ${colorsV3.gray100};
+  opacity: 0.65;
+
+  :not(:last-of-type) {
+    margin-bottom: 12px;
```
virrios, line 64: 0.75rem?

src/client/pages/Download/components/DownloadPageContent.tsx
```diff
@@ -0,0 +1,115 @@
...
+  ${LARGE_SCREEN_MEDIA_QUERY} {
+    padding-bottom: 4rem;
+  }
+`
+
+const Paragraph = styled.div`
+  font-size: 1.125rem;
+  line-height: 24px;
+  color: ${colorsV3.gray100};
+  opacity: 0.65;
```
virrios, line 61: color: ${colorsV3.gray500}; is #AAAAAA

src/client/pages/Download/index.tsx
```diff
@@ -1,55 +1,26 @@
...
-const TopBarWrapper = styled('div')`
-  color: ${colorsV3.black};
-`
+import { DownloadPageContent } from './components/DownloadPageContent'
 
 export const Download: React.FC = () => {
-  const campaigns = useRedeemedCampaignsQuery()
   const textKeys = useTextKeys()
 
-  if (campaigns.loading) {
```
virrios, line 27: We don't have any referrals here?

virrios (commented):
Such an awesome job! 💯
just some small stuff

src/client/pages/Download/components/DownloadHeadline.tsx
virrios: you mean like having a media query on span to display block/inline? 
It is absolute **_not_** a deal breaker for me 😺

src/client/pages/Download/components/DownloadPageContent.tsx
styrelius: 👍 Just took the styles from the old implementation but you are right - that does make more sense 😄

src/client/pages/Download/components/DownloadPageContent.tsx
styrelius: Yes but we should actually have a discussion on the usage of `rem` altogether I think! The way we do it now kind of doesn't make sense to me. So I just skipped it here hehe 🙈

src/client/pages/Download/index.tsx
styrelius: Ah, no the way I interpreted the old code is that we had some different copy if it was a referral, but as I wrote in the description the new design is just the same for everyone. So I removed it.

virrios: sry missed that part in the description but okey the copy should be the same regardless so nothing important in DownloadAppHedvigForeverMember? (Did not check that file 😆 )

styrelius: OK I'll check again - I can see why you would want to really make sure I know what I'm doing when I remove stuff haha! 🙈

src/client/pages/Download/components/DownloadPageContent.tsx
styrelius: But I can see that it looks messy when `rem` and `px` are mixed. So I'll change it.

src/client/pages/Download/index.tsx
styrelius: No, `DownloadAppHedvigForeverMember` just had some other copy and for some reason a check which kind of mobile device the user's on so that the only link rendered would be the one to the right app place (like if you're on iPhone you would get one link - to App Store). Which I don't get why we had _only_  for referrals 🤷‍♀️  But feel free to check `src/client/pages/Download/sections/DownloadApp.tsx` out for yourself to see if I missed something.

src/client/pages/Download/components/DownloadHeadline.tsx
styrelius: _Nevermind_  I made the `div` an `h1` and of course if worked without changing `display` of `span`s since it's a flex box 🤦‍♀️

src/client/pages/Download/index.tsx
virrios: I thought maybe there was some referral special logic, but as you say  basically just copy and some conditional rendering which app icon to display. Haha its good to sanity check when removing stuff 😆

virrios:
I am super happy but maybe gustav wants to take a look before approving?

src/client/pages/Download/components/DownloadPageContent.tsx
```diff
@@ -0,0 +1,114 @@
...
+import { TOP_BAR_HEIGHT } from 'components/TopBar'
+import { DownloadHeadline } from './DownloadHeadline'
+import { GetAppButtons } from './GetAppButtons'
+import { AppImage } from './AppImage'
+
+const IMAGE_WIDTH = 560
+
+const Page = styled.div`
+  background: ${colorsV3.gray900};
+  max-width: 100vw;
```
gustaveen, line 16: Did you get any overflow issues, or why do we need to set a max-width here? Or maybe set width `100%`

src/client/pages/Download/components/DownloadPageContent.tsx
```diff
@@ -0,0 +1,114 @@
...
+`
+
+const ContentContainer = styled.div`
+  width: 100%;
+  max-width: 24rem;
+  padding-top: 2rem;
+  display: flex;
+  flex-direction: column;
+  align-items: center;
+  justify-content: center;
```
gustaveen, line 30: Do we need to set it to `display: flex` here, or can we achieve the same layout with `display: block` if we want a single column and then add `display: flex` in the breakpoint below :)

gustaveen (commented):
Beautiful work Ellen! 👩‍🎨 Just two minor thoughts

src/client/pages/Download/components/DownloadPageContent.tsx
styrelius: Yes since that's the best way to make all the content centered I think 😊 You don't see this on mobile since all elements in the `ContentContainer` has full width then, but for smaller screens that are not _that_ small it matters.

styrelius: Although the `justify-content: center;` might not make sense 🙈

src/client/pages/Download/components/DownloadPageContent.tsx
styrelius: I don't remember, it might just have been that I have experience of overflow issues when just setting width to `100%` on elements with parent elements with widths that are not defined.

gustaveen (approved):
🎉

src/client/pages/Download/components/DownloadPageContent.tsx
styrelius: Lol turned out you were right of course, this isn't needed. Will change that in some other PR. It's really awesome that I was _so_ sure I was right. 🤦‍♀️  Lesson hopefully learned - always check again before making claims like this! 💡

PR #470, opened 2021-01-27
Feature/simple sign login screen

gustaveen:
## What?

Sign in screen used in the app to sign in for users that that signed with Simple Sign. The page is only accessible for NO at the moment but might b added for DK later. We do frontend validation to check the pattern and length of the ssn.

## Why?

In order to login after Simple Sign, signsec doesn't give us access to to ssn so we don't know what member to actually sign in. To solve this we add an intermediate step on the web where the user inputs the ssn, we call `norwegianBankIdAuth` and get a redirect url back and send them back to the app.

Review app: https://web-onboardi-feature-si-uylata.herokuapp.com/no/login & https://web-onboardi-feature-si-uylata.herokuapp.com/no-en/login

[HVG-827]

https://user-images.githubusercontent.com/6661511/105994848-76de8980-60a8-11eb-90c4-8a0bf2c7d6a3.mov

### Error state

https://user-images.githubusercontent.com/6661511/105996920-45b38880-60ab-11eb-9bd5-3821dc73f31b.mov

[HVG-827]: https://hedvig.atlassian.net/browse/HVG-827


FredrikAreschoug (dismissed):
My web skills are not that great but I think I understand :100: Nice :slightly_smiling_face:

src/client/pages/LoginApp/components/LoginForm.tsx
```diff
@@ -0,0 +1,153 @@
...
+                errors={touched.ssn && errors.ssn ? textKeys[errors.ssn]() : ''}
+                onChange={(e: React.ChangeEvent<HTMLInputElement>) =>
+                  handleInputChange(e, setFieldValue)
+                }
+                onBlur={handleBlur}
+                variant="dark"
+              />
+            </SsnField>
+            <SubmitButton
+              background={colorsV3.purple500}
```
virrios, line 137: nitpick, background not necessary

src/client/pages/LoginApp/components/LoginForm.tsx
```diff
@@ -0,0 +1,153 @@
...
+          values,
+          isSubmitting,
+          setFieldValue,
+          handleBlur,
+        }) => (
+          <LoginFormikForm>
+            <Heading>{textKeys.LOGIN_APP_HEADING()}</Heading>
+
+            <SsnField>
+              <InputField
```
virrios, line 117: When using autocomplete in the form it looks a bit wierd, the user agent sets the background.

image

src/client/pages/LoginApp/LoginAppScreen.tsx
```diff
@@ -0,0 +1,83 @@
...
+        return
+      }
+
+      if (result.data?.norwegianBankIdAuth.redirectUrl) {
+        window.location.assign(result.data?.norwegianBankIdAuth.redirectUrl)
+      }
+    } catch (e) {
+      actions.setErrors({ ssn: 'LOGIN_APP_GENERIC_ERROR' })
+      captureSentryError(e)
+    }
```
styrelius, line 63: Don't we get the `result` as a state object from the `useNorwegianBankIdAuthMutation()`? I thought with Apollo you just execute the mutation and then use the state(s) you get back, in like a `useEffect`?
```javascript
  const [norwegianBankIdAuthMutation, { data, error }] = useNorwegianBankIdAuthMutation()
```

src/client/pages/LoginApp/index.tsx
```diff
@@ -0,0 +1,23 @@
+import React from 'react'
+import { Market, useMarket } from 'components/utils/CurrentLocale'
+import { LoadingPage } from 'components/LoadingPage'
+
+const LoginAppScreen = React.lazy(() =>
+  import(/* webpackChunkName: "debugger" */ './LoginAppScreen').then((m) => ({
+    default: m.LoginAppScreen,
+  })),
+)
```
styrelius, line 9: Just a question since I'm not very familiar with `React.lazy` or `Suspense` - why do we need to lazy load this particular component? Will this make this component not part of the main bundle or something? It's a pretty small component so I guess I don't really see why

src/translations/en.json
```diff
@@ -17,8 +17,6 @@
   "ONBOARDING_CONNECT_DD_TRUSTLY_MODAL_TITLE": "Set up payment",
   "ONBOARDING_CONNECT_DD_TRUSTLY_MODAL_FAIL_BODY": "Something went wrong",
   "ONBOARDING_CONNECT_DD_TRUSTLY_MODAL_FAIL_CTA": "Try again",
-  "ONBOARDING_DOWNLOAD_PRE_HEADLINE": "Welcome!",
-  "ONBOARDING_DOWNLOAD_HEADLINE": "Download Hedvig to get started!",
```
styrelius, line 21: Just wanted to mention that we need to deploy the changes in my PR at the same time if we don't want missing text keys on `/download` page.

styrelius (commented):
Nice! ✨  Had some questions though. 
And if the idea is that this will be very short lived in our code base since the issues will be solved in the apps then maybe my first two comments are redundant but if there's a risk of this being here for a while then I think we should look at them.

src/client/pages/LoginApp/LoginAppScreen.tsx
gustaveen: Good question, I'm not sure what's the preferred way, we have it this way in a bunch of places when we trigger it when submitting, like in `AdyenCheckout`

src/client/pages/LoginApp/components/LoginForm.tsx
gustaveen: very true, good catch! 🙌

src/client/pages/LoginApp/components/LoginForm.tsx
gustaveen: ah, might be a good idea to turn of the autocomplete 👍

src/translations/en.json
gustaveen: Oh my bad, should have only commited the relevant translations, let's see if I can fix that :)

src/client/pages/LoginApp/index.tsx
gustaveen: You're absolutely right, this will not make it part of the bundle. I did this to not load the LoginScreen for non Norwegian markets instead of rendering the component and showing an empty component. And I think we can enable code splitting more in the app, even if this small component itself won't make a huge difference :)

src/client/pages/LoginApp/components/LoginForm.tsx
gustaveen: or as you say, this should not be a problem using it from the app

src/translations/en.json
styrelius: No but it's tricky the way we use Lokalise now - especially when you remove or rename text keys.. 🤔 If you don't immediately merge those changes then the risk is really high that this happens. One way of doing it would be to not remove any old text keys until the new ones are merged. But then there's the risk of forgetting to do that and as a result having a bunch of unused text keys in Lokalise and in our translations files. 🤷‍♀️

src/client/pages/LoginApp/LoginAppScreen.tsx
styrelius: After looking at [Apollo's docs](https://www.apollographql.com/docs/tutorial/mutations/#apply-the-usemutation-hook) I'm pretty sure the preferred way is actually what I described - using the state object that is returned from the mutation hook. You could also use an `onCompleted` callback but that might be unnecessarily complicated.

src/client/pages/LoginApp/index.tsx
styrelius: I kind of don't really agree 😄 I think all things that brings more complexity to the codebase, like code splitting, should only be done when it's _really needed_. And it doesn't seem like it is in this case. I would suggest a simpler way, like:
```javascript
return market === Market.No ? <LoginAppScreen /> : null
```
Or is it possible to just say in the routes file that this particular route should only be available for the Norwegian market?

styrelius: That said, this is not a blocker since I guess it's pretty urgent to get this to prod. But as a rule of thumb I don't think we should make things more complex than they need to be (and I have no idea if we would actually _really need_ to do code splitting somewhere else in the repo, I mean maybe, but then we should have that as a joint strategy)!

gustaveen: I’m all for not making things more complicated 😁 That being said I don’t necessarily think that code splitting makes it more complicated. But let’s remove it here and we can discuss when and what to use it for 🙌🏻 I’d like to keep the routes file clean without too many conditions so let’s do it like you suggested then 👍

src/client/pages/LoginApp/LoginAppScreen.tsx
gustaveen: Ok, I guess we can change it if we feel that this is the wrong way 🙃

src/client/pages/LoginApp/index.tsx
styrelius: Awesome! 🙌  And it's a good point to keep the routes file as clean as possible 👍
And I'd love to learn more about code splitting and see where it could be useful for us!

src/client/pages/LoginApp/LoginAppScreen.tsx
gustaveen: After testing this I don't see a major benefit doing it this way. Since I still want to access `setErrors` in `handleSubmit`. If you guys want to amend this tomorrow, feel free and go ahead, as long as we don't delay this feature any longer please 🙏
```
  const [
    norwegianBankIdAuthMutation,
    { data },
  ] = useNorwegianBankIdAuthMutation()

  if (data?.norwegianBankIdAuth.redirectUrl) {
    window.location.assign(data.norwegianBankIdAuth.redirectUrl)
  }

  useEffect(() => {
    if (!hasToken) {
      return
    }
  }, [hasToken])

  const handleSubmit = async (
    form: LoginFormValue,
    actions: FormikHelpers<LoginFormValue>,
  ) => {
    const { ssn } = form

    try {
      await norwegianBankIdAuthMutation({
        variables: { personalNumber: ssn },
      })
    } catch (e) {
      actions.setErrors({ ssn: 'LOGIN_APP_GENERIC_ERROR' })
      captureSentryError(e)
    }
  }
```

virrios: Yes since this is tested and verified and we want to get it out now, we can change this later if we want to

styrelius: Hm, I don't think I expressed myself clearly enough. 🙈  But never mind, we'll merge this without any changes to what you wrote initially.

styrelius (approved):
👍 🆒

PR #471, opened 2021-02-04
[HVG-908] Add script to format generated code after its creation

styrelius:
## What?

1. Change name of `gen` script to `codegen` since I think `gen` is a bit too generic (no pun intended 🙈) 
2. Add `codegen-formatted` script which formats the generated file with Prettier once it's been created.

## Why?

Since the format of the file created with the GraphQL Code Generator differs from the format we enforce with Prettier in our code base, it will look as if almost everything changed when you re-generate this file `src/client/data/graphql.tsx`. Which can be very confusing.

## Question

Would we rather like to have the formatting of this file to be run in the `codegen` script instead of having a separate script that does it?

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,8 +17,7 @@
     "test": "NODE_ENV=test jest",
     "typecheck": "tsc --noEmit",
     "lint": "eslint 'src/**/*.{ts,tsx}'",
-    "codegen": "graphql-codegen --config codegen.yml",
-    "codegen-formatted": "yarn codegen && yarn prettier --write src/client/data/graphql.tsx",
+    "codegen": "graphql-codegen --config codegen.yml && yarn prettier --write src/client/data/graphql.tsx",
     "storybook": "start-storybook -p 6008",
     "build-storybook": "build-storybook",
     "download-translations": "./bin/download-translations.sh"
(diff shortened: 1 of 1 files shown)


gustaveen (dismissed):
Ah, thank you for taking the initiative and doing this! 😍  I wouldn't mind having the `codegen` script run prettier as you suggested to remove the risk of checking in unformatted code. Feel free to change 🙌

styrelius:
> Ah, thank you for taking the initiative and doing this! 😍 I wouldn't mind having the `codegen` script run prettier as you suggested to remove the risk of checking in unformatted code. Feel free to change 🙌

Since we have a `lint-staged` pre-commit git hook that runs `prettier --write` on all staged files that risk doesn't really exist though? But that said - would we ever want to run the `codegen` script and _not_ have the generated file formatted correctly? Probably not, right?

gustaveen:
True, that wouldn't be a risk! 🙌  But I agree, why would you want to run it without it 👍  
Anyhow, awesome you added it! 🎉

styrelius:
OK, good, then I'll change it to just be baked into the `codegen` script! 👍 😀

gustaveen (approved):
🎉

PR #472, opened 2021-02-04
[HVG-907] Fix / External insurance data error

styrelius:
## What?

Add a new offer page component `StartDateLabelSwitcher.tsx` and move a bunch of things from `StartDate.tsx` there.
This component, which is then used in `StartDate`, will only be rendered if the quote property `dataCollectionId` has a value (since this means the user has a current insurer that we know of), and from here the `useExternalInsuranceDataQuery` will be called. 

This means we will only try to query for more data on the user's current insurance when we know that there really is one, instead of alway calling this hook when the user goes to `/offer` page.

Also add a `TODO` in `ExternalInsuranceData.graphql` to replace deprecated fields in this query when we implement Insurely in Norway, since it won't work otherwise.

## Why?

In order to get rid of the error that we now get when a user who doesn't have an external (meaning current) insurer goes to the `/offer` page, i.e. _every_ time someone goes through the regular ("needer") flow and not the switcher flow:

image

Jira issue: [HVG-907]

## Screenshot and GIF
**I'm not able to test with a quote where the switcher flow succeeded but it looks and works the same as before in the failed switcher flow at least**

2021-02-04-failde-switcher-offer

2021-02-04-start-date-failed-switcher

[HVG-907]: https://hedvig.atlassian.net/browse/HVG-907


src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
```diff
@@ -313,10 +254,15 @@ const DateForm: React.FC<{
...
+          <DateLabel>
+            {!hasStartDate && quote.dataCollectionId && (
+              <StartDateLabelSwitcher
+                dataCollectionId={quote.dataCollectionId}
+              />
+            )}
+            {hasStartDate && getDateLabel()}
           </DateLabel>
-          {hasStartDate && <DownArrow />}
+          <DownArrow />
```
gustaveen, line 265: So we hid `DownArrow` before if it was a Switcher? Don't understand why so I guess it's correct to remove it 😄

src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
```diff
@@ -237,35 +203,9 @@ const DateForm: React.FC<{
...
-            })}
-          </DataCollectedStartDateDescription>
-        </DataCollectedStartDateWrapper>
-      )
-    }
-
-    return textKeys.SIDEBAR_STARTDATE_CELL_VALUE_SWITCHER()
   }
 
+  // calendar, should be broken out to a regular component
```
gustaveen, line 208: Maybe add a `TODO:` here as well?

gustaveen (commented):
Awesome to get rid of that `error`! 🙏  As always a bit of a hustle to test the switcher flow but as far as I see it's all good! 🙌

src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
styrelius: Ah yes - sorry should have added that to the description of the PR! 🙈 But yes, I think it makes sense to indicate that you can still pick another date. And without the arrow that's _really_ unclear.

styrelius: So I removed that conditional.

src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
styrelius: Yes! 👌  Or just remove the comment maybe. Think I left it by accident.

styrelius: Realised why I left it there, so remade it into a `TODO`.

gustaveen (approved):

PR #473, opened 2021-02-05
Remove SessionContainer on login page

gustaveen:
## What?

Make sure the simple sign login sends the correct token. 

## Why?

Since we had wrapped the component in a `SessionContainer`, it creates a new session and overwrites the token we get from the apps.


styrelius (approved):
🎆 👑 🌟

PR #478, opened 2021-02-12
HVG-921: App Offer: Quote IDs-adapter

hedvigoscar:
## What?

This PR adds an adapter that the apps can use to insert an array of quote IDs,
using query parameters on the following format:

`?quoteIds=[id1, id2, ...]`

The procedure for retrieving the quote IDs goes as follows:

- Is there quote IDs in the session? If yes, use that
- Is there quote IDs in the query parameters? If yes, parse them, add them to
  the session and use those
- If these fail, use the same failure mode as before

## Why?

So that the apps can utilize the Offer page, while using their own version of
Embark.

_Referenced ticket [here](https://hedvig.atlassian.net/browse/HVG-921)_

### DIFF
--- a/src/client/pages/OfferNew/index.tsx
+++ b/src/client/pages/OfferNew/index.tsx
@@ -27,6 +27,7 @@ import { Checkout } from './Checkout'
 import { FaqSection } from './FaqSection'
 import { Introduction } from './Introduction'
 import { Perils } from './Perils'
+import { useQuoteIds } from './quote-ids'
 
 const createToggleCheckout = (history: History<any>, locale?: string) => (
   isOpen: boolean,
@@ -44,24 +45,28 @@ export const OfferNew: React.FC = () => {
   const localeIsoCode = getLocaleIsoCode(currentLocale)
   const currentMarket = useMarket()
   const variation = useVariation()
-  const quoteIds = storage.session.getSession()?.quoteIds ?? []
+  const history = useHistory()
   const { data: redeemedCampaignsData } = useRedeemedCampaignsQuery()
   const redeemedCampaigns = redeemedCampaignsData?.redeemedCampaigns ?? []
+  const quoteIds = useQuoteIds(storage, history)
   const { data, loading: loadingQuoteBundle, refetch } = useQuoteBundleQuery({
     variables: {
       input: {
-        ids: [...quoteIds],
+        ids: [...(quoteIds ?? [])],
       },
       locale: localeIsoCode,
     },
+    skip: quoteIds === null,
   })
 
-  const history = useHistory()
   const checkoutMatch = useRouteMatch(
     '/:locale(se-en|se|no-en|no|dk-en|dk)/new-member/sign',
   )
   const toggleCheckout = createToggleCheckout(history, currentLocale)
 
+  if (quoteIds === null) {
+    return <LoadingPage />
+  }
   if (quoteIds.length === 0) {
     return <Redirect to={`/${currentLocale}/new-member`} />
   }
--- a/src/client/pages/OfferNew/index.tsx
+++ b/src/client/pages/OfferNew/index.tsx
@@ -46,15 +46,15 @@ export const OfferNew: React.FC = () => {
   const history = useHistory()
   const { data: redeemedCampaignsData } = useRedeemedCampaignsQuery()
   const redeemedCampaigns = redeemedCampaignsData?.redeemedCampaigns ?? []
-  const quoteIds = useQuoteIds()
+  const { isLoading: quoteIdsIsLoading, quoteIds } = useQuoteIds()
   const { data, loading: loadingQuoteBundle, refetch } = useQuoteBundleQuery({
     variables: {
       input: {
-        ids: [...(quoteIds ?? [])],
+        ids: [...quoteIds],
       },
       locale: localeIsoCode,
     },
-    skip: quoteIds === null,
+    skip: quoteIdsIsLoading,
   })
 
   const checkoutMatch = useRouteMatch(
@@ -62,11 +62,11 @@ export const OfferNew: React.FC = () => {
   )
   const toggleCheckout = createToggleCheckout(history, currentLocale)
 
-  if ((loadingQuoteBundle && !data?.quoteBundle) || quoteIds === null) {
+  if ((loadingQuoteBundle && !data?.quoteBundle) || quoteIdsIsLoading) {
     return <LoadingPage />
   }
 
-  if (quoteIds?.length === 0) {
+  if (quoteIds.length === 0) {
     return <Redirect to={`/${currentLocale}/new-member`} />
   }
 
--- a/src/client/pages/OfferNew/index.tsx
+++ b/src/client/pages/OfferNew/index.tsx
@@ -27,7 +27,7 @@ import { Checkout } from './Checkout'
 import { FaqSection } from './FaqSection'
 import { Introduction } from './Introduction'
 import { Perils } from './Perils'
-import { useQuoteIds } from './quote-ids'
+import { useQuoteIds } from './useQuoteIds'
 
 const createToggleCheckout = (history: History<any>, locale?: string) => (
   isOpen: boolean,
--- a/src/client/pages/OfferNew/index.tsx
+++ b/src/client/pages/OfferNew/index.tsx
@@ -20,7 +20,6 @@ import {
 import { SwitchSafetySection } from 'pages/OfferNew/SwitchSafetySection'
 import { getOfferData } from 'pages/OfferNew/utils'
 import { useVariation, Variation } from 'utils/hooks/useVariation'
-import { useStorage } from 'utils/StorageContainer'
 import { trackOfferGTM } from 'utils/tracking/gtm'
 import { getUtmParamsFromCookie, TrackAction } from 'utils/tracking/tracking'
 import { Checkout } from './Checkout'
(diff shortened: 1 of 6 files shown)


src/client/pages/OfferNew/quote-ids.ts
```diff
@@ -0,0 +1,44 @@
...
+
+  const { quoteIds: queryParamQuoteIds } = queryString.parse(
+    history.location.search.substr(1),
+  )
+
+  if (queryParamQuoteIds && typeof queryParamQuoteIds === 'string') {
+    const withoutBrackets = queryParamQuoteIds.replace(/[[\]]/g, '')
+    const splitted = withoutBrackets.split(',')
+    storage.session.setSession({
+      ...(storage.session.getSession() ?? {}),
```
styrelius, line 37: Seems to me like `?? {}` is not needed, the result with the spreading is the same if you spread `null` or `undefined` as when you spread `{}` (the result is `{}`).
```suggestion
      ...(storage.session.getSession()),
```

src/client/pages/OfferNew/quote-ids.ts
```diff
@@ -0,0 +1,44 @@
+import queryString from 'querystring'
```
styrelius, line 1: Would be awesome if this file was named `useQuoteIds` instead - as all other files exporting hooks are named the same as the hook itself 🙏

src/client/pages/OfferNew/quote-ids.ts
```diff
@@ -0,0 +1,44 @@
...
+    const qids = getQuoteIds(storage, history)
+    setQuoteIds(qids)
+  }, [history, storage, quoteIds])
+
+  return quoteIds
+}
+
+const getQuoteIds = (
+  storage: StorageState,
+  history: History,
```
styrelius, line 22: Instead of passing these in as arguments you could just import and use `useStorage` and `useHistory` directly in this hook - makes it more encapsulated IMO which is nice 😀

src/client/pages/OfferNew/quote-ids.ts
```diff
@@ -0,0 +1,44 @@
+import queryString from 'querystring'
+import { History } from 'history'
+import { useEffect, useState } from 'react'
+import { StorageState } from 'src/client/utils/StorageContainer'
+
+export const useQuoteIds = (storage: StorageState, history: History) => {
+  const [quoteIds, setQuoteIds] = useState<ReadonlyArray<string> | null>(null)
```
styrelius, line 7: Do we need to type this as `ReadonlyArray`? 🤔  I can't see why anyone would have the idea to mutate this array. But hey, maybe better safe than sorry? Or what was the reasoning here?

src/client/pages/OfferNew/index.tsx
```diff
@@ -44,24 +45,28 @@ export const OfferNew: React.FC = () => {
...
+    skip: quoteIds === null,
   })
 
-  const history = useHistory()
   const checkoutMatch = useRouteMatch(
     '/:locale(se-en|se|no-en|no|dk-en|dk)/new-member/sign',
   )
   const toggleCheckout = createToggleCheckout(history, currentLocale)
 
+  if (quoteIds === null) {
```
styrelius, line 67: Can `quoteIds` ever be `null`? Doesn't the `useQuoteIds` hook always return an array? Or am I missing something?

src/client/pages/OfferNew/quote-ids.ts
styrelius: And also, I know we use `useHistory` in the parent component but when you want the `location` I think it's recommended to use `useLocation` instead. https://reactrouter.com/web/api/history/history-is-mutable

styrelius (commented):
Nice work with the web stuff! ✨ 🕸️ 🙌  
Just a few things 👇

src/client/pages/OfferNew/quote-ids.ts
hedvigoscar: I'm working on the principle of least privilege for the return type, so that anyone accidentally attempting to mutate the array get type errors 🙂

src/client/pages/OfferNew/index.tsx
hedvigoscar: It's returning null for the initial state, which I used to distinguish between the states `NOT LOADED` and `LOADED, BUT NO DATA`. I guess it could be made more explicit, however 🤔

src/client/pages/OfferNew/quote-ids.ts
hedvigoscar: I was on the fence about this but decided to pass in the dependencies in this manner because I wanted a guarantee that I got the same objects in both tiers.

Come to think of it though, I'm not sure that's even needed, or desirable. I'm gonna refactor to your suggestion instead!

hedvigoscar:
Quick thought I had while working through this: I approached 'backwards-compatibility' with the rest of the codebase by populating the session storage with the quote ids, which would mean that other consumers would get what they previously were expecting.

We could almost as easily switch to using the hook in all cases where we want the quote ids, though. What do you think about that?

hedvigoscar:
I've fixed all the comments. Let me know if we should investigate using the hook in all call sites, or if this is sufficient.

src/client/pages/OfferNew/index.tsx
styrelius: Ah yes of course! 🙈 But yes, maybe it would be an idea to also return an `isLoading` state from the hook instead? That would be a way to be pretty explicit?

Also, realised that we already do a return of `<LoadingPage>` on line 81 so this condition (whether we leave it as `quoteIds === null`  or do an `isLoading`) should probably just be added to line 80. Personally I prefer to only have one `return` statement per component and do all the conditioning in the JSX but I know we do both ways in this repo so that change is probably unnecessary since we haven't yet agreed on one way or the other in the web team 😄

styrelius:
> We could almost as easily switch to using the hook in all cases where we want the quote ids, though. What do you think about that?

Hm, I think we only do this `storage.session.getSession()?.quoteIds` (i.e. using `useStorage` to get quote id's "saved" there) in two other places - in `SignLoading.tsx` and the `Debugger` `Offer.tsx`. But it would of course be nice to use the hook there as well! 🙌  In that case `useQuoteIds.ts` should be located in the `src/client/utils/hooks` folder though, I guess.

What do you think @gustaveen ?

src/client/pages/OfferNew/useQuoteIds.ts
```diff
@@ -0,0 +1,52 @@
...
+import { Location } from 'history'
+import { useEffect, useState } from 'react'
+import { useLocation } from 'react-router'
+import { StorageState, useStorage } from '../../utils/StorageContainer'
+
+/**
+ * Hook that allows general access to quote IDs.
+ * @returns Either an array of quote IDs, an empty array if none were found,
+ *          or `null` before quote IDs have been loaded.
+ */
```
styrelius, line 11: I'm not sure what I think about comments like these. 🤔 But I'm leaning towards that the code is clear enough so that it's not needed - it was just me being inattentive(sloppy? 😳 ) looking at the code before I think.

src/client/pages/SignLoading.tsx
```diff
@@ -17,11 +17,11 @@ import {
   useSignStatusQuery,
 } from 'data/graphql'
 import { getOfferData } from 'pages/OfferNew/utils'
-import { useStorage } from 'utils/StorageContainer'
 import { useTextKeys } from 'utils/textKeys'
 import { handleSignedEvent } from 'utils/tracking/signing'
 import { useTrack } from 'utils/tracking/tracking'
 import { useVariation, Variation } from 'utils/hooks/useVariation'
+import { useQuoteIds } from '../utils/hooks/useQuoteIds'
```
gustaveen, line 24:
```suggestion
import { useQuoteIds } from 'utils/hooks/useQuoteIds'
```

gustaveen (commented):
A bit late to the party, but looks good to me now, good job @hedvigoscar! 🙌 As mentioned might be nice to add a loading state to the hook, but that can be done later

styrelius (commented):
Very nice indeed! Still think an `isLoading` state would be neat to return from the `useQuoteIds` hook along with the `quoteIds` array (would be more declarative = 😀).

But my main issue now is that I don't manage to get this to work locally. If I clear the cookie storage (where we save the session) and then try to go to `http://localhost:8040/se/new-member/offer?quoteIds=[fb3333a0-5fdd-11eb-9a8c-190f836df6ed]` (passing an ID of an existing quote of course) I just get GraphQL error:

image

Am I doing something wrong?

gustaveen:
You need to add the token (url encoded) as well to be able to fetch the quote 👍 `http://localhost:8040/no/new-member/offer?quoteIds=[7cb7a572-8efc-4f0a-8949-c98968dc89bf]#token=04M4%2Bta406nEYw%3D%3D.f%2FNszSEJvQIyiQ%3D%3D.zYPrOyfqH4Si%2BA%3D%3D`

hedvigoscar:
I added the `isLoading`-functionality now 🙂

gustaveen (approved):
Nice @hedvigoscar 🙌 Works good for me!

styrelius (approved):
🙌

hedvigoscar:
When I've merged this branch, do I need to take any additional steps in deploying it to production? Or will you take it from here? 😅

styrelius:
I can deploy it - seems to be working fine in staging in the browser at least. Or do you want to do some more testing in the app @hedvigoscar ?

hedvigoscar:
No, that's fine, we can make changes down the line 🙂

PR #479, opened 2021-02-23
[GRW-205] [GRW-218] Fix / External insurance data error (finally!)

styrelius:
## What?

Add a new offer page component `StartDateLabelSwitcher.tsx` and move a bunch of things from `StartDate.tsx` there.
This component, which is then used in `StartDate`, will only be rendered if the quote property `currentInsurer` has a value (since this means the user has a current insurer that we know of). 
In `StartDateLabelSwitcher.tsx` `useExternalInsuranceDataQuery` is called, with the exception of the case where there is no `dataCollectionId` (the `skip` option on the query hook is used to achieve this).

This means we will only try to query for more data on the user's current insurance when we know that there really is data to fetch (i.e. the current insurance company is integrated with Insurely), instead of alway calling this hook when the user goes to `/offer` page.

In addition to this:
1. Create an `isLoadinPickedStartDate` state in `DateForm` component in order to render `LoadinDots` when the start date is set or removed and the quote is refetched.
2. Fix some styling issues with start date components.
3. add `TODO`'s in `ExternalInsuranceData.graphql` and `StartDateLabelSwitcher.tsx` since changes need to be made here when we implement Insurely in Norway.

## Why?

Because we want to get rid of the error that we now get when a user who doesn't have an external (meaning current) insurer goes to the `/offer` page, i.e. _every_ time someone goes through the regular ("needer") flow and not the switcher flow:

image

Jira issues: 
[GRW-218] 
[GRW-205]

## Screenshots and GIF
_I'm not able to test with a quote where the switcher flow succeeds but it looks and works the same as before in the failed switcher flow and the switcher flow where Insurely is not implemented at least_

**Switcher flow with current insurance that doesn't work with Insurely, meaning the `ExternalInsuranceDataQuery` is _not_ executed**

2021-02-22-offer-page-switcher-without-insurely

<br>

**Failed switcher flow but with `dataCollectionId`, which means the `ExternalInsuranceDataQuery` is executed as it should.**

2021-02-04-failde-switcher-offer

2021-03-03-start-date-failed-switcher-loading-dots

**Faked renewal date and current insurer**

2021-03-03-CTA-box-switcher

[GRW-205]: https://hedvig.atlassian.net/browse/GRW-205

[GRW-218]: https://hedvig.atlassian.net/browse/GRW-218


src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
```diff
@@ -313,10 +262,20 @@ const DateForm: React.FC<{
...
+                dataCollectionId={quote.dataCollectionId}
+              />
+            )}
+          {isLoadingPickedStartDate && (
+            <LoadingDotsWrapper>
+              <LoadingDots color={colorsV3.gray500} />
+            </LoadingDotsWrapper>
+          )}
+          {hasStartDate && !isLoadingPickedStartDate && getDateLabel()}
+          {!isLoadingPickedStartDate && <DownArrow />}
```
gustaveen, line 278: What about checking for `isLoadingPickedStartDate` so we don't have to check that in all the other statements? :) 
```suggestion
          {isLoadingPickedStartDate ? (
            <LoadingDotsWrapper>
              <LoadingDots color={colorsV3.gray500} />
            </LoadingDotsWrapper>
          ) : (
            <>
              {!hasStartDate && hasCurrentInsurer(quote) && (
                <StartDateLabelSwitcher
                  dataCollectionId={quote.dataCollectionId}
                />
              )}

              {hasStartDate && getDateLabel()}
              <DownArrow />
            </>
          )}
```

gustaveen (commented):
Good job! 😍  Nice addition with the loading dots as well! 👌  Just a few thoughts on the conditional rendering

gustaveen:
Also, I long for the day when we can test the switcher flow a bit easier 😅

src/client/pages/OfferNew/Introduction/Sidebar/StartDateLabelSwitcher.tsx
```diff
@@ -0,0 +1,109 @@
...
+          </DataCollectedStartDateDescription>
+        </DataCollectedStartDateWrapper>
+      )}
+      {!renewalDate && !loading && (
+        <GenericLabel>
+          {textKeys.SIDEBAR_STARTDATE_CELL_VALUE_SWITCHER()}
+        </GenericLabel>
+      )}
+    </>
+  )
```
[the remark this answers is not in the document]
styrelius: I'm not sure what I think about this. I kind of like having just one return statement in a component, then when looking at the file you can easily find what will be rendered. Though I can see a point with the JSX being less cluttered. But it's also nice in a way to see what are the exact conditions for these elements to be rendered. 
That's why I'm also not a huge fan of the `? :` syntax - it can get hard to understand what conditionals actually apply for specific elements. 🤔

src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
styrelius: See previous comment 👆

styrelius: But with all that said - this can definitely be written in a slightly more elegant way 😄

src/client/pages/OfferNew/Introduction/Sidebar/StartDateLabelSwitcher.tsx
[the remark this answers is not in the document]
styrelius: But with all that said - this can definitely be written in a slightly more elegant way 😄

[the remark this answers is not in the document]
gustaveen: Good points! Yes I think that ternaries can be a bit bad for readability for sure and also nesting conditional rendering or multiple conditional statements. In some places I agree it can be a bit hard to find if you do returns in different places if you have a messy component. But I do like the pattern of checking early and for one thing at the time, if you keep it together at the end of the component and it's clear what's being rendered like this for example 😃  :

```
  if (error) {
    return <ErrorMessage />
  }

  if (loading) {
    return <Empty />
  }

  return (
    ....
```

src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
```diff
@@ -313,10 +262,22 @@ const DateForm: React.FC<{
...
+            <>
+              {!hasStartDate && hasCurrentInsurer(quote) && (
+                <StartDateLabelSwitcher
+                  dataCollectionId={quote.dataCollectionId}
+                />
+              )}
+              {hasStartDate && getDateLabel()}
+              <DownArrow />
+            </>
+          )}
```
gustaveen, line 280: Sorry for being a pain over this 🙈  I think multiple conditions are easier to follow, like you had before, over nested if cases 😅

styrelius: Really? But this is the same principle as with a ternary operator - with the difference that the condition is clearly written out instead of just being "the thing that is not that other thing that we talked about som lines ago" (which it is with `:`/`else`).

src/client/pages/OfferNew/Introduction/Sidebar/StartDateLabelSwitcher.tsx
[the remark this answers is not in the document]
styrelius: Hm I see your point but it's a bit of a Slippery Slope! ☝️ Then suddenly people are returning 50 lines of JSX in the middle of a component!!!
But if we agree on that all return statements are at the bottom of the component (like in your example) and you get severely punished if you put them somewhere else then I guess it can be fine as a principle.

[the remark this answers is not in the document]
gustaveen: I hear you on that! 🙃  As long as we review the code we can prevent that 🎉  Or have a punishment 😅

src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
gustaveen: Maybe it's just med but yes, if choosing between those two options 🙃  But I'm fine with this 👍

gustaveen (approved):
🎉

styrelius:
Thank you!! 🙏

PR #481, opened 2021-02-25
Add name to ApolloClient

sampettersson:
## What?

`Adding name to ApolloClient`

## Why?

So that Giraffe knows which client caused the originating `Query/Mutation/Subscription`.


gustaveen (approved):
👍

sampettersson:
Thx 🕺

PR #482, opened 2021-03-02
Pass in locale to angelDataQuery to resolve translations

gustaveen:
## What?

Resolve danish translations for angel data

## Why?

Because we want translations :) 

[GRW-105]

[GRW-105]: https://hedvig.atlassian.net/browse/GRW-105

### DIFF
--- a/src/client/pages/Embark/index.tsx
+++ b/src/client/pages/Embark/index.tsx
@@ -190,8 +190,8 @@ interface EmbarkRootProps {
 }
 
 const ANGEL_DATA_QUERY = gql`
-  query AngelStory($name: String!) {
-    angelStory(name: $name) {
+  query AngelStory($name: String!, $locale: String!) {
+    angelStory(name: $name, locale: $locale) {
       content
     }
   }
@@ -205,6 +205,7 @@ interface AngelData {
 
 interface AngelVariables {
   name: string
+  locale: string
 }
 
 export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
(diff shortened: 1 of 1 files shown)


styrelius (approved):
Amazing! 🙌

src/client/pages/Embark/index.tsx
```diff
@@ -214,6 +215,7 @@ export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
     [key: string]: any
   }>()
   const currentLocale = useCurrentLocale()
+  const localeIsoCode = getPickedLocaleFromCurrentLocale(currentLocale)
```
palmenhq, line 218: Should we maybe rename this from `PickedLocale` -> `IsoLocale` or something? Picked locale doesn't seem to fit this use case while the actual function does, right?

palmenhq (approved):
🆒😄

src/client/pages/Embark/index.tsx
styrelius: Do you mean that we should write another function then? That does the same thing? Since this `getPickedLocaleFromCurrentLocale()` function is used in a bunch of other places it doesn't really make sense to rename it (although I don't really understand the meaning of "picked" here tbh).

palmenhq: So the picked locale was initially intended for mapping the url's locale (`/se`/`/no-en` etc) to the member's preferred locale (which they "pick" in the first screen of the onbording in the app, hence the name), and that's in iso format (`en_NO` or `sv_SE` for example) - but now when we require iso locales which i think  is not only used for the context of the picked locale, it would make sense to rename this mapper function

gustaveen: Ah actually, should I just use `getLocaleIsoCode()` instead of `getPickedLocaleFromCurrentLocale()`? I don't fully understand why we have two functions with the same input returning `Locale` :D

palmenhq: Ooh there's already such a function? 🤩 amaze, lets use it! haha maybe we can kill this function completely then

gustaveen: haha yes if I'm not missing the use case for it. will refactor that in a new PR 🙌

PR #483, opened 2021-03-03
Remove getPickedLocaleFromCurrentLocale

gustaveen:
## What?
Suggestion:
Remove `getPickedLocaleFromCurrentLocale ` function in favour of `getLocaleIsoCode`

## Why?

Since `getLocaleIsoCode` takes the same input and returns `Locale` it seems overkill to have two functions for this. But if can we do something to make it more clear what the function is used for and what it returns please come with suggestions. 

localeIsoCode is referring to the iso standard of locales like `sv, nb, da, en` but `Locale` also consist of the country code `SE, NO, DK`. It's clear-ish for me 😄

### DIFF
--- a/src/client/App.tsx
+++ b/src/client/App.tsx
@@ -2,10 +2,7 @@ import { Provider } from 'constate'
 import React from 'react'
 import { hot } from 'react-hot-loader'
 import { Route, Switch, useHistory } from 'react-router-dom'
-import {
-  getLocaleIsoCode,
-  useCurrentLocale,
-} from 'components/utils/CurrentLocale'
+import { getIsoLocale, useCurrentLocale } from 'components/utils/CurrentLocale'
 import { TextKeyProvider } from 'utils/textKeys'
 import { reactPageRoutes } from '../routes'
 import { AppTokenRetrieval } from './utils/AppTokenRetrieval'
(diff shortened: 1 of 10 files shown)


styrelius (dismissed):
This seems fine to me, since the value will be the same as what you get from `getPickedLocaleFromCurrentLocale`. But I'm always a bit nervous when we do things that seem to have to do with the apps' usage of the web-onboarding 😅 

Regarding the `getLocaleIsoCode` name I think it's good - what you get is the ISO code for language and two-letter ISO code for country. _Maybe_  we could call it `getIsoLocale` or `getLocaleIsoCodes` (with an "s" to be clear that it's two codes) but I think my view is that we don't have to change it.

styrelius:
Maybe it's actually the `Locale` type that has the most unclear naming but I mean we can't do anything about that since it comes from GraphQL 🤷‍♀️

palmenhq (dismissed):
Beautiful!

palmenhq:
Sorry missed your comment first @styrelius . I like `getIsoLocale`! I don't think we should name `Codes` in plural because that could make it a bit ambiguous if it would return for example multiple country codes rather, or?

gustaveen:
Thank you for input! 🤗

palmenhq (approved):
Good stuff!

PR #489, opened 2021-03-09
Feature / Add Danish routes

styrelius:
## What?

1. Add all routes needed for Denmark (including for the Embark flows that we haven't created yet):
  `/home`
  `/home-accident`
  `/home-accident-travel`

2. Make `Landing` component render for `/dk` and `/dk-en`, although it's pretty empty at the moment.

## Why?

Denmark 🇩🇰

_Referenced ticket(s): [GRW-229]_

## Screenshot

**Pretty sad (but existing!) landing page**
2021-03-09-DK-onboarding-landing

[GRW-229]: https://hedvig.atlassian.net/browse/GRW-229

### DIFF
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -159,19 +159,19 @@ export const reactPageRoutes: ReactPageRoute[] = [
           case 'dk':
           case 'dk-en':
             switch (match.params.name) {
-              case 'contents':
+              case 'home':
                 return {
-                  baseUrl: '/dk/new-member/contents',
+                  baseUrl: '/dk/new-member/home',
                   name: 'Web Onboarding DK - Contents',
                 }
-              case 'contents-accident':
+              case 'home-accident':
                 return {
-                  baseUrl: '/dk/new-member/contents-accident',
+                  baseUrl: '/dk/new-member/home-accident',
                   name: 'Web Onboarding DK - Danish Contents-Accident',
                 }
-              case 'contents-accident-travel':
+              case 'home-accident-travel':
                 return {
-                  baseUrl: '/dk/new-member/contents-accident-travel',
+                  baseUrl: '/dk/new-member/home-accident-travel',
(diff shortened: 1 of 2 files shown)


src/routes.tsx
```diff
@@ -229,6 +240,7 @@ export const reactPageRoutes: ReactPageRoute[] = [
                   baseUrl: '/se/new-member/switch',
                   name: 'Web Onboarding - Swedish Switcher',
                 }
+              // TODO: Remove this or? 👇
```
gustaveen, line 243: Hm, yeah I guess this isn't used :D

src/routes.tsx
```diff
@@ -154,17 +154,28 @@ export const reactPageRoutes: ReactPageRoute[] = [
...
+              case 'home-accident':
+                return {
+                  baseUrl: '/dk/new-member/home-accident',
+                  name: 'Web Onboarding DK - Danish Contents-Accident',
+                }
+              case 'home-accident-travel':
+                return {
+                  baseUrl: '/dk/new-member/home-accident-travel',
+                  name: 'Web Onboarding DK - Danish Contents-Accident-Travel',
+                }
```
gustaveen, line 176: No reason to block this PR since we had a discussion on this, I just wonder what we will write on the landing page (haven't seen translations for that yet). Home, Home contents or just Contents? So we align what we call it with the url and the first 2 options would work, it's rather if we would call it Contents like in Norway it will be weird.

gustaveen (approved):
Awesome! 🙌

src/routes.tsx
gustaveen: It's Home Contents 🎉

src/routes.tsx
styrelius: OK let's remove it later then since I really want to merge this now 😄

PR #491, opened 2021-03-16
Upgrade packages

palmenhq:
## What?

Upgrade packages. Also remove `watch` flag from wds as we're starting it with `serve` already and this causes a warning.

## Why?

Security n stuff

_Referenced ticket(s): []_


gustaveen (approved):
Nice @palmenhq 👌 🔒

PR #492, opened 2021-03-16
GRW-68 / Feat / Danish landing page

styrelius:
## What?

A Danish version of the landing page! 🙌 

Also, the way the `Landing` page is built is refactored:
- data structures for the data of different locales: `alternateLinksData` and `productsData` in `landingPageData.ts`.
- instead of repeating the same components many times in `Landing.tsx` we iterate over this data 😀 

## Why?

🇩🇰 

The reason for the refactoring is that the previous implementation don't scale so well - with the addition of the Danish things it would have gotten really hard to get an overview of the `Landing` component (in my view it should stil be split up into more files since `Landing.tsx` is more that 300 lines, but that can be done later on).

_Referenced ticket(s): [GRW-68]_

## Images

**/dk**
2021-03-17-landing-page-DK

**/dk**
2021-03-17-landing-page-DK

**/dk-en**
2021-03-17-landing-page-DK_EN

[GRW-68]: https://hedvig.atlassian.net/browse/GRW-68


src/routes.tsx
```diff
@@ -155,29 +155,30 @@ export const reactPageRoutes: ReactPageRoute[] = [
     render: ({ match }: RouteComponentProps<any>) => {
       const getProps = () => {
         match.params
```
gustaveen, line 157: `match.params` on this line can be removed now right? :)

src/client/pages/Landing/landingPageData.ts
```diff
@@ -0,0 +1,87 @@
...
+}[]
+
+export const alternateLinksData: AlternateLinksData = [
+  { hrefLang: 'sv-se', locale: 'se' },
+  { hrefLang: 'en-se', locale: 'se-en' },
+  { hrefLang: 'nb-no', locale: 'no' },
+  { hrefLang: 'en-no', locale: 'no-en' },
+  { hrefLang: 'da-dk', locale: 'dk' },
+  { hrefLang: 'en-dk', locale: 'dk-en' },
+]
```
gustaveen, line 17: This feels like something we can use in other places so maybe move it to a generic locales file or WDYT? No need to change this now however :)

styrelius: Yes my idea is that we should have this kind of structure more generally and yes this specific array is probably useful as it is in other places! But also I think we could do that separately, it might get messy in this branch otherwise 😄

src/routes.tsx
styrelius: Yes and probably it wasn't even needed before either - like it doesn't do anything? 🤔 😄 Anyway, well spotted, will remove

src/client/pages/Landing/landingPageData.ts
```diff
@@ -0,0 +1,87 @@
+import { Market } from 'components/utils/CurrentLocale'
+
+type Locale = 'se' | 'se-en' | 'no' | 'no-en' | 'dk' | 'dk-en'
```
gustaveen, line 3: Should we name this something other than `Locale` to not mix it up with the real Locale we get from the schema? Something like `LocalePath` to highlight what it's used for 🤷‍♂️

src/client/pages/Landing/landingPageData.ts
```diff
@@ -0,0 +1,87 @@
...
+  { hrefLang: 'en-se', locale: 'se-en' },
+  { hrefLang: 'nb-no', locale: 'no' },
+  { hrefLang: 'en-no', locale: 'no-en' },
+  { hrefLang: 'da-dk', locale: 'dk' },
+  { hrefLang: 'en-dk', locale: 'dk-en' },
+]
+
+type ProductData = {
+  id: string
+  linkSlug: string
```
gustaveen, line 21: 🔗  🐌

gustaveen: 👆 (Me laughing at my own jokes, how sad)

gustaveen (approved):
Great job @styrelius! 🛬 🇩🇰  🎉

src/client/pages/Landing/landingPageData.ts
styrelius: Hahah! 😍

src/client/pages/Landing/landingPageData.ts
styrelius: Hm, yes probably - good point! Since this is only used locally (hehe) in this file maybe change this later on when we will probably use this globally?

PR #493, opened 2021-03-17
Make sure we pass correct quoteIds for danish bundles

gustaveen:
## What?

Pass correct quoteIds for danish bundles. Not the smartest solution, but it works and enable us to have a test session :)

## Why?

Since we used named quoteIds when creating multiple quotes, we need to pass them down to `EmbarkProvider` them explicitly.


styrelius (approved):
Yeah it _really_ feels like this could be done in a more elegant way in the future 😄 Nicely found and fixed though! ✨

gustaveen:
Indeed! 😄  @styrelius

PR #495, opened 2021-03-24
GRW-79 / Fix / Details Modal household size

styrelius:
## What?

Fixed labels for `householdSize` / `coInsured` input fields in the `DetailsModal` to correspond with if the value is actually the household size (like in Sweden) or number of co-insured people (like in Norway and Denmark):
- Separate text keys depending on whether we're dealing with household size or co-insured
- More specifically the Danish translation is now re-formulated to mean _Number of co-insured_  instead of _Number of people (including myself)_

All placeholders are replaced with empty strings (some of them couldn't be removed so that's why I left empty strings) since all of them just had the same value as the labels which is not helpful, but rather adds noise and makes it harder to spot where something is missing. 

## Why?

For the Danish Details Modal the label previously says that the value is household size when it really is number of co-insured people. Initially the idea was to fix the actual value to be household size but the details modal implementation made it really tricky to do so quickly. Hence this solution instead.

_Referenced ticket(s): [GRW-79]_

## Screenshots

**/dk:**
2021-03-24-DK-co-insured-details-modal

**/se:**
2021-03-24-SE-household-size-details-modal

[GRW-79]: https://hedvig.atlassian.net/browse/GRW-79


gustaveen (approved):
Perfect! 🙌  We should maybe look over where we could add some more relevant placeholders, as you said, weird to just repeat the label 😄

PR #496, opened 2021-03-24
GRW-243 / Fix / Bundle discount badge

styrelius:
## What?

Be a bit more specific with the condition for rendering a `<Badge>` with the Norwegian bundle discount in the `Sidebar` component - it should not only be a bundle, but Norwegian as well.

## Why?

We will have bundles in Denmark as well, but no bundle discounts (the text key rendered here is specifically for Norway and shouldn't even exist in the Danish translation files so there shouldn't really be an issue here but still - it's good to be more precise 😇 )

_Referenced ticket(s): [GRW-243]_

[GRW-243]: https://hedvig.atlassian.net/browse/GRW-243


gustaveen (approved):
🥞 🙌

PR #497, opened 2021-03-25
Fix / Correct privacy policy links in Sign Disclaimer

styrelius:
## What?

New sign disclaimer "translations" with correct links to privacy policy for `/se-en` (plus `/dk` and `dk-en` but those pages don't exist yet).

## Why?

The links were incorrect.

~_Referenced ticket(s):_~


gustaveen (approved):
✍️

PR #498, opened 2021-04-06
Upgrade embark

sampettersson:
## What?

- Upgrading Embark to support new tracking filtering, remove `includeAllKeys` support.
- Insert quoteIds received in `ExternalRedirect` handler into session storage, this would down the line make the listeners on store quoteIds unnecessary as these would be defined directly in the embark flow instead (see [L375](https://github.com/HedvigInsurance/web-onboarding/blob/5998027cfe47ca925329d6e4f8908e6fd464233a/src/client/pages/Embark/index.tsx#L375)).

### Tracking

We were sending a bunch of unnecessary PII to Mixpanel

### QuoteIDS

We where reluctant to hard-code which store keys correspond to a quote id in the apps, so we came up with a more generic scalable solution to handle that.

_Referenced ticket(s): [APP-143]_

[APP-143]: https://hedvig.atlassian.net/browse/APP-143

### DIFF
--- a/package.json
+++ b/package.json
@@ -115,7 +115,7 @@
     "@emotion/core": "^10.0.22",
     "@emotion/styled": "^10.0.23",
     "@hedviginsurance/brand": "^4.0.2",
-    "@hedviginsurance/embark": "^2.0.11",
+    "@hedviginsurance/embark": "^2.0.15",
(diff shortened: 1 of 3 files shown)


gustaveen (approved):
🙌

PR #501, opened 2021-04-14
Feature / Checkout summary and terms links

styrelius:
## What?

Updated checkout summary section and sign disclaimer:

- Checkout summary has some new rows, and some of the old ones removed (see screenshots for the updated versions), and there is now only one section - with data from the main quote (being home contents quote if it's a bundle and just the quote if it's not - logic for this is added to `InsuranceSummary.tsx`). 
Since all information we need will exist on the main quote and not differ between quotes in a bundle there's no need to iterate over all quotes and display data from all of them.
Most importantly there is now a **Type of residence** row in all markets. The values for Denmark will probably have to be changed before release, but it's depending on what's decided regarding whether or not we care about if potential members live in apartments or houses. 

- There is a new section with **Terms and conditions** links, made in accordance with new design. 
These links corresponds to the links in the grey boxes on `/offer`, which means their titles come from the back-end, and as well as all url's except the first one (where `termType === 'TERMS_AND_CONDITIONS'`) - there we use the temporary hard coded one.

- All T&C links are removed from the `SignDisclaimer` and replaced with just one link to our privacy policy.
Either we get this url from the backend or we will just link to `/[currentLocale]/privacy`- this should probably be made in a smarter way in the future. 
The `CHECKOUT_SIGN_DISCLAIMER` text key now also has a `SIGN_BUTTON_LABEL` parameter, that will always be the same text key as the one used for the sign button. This way we can reference the sign button in the disclaimer text and know that the copy will always correspond with the copy on the button.

## Why?

Mange made some new design for this part of the checkout ages ago and since we needed to get the sign disclaimer and T&C links in place for the Danish `/sign` page now we agreed that it seemed like a good opportunity to also change the design and layout. [Here is the original new design](https://www.figma.com/file/eRR5ACHtbsPWDd4eUauWv1/Onboarding-Offer-page?node-id=102%3A930). I've made some tweaks but this has of course been checked with Mange.

With this design it's a lot easier to find the links to the T&C's etc, and it also means no more double disclaimer texts for bundle quotes 🥳 

Overall this part of the page should be super clear so members are 100% sure about what they're signing up for and that they actually submitted the correct information.

_Referenced ticket(s): [GRW-36] [GRW-81] [GRW-249]_

## Screenshots

image

image

image

image

image

image

image

image

image

image

image

image

[GRW-36]: https://hedvig.atlassian.net/browse/GRW-36
[GRW-81]: https://hedvig.atlassian.net/browse/GRW-81
[GRW-249]: https://hedvig.atlassian.net/browse/GRW-249


gustaveen:
Very good digging with this! @styrelius 🎉 Will be much a nicer summary to have this in place 🙌 
A initial question, should we tell BE to remove the Personal information link so we don't output it twice? Or add it to NO or DK, don't know what we prefer :)

styrelius:
> A initial question, should we tell BE to remove the Personal information link so we don't output it twice? Or add it to NO or DK, don't know what we prefer :)

Ah! Didn't think about it being outputted twice in Sweden. We should probably remove it from the T&C's section, since the privacy policy don't really fall under the umbrella of Terms & Conditions? Or maybe it does? 🤔  Then I don't think it matters that much that we have it in both places. 
But in the long run I think it makes most sense to get that from the back-end as well?

src/client/pages/OfferNew/Checkout/InsuranceSummary.tsx
```diff
@@ -1,256 +1,74 @@
...
   const textKeys = useTextKeys()
 
+  const mainQuoteInBundle = offerData.quotes.filter((quote) => {
+    return quoteDetailsHasAddress(quote.quoteDetails)
+  })
+
+  const mainQuote = isBundle(offerData)
+    ? mainQuoteInBundle[0]
+    : offerData.quotes[0]
+
```
gustaveen, line 58: I saw that we have a function getting the main quote `getMainOfferQuote`, so should we use that one? It also checks for contract type  (home content). I know we looked at this together that the first quote was home contents but it feels nice to check if it's the correct type if this would change for some reason :)

gustaveen (commented):
I agree, personally I don't think it should be under the T&C section 👍 But let's check that later 🙌

src/client/pages/OfferNew/Checkout/InsuranceSummary.tsx
styrelius: But that check is done when we get the `mainQuoteInBundle`. Or I check if quote has address, which means it's a home contents quote. But if there already is a function for this, sure I can use that.
image

styrelius: This solution is market agnostic, since it only checks for whether or not the quote has an address on it. The other solution is more specific and explicit but we need to add more logic for every market with these kinds of bundles. I'm not sure what's best, maybe the solution above makes too many assumptions?

styrelius:
> I agree, personally I don't think it should be under the T&C section 👍 But let's check that later 🙌

Hm, no but I can remove it from the Terms and Conditions section!

src/client/pages/OfferNew/Checkout/InsuranceSummaryTermsLinks.tsx
```diff
@@ -0,0 +1,74 @@
...
+    <Group>
+      {insuranceTermsArr.map(({ termType, data }) => {
+        const { displayName, url } = data
+        return (
+          <Row key={termType}>
+            <LinkWrapper>
+              <Link
+                href={getUrl({ termType, currentLocale, urlFromBackend: url })}
+                target="_blank"
+              >
```
gustaveen, line 64: For safety 🔒
```suggestion
                target="_blank"
                rel="noreferrer noopener"
              >
```

src/client/pages/OfferNew/Checkout/SignDisclaimer.tsx
```diff
@@ -23,61 +21,51 @@ const Wrapper = styled('div')`
...
+  const linkFromBackend = quote.insuranceTerms.get(
+    InsuranceTermType.PrivacyPolicy,
+  )?.url
+
+  if (linkFromBackend) {
+    return linkFromBackend
+  }
+
+  return `https://www.hedvig.com/${currentLocale}/privacy`
+}
```
gustaveen, line 40: As we talked about it would be best to get this from BE in the end, since the URL should be localised

src/client/pages/OfferNew/Checkout/InsuranceSummary.tsx
```diff
@@ -1,256 +1,74 @@
...
 
-const Wrapper = styled('div')`
-  padding: 0 0.5rem;
+const Wrapper = styled.div`
+  padding: 0 8px;
 `
-const Title = styled('h3')`
-  font-size: 1rem;
+const Title = styled.div`
+  font-size: 18px;
```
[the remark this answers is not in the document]
styrelius: Oh, I also think we should be consistent but it just doesn't make sense to me to use `rem` when you have to do like `1.125rem`. In my mind we should as soon as possible replace all `font-size` values (and a bunch of other things as well) with global values anyway. But maybe the switch will be easier if we continue to be consistent with `rem` 🤔 I'll change it.

src/client/pages/OfferNew/Checkout/SignDisclaimer.tsx
styrelius: Yeah I know this is not perfect - but it works as a temporary solution. And as long as we add redirects in Storyblok when url's are localised we're kind of OK? Maybe 😄

src/client/pages/OfferNew/Checkout/InsuranceSummary.tsx
gustaveen: I hear you, this something that will grow in complexity for the more markets we add. As you say it can be that checking for address is assuming too much. We don't need to pre optimize right now, but it would be nice if we checked for the same thing in all places by using the `mainQuouteInBundle`. And we can rewrite that to check for address if you feel it makes more sense 👍

gustaveen: Should we look this at a later stage in order to get it merged before tomorrow? @styrelius

styrelius: So sorry, didn't see your last comment until now. Have made the changes though 😀

styrelius:
### New headlines!
And a few other minor copy changes

🇳🇴 
Screenshot 2021-04-15 at 15 41 20
Screenshot 2021-04-15 at 15 41 32

🇩🇰 
Screenshot 2021-04-15 at 15 44 48
Screenshot 2021-04-15 at 15 43 36

🇸🇪 
Screenshot 2021-04-15 at 15 42 02

gustaveen (approved):
🙌 🛒

PR #505, opened 2021-04-23
Fix/set same site attribute

gustaveen:
## What
Set `Same Site` cookie attribute to `None`. Update cookie-storage package

## Why
This is due to an update from Chrome where the default value of the `SameSite` attribute switched from `None` to `Lax`.
This caused problems for NO members when connecting payments via Adyen. We could only recreate the problem when using 3D secure providers. When redirecting back to the onboarding,`hvsession` cookie was missing. 

### Review app
https://web-onboardi-fix-set-sa-o5k9oq.herokuapp.com/no/new-member

https://user-images.githubusercontent.com/6661511/115884495-42802c80-a44f-11eb-995a-7ae37876e568.mov

### DIFF
--- a/package.json
+++ b/package.json
@@ -128,7 +128,7 @@
     "axios": "^0.21.1",
     "color": "^3.1.2",
     "constate": "0.9.0",
-    "cookie-storage": "^3.2.0",
+    "cookie-storage": "^6.1.0",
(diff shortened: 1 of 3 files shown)


styrelius (approved):
🌟

PR #506, opened 2021-04-26
Fix/set secure cookie

gustaveen:
## What 
Attempt to fix the issue https://sentry.io/organizations/hedvig/issues/2358727634/?project=1308938 that came up after specifying `sameSite` attribute in this PR https://github.com/HedvigInsurance/web-onboarding/pull/505

The problem is that we cannot set secure cookies over unsecure connections. We have tried to reproduce the problem locally but only managed to reproduce it once so very hard to tell what causes the issue and to know if this fix actually does anything...

Adding the `app.proxy = true` sounded like it should fix the problem but hard to tell. Reading similar issues makes me think no :/ https://github.com/koajs/koa/issues/974 & https://github.com/webpack/webpack-dev-server/issues/933

Also, we should specify the same attribute for all cookies we set.

### DIFF
--- a/src/client/utils/storage/ServerCookieStorage.ts
+++ b/src/client/utils/storage/ServerCookieStorage.ts
@@ -23,6 +23,8 @@ export class ServerCookieStorage implements MinimalStorage {
       signed: false,
       expires: undefined,
       path: '/',
+      sameSite: 'None',
+      secure: true,
       httpOnly: false,
     })
   }
--- a/src/client/utils/storage/ServerCookieStorage.ts
+++ b/src/client/utils/storage/ServerCookieStorage.ts
@@ -23,9 +23,11 @@ export class ServerCookieStorage implements MinimalStorage {
       signed: false,
       expires: undefined,
       path: '/',
-      sameSite: 'None',
-      secure: true,
       httpOnly: false,
+      ...(process.env.NODE_ENV !== 'development' && {
+        sameSite: 'None',
+        secure: true,
+      }),
     })
   }
 
--- a/src/server/serverEntry.tsx
+++ b/src/server/serverEntry.tsx
@@ -41,6 +41,7 @@ const router = new Router()
 
 configureAssets(app)
 
+app.proxy = true
 appLogger.info(`Booting server on ${getPort()} 👢`)
 appLogger.info(
   `Sentry is ${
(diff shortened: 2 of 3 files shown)


src/client/utils/storage/ServerCookieStorage.ts
```diff
@@ -23,6 +23,8 @@ export class ServerCookieStorage implements MinimalStorage {
       signed: false,
       expires: undefined,
       path: '/',
+      sameSite: 'None',
+      secure: true,
```
palmenhq, line 27: shall we make this configurable?

gustaveen: You mean setting it as a .env config? Why do we need that? Setting it to secure in `localhost` works as far as I can see or am I missing something?

gustaveen: Ah, or did just you mean configurable for the function? :)

gustaveen:
So I found a way to reproduce the bug locally (I think). I'm having a hard time understanding how chrome has rolled out these changes. But you can make sure that your browser checks for this by going to [chrome://flags/](chrome://flags/) and setting SameSite options to enabled. Without the SameSite updates we did, it causes the 500 error when testing adyen with 3Dsecure payments. Adding them back makes it work so that's good at least.
Screenshot 2021-04-27 at 23 35 57

src/client/utils/storage/ServerCookieStorage.ts
palmenhq: no as an env var. It just seems weird to me to set this over an insecure connection.

styrelius: But if it works locally then maybe it doesn't matter, or? I just think maybe we shouldn't use env variables unless we need to? For complexity reasons that is 🤔

gustaveen: yeah this is a bit unclear to me. have read a bunch where it says it shouldn't work locally with secure cookies but it seems to do it :) I pushed a solution where we don't set it locally. but this will of course break 3d secure payments locally with the samesite flag turned on, but maybe that's ok 🙃

styrelius (dismissed):
Let's try this 🦺 🍪

palmenhq (dismissed):
amaze!

gustaveen:
Ok last update, let's try now! 😄

palmenhq (approved):

styrelius (approved):

PR #508, opened 2021-04-28
Chore / Improved Sentry logging of Underwriting limits hits

styrelius:
## What?

Update what arguments we pass to `useUnderwritingLimitsHitReporter` in `CheckoutContent`.

## Why?

We currently pass an array of the `description` property of the `limits` objects, but that one is deprecated and won't give any info. So it's replaced with to an array of the limits' `code`s.

~_Referenced ticket(s): []_~

## Screenshots

**new error**
image

**old/current error**
image


gustaveen (approved):
🙌

PR #510, opened 2021-04-30
Feat/update peril design

gustaveen:
## What?
Update Perils design

## Why?
We should _really_ have this in place before the test session. Copied from market-web and adjusted a bit. Long Peril names are squashing the icons as it is now, should also have the new brand design

**Review app**
https://web-onboardi-feat-updat-tf7i9t.herokuapp.com/se/new-member/debugger

### SE

https://user-images.githubusercontent.com/6661511/116655088-bf009700-a98a-11eb-800a-5867d469a859.mov

### NO

https://user-images.githubusercontent.com/6661511/116655115-c9229580-a98a-11eb-8735-269c3eb10a29.mov

### DK

https://user-images.githubusercontent.com/6661511/116655127-cde74980-a98a-11eb-810b-ebc97047ec26.mov

### Mobile
Screenshot 2021-04-30 at 08 03 04

### DIFF
--- a/src/client/pages/OfferNew/Perils/PerilCollection/PerilItem.tsx
+++ b/src/client/pages/OfferNew/Perils/PerilCollection/PerilItem.tsx
@@ -1,5 +1,5 @@
 import styled from '@emotion/styled'
-import { colorsV2, colorsV3, fonts } from '@hedviginsurance/brand'
+import { colorsV3 } from '@hedviginsurance/brand'
 import React from 'react'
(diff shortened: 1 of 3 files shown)


styrelius (approved):
Awesome! Looks a million times better! ✨ 💅  In my view a bit too much unnecessarily  complicated CSS and too many breakpoints that aren't really needed but it feels more important to get this out now than to have a discussion about that 😄

PR #512, opened 2021-04-30
GRW-281 Truncate overflowing peril titles

gustaveen:
## What?

Truncate overflowing peril titles

## Why?

Brought up a couple of times in the test session so adding this fix since we won't have time to update the peril modal before the first release.

**Before**
Screenshot 2021-04-30 at 16 57 47

**After**
Screenshot 2021-04-30 at 16 50 38


styrelius (approved):
✂️ 👍

gustaveen:
Bettet than before at least 😄

PR #513, opened 2021-05-03
Fix / Update setup instructions

robinandeer:
## What?

Add build step to initial setup in README.

## Why?

The initial build fails since it doesn't find the build/server.js file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,11 +38,16 @@ $ yarn
 ## Development
 
 ```sh
+# Compile the source files (only needed the first time)
+$ yarn build
+
 # Start the app
 $ yarn watch
 ```
 
 The app is running on `http://localhost:8080`. Visit [http://localhost:8080/se/new-member/](http://localhost:8080/se/new-member/) to view the first screen of the onboarding flow.
+
+> The port might differ on your machine. Check the logs for this message: `Booting server on 8040 👢`.
 
 ## Analytics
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -38,12 +38,17 @@ $ yarn
...
+$ yarn build
+
 # Start the app
 $ yarn watch
 ```
 
 The app is running on `http://localhost:8080`. Visit [http://localhost:8080/se/new-member/](http://localhost:8080/se/new-member/) to view the first screen of the onboarding flow.
 
+> The port might differ on your machine. Check the logs for this message: `Booting server on 8040 👢`.
+
```
gustaveen, line 51: Ah good catch! 🙌 The port was changed at some point and it hasn't been updated here, should be 8040 all the time 😄
```suggestion
The app is running on `http://localhost:8040`. Visit [http://localhost:8040/se/new-member/](http://localhost:8080/se/new-member/) to view the first screen of the onboarding flow.

```

robinandeer: Thanks - updated now and removed the callout section

README.md
```diff
@@ -54,7 +57,7 @@ Text keys live in [Lokalise](https://lokalise.com/) and are exported from there
 
 This is the process for updating/adding text keys:
 
-1. Make updates to text keys in Lokalise, i.e. add new text keys or update translations. Preferrably we use the Figma/Sketch
+1. Make updates to text keys in Lokalise, i.e. add new text keys or update translations. Preferably we use the Figma/Sketch
```
styrelius, line 60: We actually don't use the Figma plugin _at all_ hehe.. 😳  But would be nice to use that though, so maybe we can leave that sentence to force it to implement it at some point? 😅

styrelius (approved):
Nice! ✨

PR #515, opened 2021-05-04
Add support for Danish locale in AydenCheckout

robinandeer:
## What?

Add support for Danish locale in AydenCheckout component.
Make en-US default locale.

Screenshot 2021-05-04 at 11 12 56

(We haven't figured out how to localize "Credit Card" label...)

### DIFF
--- a/src/client/pages/ConnectPayment/components/AdyenCheckout.tsx
+++ b/src/client/pages/ConnectPayment/components/AdyenCheckout.tsx
@@ -199,9 +199,9 @@ const createAdyenCheckout = ({
 }: AdyenCheckoutProps) => {
   const locale = match([
     ['sv_SE', 'sv-SE'],
-    ['en_SE', 'en-US'],
     ['nb_NO', 'no-NO'],
-    ['en_NO', 'en-US'],
+    ['da_DK', 'da-DK'],
+    [match.any(), 'en-US'],
   ])(getIsoLocale(currentLocale))
 
   const returnUrl = `${window.location.origin}/${currentLocale}/new-member/connect-payment/adyen-callback`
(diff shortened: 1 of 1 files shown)


gustaveen (dismissed):
Woho! 💵 👌

src/client/pages/ConnectPayment/components/AdyenCheckout.tsx
```diff
@@ -199,9 +199,9 @@ const createAdyenCheckout = ({
 }: AdyenCheckoutProps) => {
   const locale = match([
     ['sv_SE', 'sv-SE'],
```
styrelius, line 201: Should we remove `['sv_SE', 'sv-SE']` since we don't use Adyen in Sweden?

styrelius (commented):
This has nothing to do with these changes but I just realised that if you change language with the language picker thingy in the top right corner the language of this component doesn't change, but only if the page is reloaded is the component re-rendered with the new language. Might be nice to fix that as well in this PR? 🤔 

2021-05-04-adyen

src/client/pages/ConnectPayment/components/AdyenCheckout.tsx
robinandeer: Hmm didn't think about that... perhaps removing it will avoid some confusion

styrelius:
> That content is in an iframe so makes it a little trickier but we could handle that. The other option would be to always reload the whole page whenever you switch language. It's not something people do more than once and should take care of most a lot of edge cases.

I guess a page reload could be fine? Or what do you think @gustaveen ? I mean yes, probably no-one even uses this language switch, feels like if you made it this far in one language there wouldn't really be a reason to switch.. Maybe we should just not do anything about this? Or remove the possibility to switch language on this page?

gustaveen:
Ah, sorry missed the conversation here! 
I see, yeah it's the same in NO as well I assume. I think a page reload is fine in this case 👍  We can add a ticket for this and adress this in a new PR for sure @robinandeer

gustaveen (approved):
🎉

PR #516, opened 2021-05-04
Add section about debugger UI

robinandeer:
## What?

Add information and a screenshot of debugger UI to README.

## Why?

It's hard to know it exists unless you are introduced to it.


README.md
```diff
@@ -44,6 +44,20 @@ $ yarn watch
...
+You access it on this URL: `http://localhost:8040/se/new-member/debugger`.
+
+![Debugger UI Preview](.github/assets/debugger-ui-preview.png)
+
+> You might experience problems due to existing state interfering with the session. Click the "Nuke all state 💣" button to perform a reset and try again.
+
+1. Fill in offer information.
+1. Click the "Create quote" button.
+1. You should see a bunch of GraphQL data and a button to take you to the "Offer page".
+
```
styrelius, line 60:
```suggestion
Or, if you have an id of an existing quote you'd like to use it's possible to replace the quote id generated for this session - the existing quote will be fetched as soon as you add the id to the input field.
```

README.md
```diff
@@ -44,6 +44,20 @@ $ yarn watch
 
 The app is running on `http://localhost:8080`. Visit [http://localhost:8080/se/new-member/](http://localhost:8080/se/new-member/) to view the first screen of the onboarding flow.
 
+### Debugger UI
+
+Manually stepping through the onboarding flow becomes tedious. Therefore, you can use the "debugger UI" that let's you bypass the initial steps and takes you straight to the "Offer page".
+
+You access it on this URL: `http://localhost:8040/se/new-member/debugger`.
```
styrelius, line 51:
```suggestion
You access it locally on this URL: `http://localhost:8040/se/new-member/debugger`.
It also works in staging: `https://www.dev.hedvigit.com/se/new-member/debugger`.
```

styrelius (commented):
Wow, awesome!! 🥳 🤸‍♀️ Of course we should have this in the README, but I hadn't even thought about it 🙈 

A few suggested additions 👇

README.md
```diff
@@ -44,6 +44,20 @@ $ yarn watch
...
+
+> You might experience problems due to existing state interfering with the session. Click the "Nuke all state 💣" button to perform a reset and try again.
+
+1. Fill in offer information.
+1. Click the "Create quote" button.
+1. You should see a bunch of GraphQL data and a button to take you to the "Offer page".
+
 ## Analytics
 
 We use Segment, Mixpanel and Google Tag Manager as our analytics tools. You can read about the setup in [Notion](https://www.notion.so/hedviginsurance/Mixpanel-Setup-iOS-Web-Embark-d1abeb9ba7634adea6155f847d32cd8d)
```
styrelius, line 63: Also just realised that maybe Google Analytics should be mentioned here? Not just Google Tag Manager? Or?

styrelius: @gustaveen 👆 ❔  And we haven't removed Segment yet right?

gustaveen: Correct! We still use Segment. We could explain more maybe but I'm not sure if we should have it here in the readme or WDYT? Because it's a bit har d to explain in a few sentences. Google Analytics is where Sonny makes dashboards and analyse the data but all the data that goes through GTM then goes to GA. But it doesn't hurt to mention for sure 👍

gustaveen (dismissed):
Thank you for doing this @robinandeer This is perfect! 🙏

README.md
styrelius: Hm, maybe leave it for now. It's not that important I guess. But that Notion page we link to seems like we should replace or update since it just talks about Mixpanel

gustaveen (dismissed):

README.md
robinandeer: Sooooo... are we also still use Mixpanel? Can only find a reference to it under Koa Helmet "contentSecurityPolicy > directives > defaultSrc".

Perhaps we can link to this notion page instead: https://www.notion.so/hedviginsurance/Data-Analytics-4eeb34a5b1e043f798ec94e71aec2533

But perhaps it's better to change that in a separate PR :)

gustaveen (dismissed):
I approve again if you don't want to amend something else :)

README.md
gustaveen: It's a hard to grasp 😆  Data goes to Segment -> Mixpanel

gustaveen (approved):

PR #517, opened 2021-05-04
Fix/update translations

gustaveen:
## What?

Update translations

## Why?

- In danish it's called CPR nummer, not Personnummer
- Don't write "Andelsbolig" for homeowner. This will be updated for other keys as well later (will sync with Isabella)
- Update translations for coming soon communication for travel bundle (danish translations are pending)


styrelius (approved):
👍

robinandeer (approved):

PR #518, opened 2021-05-04
Add danish app mockups on download page

gustaveen:
## Why?

All mockups were in english, add mockups with danish text

Screenshot 2021-05-04 at 17 16 08

Screenshot 2021-05-04 at 17 15 56


robinandeer (dismissed):
Looks good!

Do we run images through some minifier before they go live?
I just tried running this through tinypng and it manages to compress it 70% smaller than the original:

hedvig-connect-payment-dk-en-min

styrelius:
> Do we run images through some minifier before they go live?

@robinandeer  Not that I know of. Is that the best way to do it? Or should we provide a smaller version as well in a `srcset`? _Really_ not an expert so asking because I don't know anything about best practice here 😅

gustaveen:
Good point @robinandeer 🙌  We don't have any image processing in place. It has been a bit of a mixed workflow :D But I export JPEGs through photoshop or ask the one who sends images to compress them down to around 30% if that still looks good to reduce file size. But I have been bad at doing it for PNGs since you can't tweak it that much in Photoshop. But let's compress them and amend to the PR since they were rather big (305kb) 👌

styrelius:
@gustaveen @robinandeer  But isn't it still a good idea to provide a `srcset`?

robinandeer:
srcset is used to serve different resolutions to different devices if I remember correctly (or modern image formats like WebP with a fallback).

**On top of that**, you can also compress image assets. There are some really clever algorithms that can compress images without losing visual fidelity.

gustaveen:
I think it nice to use srcset for large images like heroes where you could serve a smaller image for mobile devices for example. In this case where the image isn't "fullscreen" I think it's sufficient to compress it and just serve one image since the size difference would be rather small I think :)

styrelius:
🆗 🆒 No maybe the difference isn't big enough between the size of this image on desktop v.s. mobile 👍

gustaveen:
Updated with compressed PNGs! 🤗

robinandeer (dismissed):
LGTM

robinandeer (approved):

PR #519, opened 2021-05-05
Remove home-accident-travel bundle route (DK)

robinandeer:
## What?

Remove home-accident-travel bundle route for DK.
Combine no + no-en embark routes.
Use optional chaining to access props.
Redirect to landing page for URLs that don't match existing offers.

_Referenced ticket(s): [GRW-294]_

## Why?

Will not go live with that offer.

[GRW-294]: https://hedvig.atlassian.net/browse/GRW-294

### DIFF
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -15,6 +15,23 @@ import { LoginApp } from './client/pages/LoginApp'
 import { OfferNew } from './client/pages/OfferNew'
 import { SignLoading } from './client/pages/SignLoading'
 
+enum EmbarkStory {
+  DenmarkContents = 'Web Onboarding DK - Contents',
+  DenmarkContentsAccident = 'Web Onboarding DK - Danish Contents-Accident',
+
+  NorwayContentsNorwegian = 'Web Onboarding NO - Norwegian Contents',
+  NorwayContentsEnglish = 'Web Onboarding NO - English Contents',
+  NorwayTravelNorwegian = 'Web Onboarding NO - Norwegian Travel',
+  NorwayTravelEnglish = 'Web Onboarding NO - English Travel',
+  NorwayComboNorwegian = 'Web Onboarding NO - Norwegian Combo',
+  NorwayComboEnglish = 'Web Onboarding NO - English Combo',
+
+  SwedenNeederSwedish = 'Web Onboarding - Swedish Needer',
+  SwedenNeederEnglish = 'Web Onboarding - English Needer',
+  SwedenSwitcherSwedish = 'Web Onboarding - Swedish Switcher',
+  SwedenSwitcherEnglish = 'Web Onboarding - English Switcher',
+}
+
 export interface ServerSideRoute {
   path: string | RegExp
   titleTextKey: string
@@ -162,12 +179,12 @@ export const reactPageRoutes: ReactPageRoute[] = [
               case 'home':
                 return {
                   baseUrl: `/${locale}/new-member/home`,
-                  name: 'Web Onboarding DK - Contents',
+                  name: EmbarkStory.DenmarkContents,
                 }
               case 'home-accident':
                 return {
                   baseUrl: `/${locale}/new-member/home-accident`,
-                  name: 'Web Onboarding DK - Danish Contents-Accident',
+                  name: EmbarkStory.DenmarkContentsAccident,
                 }
             }
             break
@@ -177,23 +194,26 @@ export const reactPageRoutes: ReactPageRoute[] = [
               case 'contents':
                 return {
                   baseUrl: `/${locale}/new-member/contents`,
-                  name: `Web Onboarding NO - ${
-                    locale === 'no' ? 'Norwegian Contents' : 'English Contents'
-                  }`,
+                  name:
+                    locale === 'no'
+                      ? EmbarkStory.NorwayContentsNorwegian
+                      : EmbarkStory.NorwayContentsEnglish,
                 }
               case 'travel':
                 return {
                   baseUrl: `/${locale}/new-member/travel`,
-                  name: `Web Onboarding NO - ${
-                    locale === 'no' ? 'Norwegian Travel' : 'English Travel'
-                  }`,
+                  name:
+                    locale === 'no'
+                      ? EmbarkStory.NorwayTravelNorwegian
+                      : EmbarkStory.NorwayTravelEnglish,
                 }
               case 'combo':
                 return {
                   baseUrl: `/${locale}/new-member/combo`,
-                  name: `Web Onboarding NO - ${
-                    locale === 'no' ? 'Norwegian Combo' : 'English Combo'
-                  }`,
+                  name:
+                    locale === 'no'
+                      ? EmbarkStory.NorwayComboNorwegian
+                      : EmbarkStory.NorwayComboEnglish,
                 }
             }
             break
@@ -203,16 +223,18 @@ export const reactPageRoutes: ReactPageRoute[] = [
               case 'new':
                 return {
                   baseUrl: `/${locale}/new-member/new`,
-                  name: `Web Onboarding - ${
-                    locale === 'se' ? 'Swedish Needer' : 'English Needer'
-                  }`,
+                  name:
+                    locale === 'no'
+                      ? EmbarkStory.SwedenNeederSwedish
+                      : EmbarkStory.SwedenNeederEnglish,
                 }
               case 'switch':
                 return {
                   baseUrl: `/${locale}/new-member/switch`,
-                  name: `Web Onboarding - ${
-                    locale === 'se' ? 'Swedish Switcher' : 'English Switcher'
-                  }`,
+                  name:
+                    locale === 'no'
+                      ? EmbarkStory.SwedenSwitcherSwedish
+                      : EmbarkStory.SwedenSwitcherEnglish,
                 }
             }
             break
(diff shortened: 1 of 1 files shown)


robinandeer:
@gustafguner / @styrelius - What should happen if a user lands on this route: `/dk/new-member/home-accident-travel` (or any route that doesn't match one of the supported bundles?

Right now you see the landing page with the different offers. But shouldn't you get to a 404-page?

I can come up with a solution for it but the `FourOhFour` route component seems to use the old branding so wasn't sure if we can use that 😅

styrelius:
@robinandeer (Updated your comment to use the right Gustav 😄) 
I think it's OK that you just get redirected to the landing page. I mean that should be where you want to go anyway, in order to pick a legit option? But maybe it's confusing and it's better to have a 404-page with a link to the landing page? 

Anyway, it's correct that the `FourOhFour` component uses old branding so no, that shouldn't be used. And it seems like maybe not the right focus at the moment to build a new 404-page?

robinandeer:
@styrelius thx 😅 

Yeah, I wouldn't say anything is broken at the moment. Users won't be able to click that offer anyway so it's not clear how they would end up on any offer that doesn't exist.

Perhaps I leave a comment in the code instead to remember to update it later?

styrelius (dismissed):
C.L.E.A.N 👌

styrelius:
@robinandeer Maybe add a ticket in Jira? I think that might be better than a comment actually

src/routes.tsx
```diff
@@ -169,84 +169,53 @@ export const reactPageRoutes: ReactPageRoute[] = [
...
           case 'no-en':
             switch (name) {
               case 'contents':
                 return {
-                  baseUrl: '/no-en/new-member/contents',
-                  name: 'Web Onboarding NO - English Contents',
+                  baseUrl: `/${locale}/new-member/contents`,
+                  name: `Web Onboarding NO - ${
+                    locale === 'no' ? 'Norwegian Contents' : 'English Contents'
+                  }`,
```
robinandeer, line 182: How about this thing? Do we want to keep the "English/Norwegian" as part of the name property. I saw that we did it slightly differently in DK.

What is it used for?

styrelius: Yes unfortunately we need to keep that! 😔  The reason is that when the Danish flows were created we had added an integration with Lokalise in order to handle different languages. Hence only one Embark/Angel flow per insurance product is needed.
But the older flows, the Norwegian and Swedish ones, don't have that. So there we have two flows per product - one in English and one in the local language. 
I hope that's understandable 😅 It's a bit complicated hehe..

styrelius: The `name` here is what we use as input to fetch the correct flow in `client/pages/Embark/index.tsx`
image

gustaveen: Ah of course, that was me that mislead you @robinandeer since I was like "it's much cleaner for the DK routes, you can probably go ahead and refactor that" 😂

src/routes.tsx
```diff
@@ -257,8 +226,8 @@ export const reactPageRoutes: ReactPageRoute[] = [
       return (
         <EmbarkRoot
           language={match.params.locale}
-          name={(props && props.name) || undefined}
-          baseUrl={(props && props.baseUrl) || undefined}
+          name={props?.name}
+          baseUrl={props?.baseUrl}
```
gustaveen, line 230: 👌

gustaveen (dismissed):
Good job! 🙏
Agreed! I think it's cool to send the user to the startpage in these cases 🙌

src/routes.tsx
robinandeer: haha okey thanks @styrelius! I made a quick update to bundle them in an enum to make it more obvious :)

src/routes.tsx
```diff
@@ -162,104 +191,81 @@ export const reactPageRoutes: ReactPageRoute[] = [
...
+          case 'se':
           case 'se-en':
             switch (name) {
               case 'new':
                 return {
-                  baseUrl: '/se-en/new-member/new',
-                  name: 'Web Onboarding - English Needer',
+                  baseUrl: `/${locale}/new-member/new`,
+                  name:
+                    locale === 'no'
```
styrelius, line 239:
```suggestion
                    locale === 'se'
```

styrelius: 🙈

src/routes.tsx
```diff
@@ -162,104 +191,81 @@ export const reactPageRoutes: ReactPageRoute[] = [
...
+                      ? EmbarkStory.SwedenNeederSwedish
+                      : EmbarkStory.SwedenNeederEnglish,
                 }
               case 'switch':
                 return {
-                  baseUrl: '/se-en/new-member/switch',
-                  name: 'Web Onboarding - English Switcher',
+                  baseUrl: `/${locale}/new-member/switch`,
+                  name:
+                    locale === 'no'
```
styrelius, line 247:
```suggestion
                    locale === 'se'
```

styrelius (changes requested):
Nice! 🌸 Seems like a few copy-paste mistakes though 👇

src/routes.tsx
```diff
@@ -151,7 +168,19 @@ export const reactPageRoutes: ReactPageRoute[] = [
...
   {
-    path: LOCALE_PATH_PATTERN + '/new-member/:name?/:id?',
+    path: LOCALE_PATH_PATTERN + '/new-member',
+    render: ({ match }: RouteComponentProps<any>) => (
+      <EmbarkRoot
+        language={match.params.locale}
+        name={undefined}
+        baseUrl={undefined}
+        showLanding={true}
+      />
```
styrelius, line 178: Maybe we could just render `<Landing language={match.params.locale} />` here instead? Then we could actually remove the `showLanding` prop and everything related to that from `EmbarkRoot`. Would make it easier to understand what's going on IMO. But I might be missing something @gustaveen ?

It works fine locally for me, the only thing is of course that with the `<Redirect to={`/${match.params.locale}/new-member`} />` we get like a new page load (with white screen) instead of a smooth transition to the landing page. But I think it's worth it, seeing that the routing here gets more declarative.

robinandeer: Hmm, wouldn't expect it to behave like that. `<Redirect />` should only perform a client side navigation, no?

I agree that this could be taken a step further but didn't want to change too many things before you could have a look!

I'm also realizing I don't need to pass `undefined` to the props - could just leave them out.

robinandeer: Did some investigation. There are a few changes that need to be made to make it work. However, I think separating `Landing` from `EmbarkRoot` would simplify things. Would recommend doing it as part of a separate PR though.

styrelius: Yeah, I was confused by the white screen. But `<Redirect />` does replace the current location in the history stack though - but that should still not be the "real" history stack if you get what I mean 🙃 Or? 
But anyways, I don't think it matters since it's just if you mess with the url after already being in a flow and I don't know why anyone would do that. Truly an edge case

robinandeer: Okey, cool I think I understand! I thought about adding the "push" prop to add the redirect to history but seemed better just to replace (since that endpoint doesn't exist).

Could it be because of the fade animation (opacity 0 -> 100)?

I made a draft of the changes needed to extract Landing from EmbarkRoot here btw: https://github.com/HedvigInsurance/web-onboarding/pull/523/files?diff=split&w=1

styrelius: No I agree, it makes more sense to just replace the location 👌 

Also, nice draft PR - I really think we should do it that way! 🙌  It's so confusing that the landing page, which really has nothing to do with Embark, is baked into the `EmbarkRoot` component

styrelius: And yeah, maybe it's because of the animation 🧠 Then I guess we could just add the same black background to the `Landing` component so it's black before the stuff fades in? Then it would look the same as when it's rendered in `EmbarkRoot`

src/routes.tsx
```diff
@@ -15,6 +15,23 @@ import { LoginApp } from './client/pages/LoginApp'
...
+  NorwayTravelNorwegian = 'Web Onboarding NO - Norwegian Travel',
+  NorwayTravelEnglish = 'Web Onboarding NO - English Travel',
+  NorwayComboNorwegian = 'Web Onboarding NO - Norwegian Combo',
+  NorwayComboEnglish = 'Web Onboarding NO - English Combo',
+
+  SwedenNeederSwedish = 'Web Onboarding - Swedish Needer',
+  SwedenNeederEnglish = 'Web Onboarding - English Needer',
+  SwedenSwitcherSwedish = 'Web Onboarding - Swedish Switcher',
+  SwedenSwitcherEnglish = 'Web Onboarding - English Switcher',
+}
```
styrelius, line 33: Really nice to have these collected like this 👆 Just a question out of curiosity - what is the value here of using an enum instead of a plain ol' JS object?

styrelius (approved):
🙌

src/routes.tsx
robinandeer: Yeah that would be a smart thing to add indeed! Will look into it

src/routes.tsx
robinandeer: Hmm, no we are not using it to type anything so a JS object would work just as well in this case I believe

gustaveen (approved):
Really nice Robin! 🙌  Makes it much clearer with the enum that it's not a just a name but actually maps to an Embarkstory 😄

src/routes.tsx
robinandeer: Body is already black by default so then I don't think we can do much about the "white flash". Perhaps it only happens in local dev though...

PR #520, opened 2021-05-05
Add coming soon communication to Travel bundle 🚧

gustaveen:
## What?

Disable Home, Accident & Travel bundle on startpage. Added `disable` prop to Card component to change colors etc and not render a `Link` component

## Why?
Not going live with that bundle :) 

## Demo

https://user-images.githubusercontent.com/6661511/117133978-455f1380-ada5-11eb-8d97-947c59c20aa8.mov


styrelius (approved):
Nice! 👌 🌸 🗺️

robinandeer (approved):
🎉

PR #521, opened 2021-05-05
Feat / Birthdate and SSN format

styrelius:
## What?

- Process of centralizing locale specific data is started, by creating `locales.ts` in new folder `/l10n` directly in `/client` folder.
  This file contains a data structure with all the locales and the data specific to that locale. This includes SSN and birthdate formats (with both example strings and regex's used for validating strings)
- Function `getFormattedBirthdate` in `/OfferNew/utils.ts` which change the format of a birthdate to fit the localized variant, plus a few tests for that function.
- Birthdate in checkout summary is formatted with above mentioned function.
- Validation of SSN in `LoginForm.tsx` (the one only used temporarily in the apps to login) and in `UserDetailsForm.tsx` (part of the checkout) is replaced with using regex etc from `locales.ts`.
- Placeholder is added to birthdate input in the Edit Details modal. Formatting the value back and forth turned out to be pretty complicated though, partly due to the way the code is structured in those files.

## Why?

- We've been getting feedback that the format of birthdate in checkout is incorrect in Denmark (and Norway)
- If you delete your birthdate in the Edit Details modal there is no hint on what format you have to use
- All of the logic handling SSN formats and birthdate formats was spread across components, with market checks etc, and it's probably beneficial to have this in one place. The idea is to have the current locale data object to just be saved in a context/store so no component will have to do any of those checks and if/switch statements. 

_Referenced ticket(s): [GRW-250] [GRW-282]_

## Screenshots

**checkout summary with local birthdate format /se:**

2021-05-05-checkout-summary-se

<br />

**checkout summary with local birthdate format /no:**

2021-05-05-checkout-summary-no

<br />

**checkout summary with local birthdate format /dk-en:**

2021-05-05-checkout-summary-dk

<br />

**edit details modal - "Personal information" section with birthdate placeholder:**

2021-05-05-details-modal-dk

2021-05-05-details-modal-dk-en

👆 _This looks the same in Norway, in Sweden the input field for editing birthdate is missing (should be added though)_

[GRW-250]: https://hedvig.atlassian.net/browse/GRW-250
[GRW-282]: https://hedvig.atlassian.net/browse/GRW-282


src/client/l10n/locales.ts
```diff
@@ -0,0 +1,147 @@
...
+const yearMonthDayFormat = /^(19[0-9]{2}|20[0-9]{2})-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])$/
+const dayMonthYearFormat = /^(0[1-9]|[1-2][0-9]|3[0-1])-(0[1-9]|1[0-2])-(19[0-9]{2}|20[0-9]{2})$/
+
+export const birthdateFormats: Record<MarketLabel | 'default', RegExp> = {
+  default: yearMonthDayFormat,
+  SE: yearMonthDayFormat,
+  NO: dayMonthYearFormat,
+  DK: dayMonthYearFormat,
+}
+
```
styrelius, line 49: Maybe all this SSN and birthdate stuff should be in its own file in the same folder? Or what do you think?

src/client/l10n/locales.ts
```diff
@@ -0,0 +1,147 @@
...
+const yearMonthDayFormat = /^(19[0-9]{2}|20[0-9]{2})-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])$/
+const dayMonthYearFormat = /^(0[1-9]|[1-2][0-9]|3[0-1])-(0[1-9]|1[0-2])-(19[0-9]{2}|20[0-9]{2})$/
+
+export const birthdateFormats: Record<MarketLabel | 'default', RegExp> = {
+  default: yearMonthDayFormat,
+  SE: yearMonthDayFormat,
+  NO: dayMonthYearFormat,
+  DK: dayMonthYearFormat,
+}
```
styrelius, line 48: 👆  Maybe all this SSN and birthdate stuff should be in its own file in the same folder? Or what do you think?

src/client/pages/LoginApp/components/LoginForm.tsx
```diff
@@ -78,8 +75,12 @@ const SubmitButton = styled(Button)<{ disabled?: boolean }>`
 export const LoginForm: React.FC<LoginFormProps> = ({ onSubmit }) => {
   const textKeys = useTextKeys()
   const market = useMarket()
-  const ssnMaxLength = ssnLengthByMarket[market]
-  const ssnRegExp = ssnRegExpByMarket[market]
+
+  const currentLocale = useCurrentLocale()
```
robinandeer, line 79: Should the return value of this hook to updated to `LocaleLabel`? Or Another hook that directly returns "currentLocaleData"?

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
```diff
@@ -208,10 +214,17 @@ const getHouseholdSizeValue = (householdSize: number, textKeys: TextKeyMap) => {
...
 }
 
-const getPersonalDetails = (
-  person: OfferPersonInfo,
-  textKeys: TextKeyMap,
-): DetailsGroup => {
+type GetPersonalDetailsParams = {
+  person: OfferPersonInfo
+  textKeys: TextKeyMap
+  currentLocale: string
```
robinandeer, line 220: Should this be typed as `LocaleLabel`?

src/client/pages/OfferNew/utils.ts
```diff
@@ -347,37 +347,33 @@ export const getInsuranceTitle = (
...
-      return ssnRegExpByMarket[Market.Se].test(ssn)
+  if (hasCorrectFormat) {
+    return birthdate
   }
+
+  const defaultFormat = birthdateFormats.default // This is the format we expect from back-end
+  const hasExpectedFormat = defaultFormat.test(birthdate)
+
+  if (!hasExpectedFormat) {
+    throw `Format of birthdate "${birthdate}" doesn't match the expected default format YYYY-MM-DD`
```
robinandeer, line 372: Are we strict about throwing only `Error`? If we just throw `string` we miss out on nice debugging features (or can Sentry still handle that?)

robinandeer (dismissed):
Great with the centralization of l10n stuff!

Just left some comments about possibly being more strict with typing to avoid downcasting.

src/client/l10n/locales.ts
```diff
@@ -0,0 +1,147 @@
...
+type Ssn = {
+  length: number
+  formatExample: string
+  formatRegex: RegExp
+}
+
+type Birthdate = {
+  formatExample: string
+  formatRegex: RegExp
+  defaultFormatExample: string
```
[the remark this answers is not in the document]
styrelius: The `defaultFormatExample` is an example string of the format we get from and send to back-end. The `formatExample` is the localized one. If you have a suggestion of better naming that would be awesome! 😀

src/client/pages/LoginApp/components/LoginForm.tsx
styrelius: I think it probably should. But then my idea is to get rid of that hook altogether, using a context to store the current locale with all the locale specific data. So that's why I left it like this. It's kind of an in-between state, but I didn't want to do the whole big switch here, seeing that we have a lot of other things to do etc. 😅

src/client/pages/OfferNew/utils.ts
styrelius: Ah, cool! Didn't know the difference 😳 Not sure if Sentry catches this, will investigate 🕵️‍♀️

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
styrelius: Probably, as I wrote in another comment this is kind of in-between ways to do this. I'm unsure if it's worth is messing too much with this in this PR 🤔 But I _really_ appreciate you commenting on this sub-optimal typing situation! I completely agree that it _shouldn't_ look like this

src/client/l10n/locales.ts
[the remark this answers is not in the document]
gustaveen: Yeah that got me a bit confused as well what the default format was. WDYT about just naming it something like that, `backendDefaultFormat`

src/client/l10n/locales.ts
gustaveen: I think it would be nice to separate the locale data and and ssn/birthdate data in the long run. In my head locales is the data that you added:
```
    path: 'se',
    isoLocale: IsoLocale.SvSe,
    marketLabel: 'SE',
    htmlLang: 'sv',
```
So it might be a nicer separation to keep the ssn/birthdate data on it's own.

src/client/pages/LoginApp/components/LoginForm.tsx
gustaveen: I would vote for a separate hook in that case :)

src/client/pages/OfferNew/utils.ts
robinandeer: Sure, I believe it's especially stack traces that are missing unless you `throw new Error(message)`

> The fundamental benefit of Error objects is that they automatically keep track of where they were built and originated.

🙂

src/client/l10n/locales.ts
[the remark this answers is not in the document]
styrelius: Yes, that's much better, thank you! 🙏

src/client/pages/OfferNew/utils.ts
styrelius: Aha, nice! Will change that! 🙏

src/client/l10n/locales.ts
gustaveen: Just to clarify, this is not a blocker for this PR, we can address this later 🙌

robinandeer (approved):

PR #523, opened 2021-05-06
Extract Landing page from EmbarkRoot

robinandeer:
## What?

Extract Landing page from EmbarkRoot
- Move entry animation directly to Landing page component.
- Remove "showLanding" prop from EmbarkRoot.
- Show Landing for path /new-member.

> The diff looks more complicated than it is. Here you can view without whitespace changes: https://github.com/HedvigInsurance/web-onboarding/pull/523/files?diff=split&w=1

## Why?

Seems cleaner to separate these pages base on the route. EmbarkRoot doesn't need to be concerned about rendering Landing page.


src/routes.tsx
```diff
@@ -170,12 +171,7 @@ export const reactPageRoutes: ReactPageRoute[] = [
   {
     path: LOCALE_PATH_PATTERN + '/new-member',
     render: ({ match }: RouteComponentProps<any>) => (
-      <EmbarkRoot
-        language={match.params.locale}
-        name={undefined}
-        baseUrl={undefined}
-        showLanding={true}
-      />
+      <Landing language={match.params.locale} />
```
styrelius, line 174: We could also remove the `language` prop here and replace it with using `useCurrentLocale` inside the `Landing` component where `language` is used now (which is only one place) -  might make this even more clean 😀

styrelius (commented):
Super nice 👌 ✨ 

One suggestion that might improve this even further 👇

src/routes.tsx
robinandeer: Aha, `language` and `currentLocale` are the same things!

styrelius:
@robinandeer Maybe you've already investigated this but the reason bor the failed build is that type checking fails since we also need to remove the `language` prop from the `Landing` component in `Landing.stories.tsx` 😄

robinandeer:
> @robinandeer Maybe you've already investigated this but the reason bor the failed build is that type checking fails since we also need to remove the `language` prop from the `Landing` component in `Landing.stories.tsx` 😊

Thanks -- good reminder to run test/lint locally 🙃

styrelius (approved):
This is great! 🤩  Makes so much more sense that what we had previously. Checked locally that the session token still was created and stored as it should and worked fine to go through the flow, get an offer and sign.

It also seems to work with the apps - I could sign after going through the flow with query param `variation=IOS`, and the signed member could be found in Hope (had some issues with this last time I checked but I probably did something wrong then since it's working totally fine now)

gustaveen:
> This is great! 🤩 Makes so much more sense that what we had previously. Checked locally that the session token still was created and stored as it should and worked fine to go through the flow, get an offer and sign.
> 
> It also seems to work with the apps - I could sign after going through the flow with query param `variation=IOS`, and the signed member could be found in Hope (had some issues with this last time I checked but I probably did something wrong then since it's working totally fine now)

Awesome, then I'm all for it! 🎉

gustaveen (approved):

robinandeer:
Woho! I'll leave this open until we go live with the current staging build to have that to roll back to JIC

PR #524, opened 2021-05-06
Reload page on language change

robinandeer:
## What?

Use regular link in LanguagePicker to make sure page is reloaded.

## Why?

Takes care of edge cases like reloading iframes with correct locale.

### DIFF
--- a/src/client/pages/Embark/LanguagePicker.test.tsx
+++ b/src/client/pages/Embark/LanguagePicker.test.tsx
@@ -15,10 +15,10 @@ describe('Swedish language picker renders without 💥', () => {
   )
 
   it('has the correct links', () => {
-    expect(wrapper.find('Link[to*="/se/new-member"]')).toHaveLength(1)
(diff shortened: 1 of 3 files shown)


gustaveen:
Nice! That solves the Adyen problem 🙏 
The drawback is that it feels a bit sluggish if you change language with the long reload and white screen in between or WDYT? :)

### With reload

https://user-images.githubusercontent.com/6661511/117445668-40cc6380-af3b-11eb-890a-0db014377424.mov


### Without reload


https://user-images.githubusercontent.com/6661511/117445836-7e30f100-af3b-11eb-9da3-b6445674e285.mov

styrelius:
Hm, maybe we should just remove the language picker from that page? When I think about it I don't really see why anyone would use it - if you've made it this far in one language it seems improbable that you would feel the need to switch language here?

robinandeer:
We will continue by removing possibility from changing language on connect payment page.

PR #525, opened 2021-05-06
Add support for danish in Forever language picker

robinandeer:
## What?

Centralize language options by market.
Dynamically render language options on "Forever" page.

## Why?

This change was missed likely because we have two isolated language pickers. So I thought it made sense to centralize the language options so when we add the next language we only need to do it in one place.

Perhaps these changes should be moved to the `locales.ts` module @styrelius is working on.

### DIFF
--- a/src/client/pages/Forever/components/LanguagePicker.tsx
+++ b/src/client/pages/Forever/components/LanguagePicker.tsx
@@ -38,19 +38,14 @@ export const LanguagePicker: React.FC = () => {
 
   return (
     <LanguagePickerWrapper>
-      {languagePickerData[market]
-        .map((item) => (
-          <Language key={item.linkTo} to={`/${item.linkTo}/forever/${code}`}>
+      {languagePickerData[market].map((item, index) => (
+        <React.Fragment key={item.linkTo}>
+          {index !== 0 && <Divider />}
+          <Language to={`/${item.linkTo}/forever/${code}`}>
             {item.shortTitle}
           </Language>
-        ))
-        .reduce((prevOptions, option) => (
-          <>
-            {prevOptions}
-            <Divider />
-            {option}
-          </>
-        ))}
+        </React.Fragment>
+      ))}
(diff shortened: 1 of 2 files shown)


src/client/pages/Forever/components/LanguagePicker.tsx
```diff
@@ -35,20 +36,21 @@ export const LanguagePicker: React.FC = () => {
...
-      <Language to={`/se/forever/${code}`}>Sv</Language>
-      <Divider />
-      <Language to={`/se-en/forever/${code}`}>En</Language>
+      {languagePickerData[market]
+        .map((item) => (
+          <Language key={item.linkTo} to={`/${item.linkTo}/forever/${code}`}>
+            {item.shortTitle}
+          </Language>
+        ))
+        .reduce((prevOptions, option) => (
```
styrelius, line 47: Is the `reduce` just to add the divider correctly? IMO that might be unnecessarily complex. What do you think about using `index` and conditionally render the `<Divider />` based on that? Something like
```suggestion
        .map((item, index) => (
        <>
            <Language key={item.linkTo} to={`/${item.linkTo}/forever/${code}`}>
              {item.shortTitle}
            </Language>
            {index === 0 && <Divider />}
          </>
        ))
```
Or am I missing some other awesomeness with the `reduce`?

styrelius (commented):
Awesome!! 🧠 ✨ 
Just a question on... iterating in JSX and adding elements between other elements? 😄

src/client/pages/Forever/components/LanguagePicker.tsx
robinandeer: Yeah I agree, it was a test but I think you solution is more straight forward so let's go with that!

styrelius: Oh yes of course, the `key` 🤦‍♀️

gustaveen (approved):
🙏

PR #526, opened 2021-05-06
Update danish translation

gustaveen:
## What?

Various updates :)


src/translations/da_DK.json
```diff
@@ -173,7 +173,7 @@
   "DETAILS_MODULE_EXTRABUILDINGS_TABLE_SIZE_CELL_LABEL_HOUSE": "Størrelse",
   "DETAILS_MODULE_EXTRABUILDINGS_TABLE_WATER_CELL_LABEL_HOUSE": "Indlagt vand",
   "SIDEBAR_INSURANCE_TYPE_RENT": "Lejebolig",
-  "SIDEBAR_INSURANCE_TYPE_BRF": "Indboforsikring",
+  "SIDEBAR_INSURANCE_TYPE_BRF": "Bolig jeg ejer",
```
gustaveen, line 176: We should look at changing this text keys since it only makes sense for SE 😁

robinandeer (approved):
🍀

styrelius (changes requested):
I'm not sure about this so I'm blocking this hehe

src/translations/da_DK.json
styrelius: Oh, but I don't think this should be changed actually since I think it's used in other places as the type of insurance. If this is the one messing up the details modal then the details modal needs another text key.

styrelius: But I might be wrong though

gustaveen: I did check where it's used only in the detailsmodal except from here what I could see:
Screenshot 2021-05-06 at 15 47 35

styrelius: No it's fine - it's just a bit messy with this keys 😅 Sorry for being dramatic LOL

gustaveen: So wouldn't mess anything up as far as I understand 🙃

styrelius (approved):
Sorry again, I remembered incorrectly 🤕

src/translations/da_DK.json
gustaveen: For displaying danish insurance titles we use these textkeys:
Screenshot 2021-05-06 at 15 51 29

styrelius: But we really shouldn't use that text key in the details modal honestly since it's more of referring to the type of insurance contract/quote and in the details modal we just want to say what type of home/residence it is

styrelius: Anyways, it doesn't matter now, never mind me hehehe

gustaveen: For sure, let's change that later, makes no sense 😄

PR #528, opened 2021-05-06
Reverse order of symbol and label in Card component

robinandeer:
## What?

Add new Card story without badge.
Reverse flex direction in CardHeader on large viewports.

## Why?

This was the easiest way to accomplish it. Not sure if it's a good or bad idea 😅

Screenshot 2021-05-06 at 16 01 03

Screenshot 2021-05-06 at 16 00 56

### DIFF
--- a/src/client/pages/Landing/components/Card.tsx
+++ b/src/client/pages/Landing/components/Card.tsx
@@ -80,6 +80,7 @@ const CardHeader = styled.div`
   }
 
   @media (min-width: 1020px) {
+    flex-direction: row-reverse;
     padding-bottom: 3rem;
     svg {
       display: block;
--- a/src/client/pages/Landing/components/Card.tsx
+++ b/src/client/pages/Landing/components/Card.tsx
@@ -71,6 +71,7 @@ const CardHeader = styled.div`
   min-height: 1.5rem;
   padding-bottom: 0.5rem;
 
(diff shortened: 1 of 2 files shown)


src/client/pages/Landing/components/Card.tsx
```diff
@@ -80,6 +80,7 @@ const CardHeader = styled.div`
   }
 
   @media (min-width: 1020px) {
+    flex-direction: row-reverse;
     padding-bottom: 3rem;
     svg {
       display: block;
```
gustaveen, line 86: This works fine I think 🙌. Alternatively we can just switch the order of `<Badge>` and `<HedvigSymbol>` in the component and set `margin-left:auto` on the svg, if that feels simpler 🙃
```suggestion
  @media (min-width: 1020px) {
    padding-bottom: 3rem;
    svg {
      display: block;
      margin-left: auto;
```

robinandeer: I just want to make it clear - it's not obvious what "row-reverse" does here :P

Will sleep on it - thanks for the suggestion!

gustaveen (dismissed):
➡️   🎉

vonElfvin (approved):
Hihi

styrelius (approved):
🇩🇰 🎊

PR #533, opened 2021-05-10
Fit `Card` headline on single line (as much as possible)

robinandeer:
## What?

Group card components in same module.
Update Card stories with proper text components.
Unify breakpoints in card component.
Align arrow icon with paragraph text.
Make headline text slightly smaller on large screens.

## Why?

To fit the card headlines on a single line (for native market languages at least).

I grouped all the components together since they are now so linked in the layout.

## Sweden

Screenshot 2021-05-10 at 12 08 31

Screenshot 2021-05-10 at 12 08 24

## Norway

Screenshot 2021-05-10 at 12 08 42

Screenshot 2021-05-10 at 12 08 47

## Denmark

Screenshot 2021-05-10 at 12 08 53

Screenshot 2021-05-10 at 12 09 02

## Mobile

Screenshot 2021-05-10 at 12 09 48

## Tablet

Screenshot 2021-05-10 at 12 10 09

### DIFF
--- a/src/client/pages/Landing/components/Card.tsx
+++ b/src/client/pages/Landing/components/Card.tsx
@@ -76,7 +76,7 @@ const CardHeader = styled.div`
     display: none;
   }
 
-  @media (min-height: 620px) {
+  @media (min-height: 600px) {
     padding-bottom: 1rem;
   }
 
@@ -93,27 +93,31 @@ const CardHeader = styled.div`
   }
 `
 
-const CardContent = styled.div`
-  padding-right: 2rem;
-`
+const CardContent = styled.div``
 
 const CardLink = CardComponent.withComponent(Link)
 
 const ArrowWrapper = styled.span`
   position: absolute;
   right: 1.25rem;
-  bottom: 1.25rem;
+  bottom: 0.8rem;
 
   @media (min-width: 600px) {
     right: 1.5rem;
-    bottom: 2rem;
+    bottom: 0.7rem;
+
     svg {
       font-size: 1.75rem;
     }
   }
 
+  @media (min-width: 600px) {
+    bottom: 1.2rem;
+  }
+
   @media (min-width: 1020px) {
     right: 2rem;
+    bottom: 2rem;
   }
 `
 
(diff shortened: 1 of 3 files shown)


src/client/pages/Landing/components/Card.tsx
```diff
@@ -76,7 +76,7 @@ const CardHeader = styled.div`
     display: none;
   }
 
-  @media (min-height: 620px) {
+  @media (min-height: 600px) {
```
gustaveen, line 79: Wasn't all `@media (min-height:` set to `620px` before? :) Or maybe this change makes sense, just wanted to check that you don't mix it up with the `min-width` media query that's set to `600px`

src/client/pages/Landing/components/Card.tsx
```diff
@@ -128,11 +132,58 @@ const CardContainer: React.FC<{
   )
 }
 
-export const Card: React.FC<{
+export const CardHeadline = styled.h2<{ disabled?: boolean }>`
```
gustaveen, line 135: Just for the sake of keeping the css before the react components, maybe put this above `CardContainer` :)

src/client/pages/Landing/components/Card.tsx
```diff
@@ -128,11 +132,58 @@ const CardContainer: React.FC<{
...
+  color: ${(props) => (props.disabled ? colorsV3.gray500 : colorsV3.gray900)};
+
+  @media (min-width: 600px) {
+    margin-bottom: 0.25rem;
+    font-size: 1.5rem;
+    line-height: 1.25;
+  }
+
+  @media (min-width: 1020px) {
+    font-size: 1.9rem;
```
gustaveen, line 149: This solves the layout issue for the DK case, just wondering if we should go outside our defined font-size? (You don't know about them of course so hard for you to know:D) https://www.figma.com/file/zlX4FcM5dMRz9stsCNQJXB/Web-UI-Kit?node-id=0%3A1 I created a review app so we can let Mange have a look.

gustaveen (commented):
Good job Robin! 🙌  Just some thoughts :)

src/client/pages/Landing/components/Card.tsx
robinandeer: Good call! Will change back

robinandeer:
What do you think about the latest solution?

We keep the 32px font size but remove the edge-margin for the first/last Card. I also moved the responsibility to add margin between `Card` to the `CardContainer` -- how does that work in your mind?

## Updated screenshots

Screenshot 2021-05-10 at 16 07 50

Screenshot 2021-05-10 at 16 07 57

gustaveen (approved):
🙌

PR #535, opened 2021-05-10
Prevent peril text overflow

robinandeer:
## What?

Grow peril box to fit long text.
Create space between icon + text on medium-sized screens.

> Does not fix issue when there's a long word that doesn't fit in the box

## Why?

The layout looks broken if the text overflows (happens on tablets).

## Before

Screenshot 2021-05-10 at 15 15 04

## Mobile

Screenshot 2021-05-10 at 15 25 46

## Tablet - portrait

Screenshot 2021-05-10 at 15 26 01

## Tablet - landscape

Screenshot 2021-05-10 at 15 26 17

## Desktop

Screenshot 2021-05-10 at 15 26 58


styrelius (commented):
But shouldn't we also cut words that are too long? I know Mange said he will change the copy of this one to use more shorter words instead of this long one but isn't it a good idea to do the `text-overflow: ellipsis;` as a fallback of some sort? Or should we just always use shorter words? 😄 
Screenshot 2021-05-10 at 15 33 15

gustaveen (approved):
Boom, clean and simple! 🙌

PR #536, opened 2021-05-11
Fix / Edit all bundle quotes

styrelius:
## What?

Fix edit details functionality so that all quotes, regardless of whether or not they are part of a bunde, are updated with all information in Details Modal form.
- Use the same function `onSubmit` in `/DetailsModal/index.tsx` for running `editQuoteMutation` regardless of number of quotes
- Loop through all quotes and run mutation for each of them
- Explicitly pass in personal details (`firstName`, `lastName` and `birthDate`) as input to the mutation, in addition to the nen quote details

## Why?

Currently we only update the "main" quote in a bundle (i.e. the home content quote) with new personal information if you edit them in the Details Modal. The result is that it's impossible to sign, since the information on the quotes will differ after the edit which makes back-end scream (and rightly so).

_Referenced ticket(s): [GRW-318]_

### Previously after editing personal details with a bundle:
2021-05-10-edit-quote-bundle-fail

### With changes in this PR:
2021-05-11-edit-quote-bundle-success

[GRW-318]: https://hedvig.atlassian.net/browse/GRW-318


src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/index.tsx
```diff
@@ -118,36 +118,23 @@ export const DetailsModal: React.FC<ModalProps & DetailsModalProps> = ({
...
-    if (
-      hasEditQuoteErrors(mainQuoteResult) ||
-      isUnderwritingLimitsHit(mainQuoteResult)
-    ) {
-      return mainQuoteResult
-    }
-    const quoutesToUpdate = offerData.quotes.filter(
-      (quote) => quote.id !== form.id,
-    )
+  const editQuotes = async (form: EditQuoteInput, offerData: OfferData) => {
```
gustaveen, line 121: 🙌 Makes a whole lot of more sense! I don't understand why we did it like that in the first place? 🤔

gustaveen (commented):
Wohoo! Great stuff Ellen and Robin! 🙏 Just gonna try it out in quickly in the review app

gustaveen (approved):
Works like a charm! 🎉

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/index.tsx
styrelius: No it's a mystery 🤷‍♀️

PR #537, opened 2021-05-11
GRW-270 Update bg image on Landingpage

gustaveen:
## What?
Add new background pictures with different crops for landscape and portrait orientation. Made use of `<picture>` element to be able to serve different crops depending on orientation.

## Why?
More control on what images are served on what devices and hence loading smaller images for mobile. `<picture>` element has the same browser support as `srcset`. To have better backwards compatibility I changed from `srcset` to `src` on the fallback image. 

The `onLoad` event will still trigger since the way it works with `picture` and `source` is that the browser switches the `src` on the `` calculated on the images and conditions you provide. 

The image sizes and crops are based off the most common screen resolutions we have https://datastudio.google.com/reporting/e4b4bd75-04f9-4479-89fb-01c9476f68f3
The browser calculates what image to choose based on screen size and pixel density.

### Mobile
Screenshot 2021-05-11 at 14 47 23

### Tablet
Screenshot 2021-05-11 at 15 23 18

### Laptop 1440
Screenshot 2021-05-11 at 14 49 11


styrelius (approved):
🖼️ 👏 👏

PR #541, opened 2021-05-12
Feat / New hero image on offer page

styrelius:
## What?

New shiny hero image on `/offer` page :)

- A new `BackgroundImage` is created, in its ow file under `/components`. This is constructed so that it can be used for both full screen background images (like in `Landing`), or a hero image like in the offer `Introduction` component .
- The old image components in `Landing.tsx` and `/OfferNew/Introduction/index.tsx` are replaced with this new component.

## Why?

The reason for this refactoring is that we want these images (the background on landing page and hero on offer page) to function in the same way, with the same transition animation etc. So to me it made sense to extract it into a slightly more generic component and use that in both places. It also makes the parent components smaller and easier to get an overview of.

If we in the future would like different images on these pages it's also pretty easy to just add the image sources as props to  `BackgroundImage` so that they will still have the same feeling to them and work the same way.

_Referenced ticket(s): [GRW-270]_

## Screenshots

**desktop:**

2021-05-12-offer-new-hero-desktop

**mobile (iPhone X according to Google DevTools:**

2021-05-12-offer-new-hero-mobile

[GRW-270]: https://hedvig.atlassian.net/browse/GRW-270


src/client/components/BackgroundImage.tsx
```diff
@@ -0,0 +1,76 @@
+import React, { useState } from 'react'
+import styled from '@emotion/styled'
+import { css } from '@emotion/core'
+import { colorsV3 } from '@hedviginsurance/brand'
+import { LARGE_SCREEN_MEDIA_QUERY } from 'utils/mediaQueries'
```
robinandeer, line 5: I missed that we had defined these - nice!

robinandeer (commented):
Nice! I think it's an improvement :D

One thing to consider is just allowing styles to be overwritten by passing "className" to the `BackgroundImage` components and using it like so:

```javascript
const LandingBackgroundImage = styled(BackgroundImage)`
  z-index: -1;
`
```

robinandeer (approved):

src/client/components/BackgroundImage.tsx
styrelius: Yep, I added them a while back but it's only me using them I think lol

gustaveen:
Very nice to make a component for this @styrelius 🙌 One thing I thought about is if we should keep the overlay though instead of having opacity on the image. That also makes it easier to match the design. This image has a 0.3 black opacity overlay from the design in Figma. Might be nice to add the opacity as a prop also :)

styrelius:
> Very nice to make a component for this @styrelius 🙌 One thing I thought about is if we should keep the overlay though instead of having opacity on the image. That also makes it easier to match the design. This image has a 0.3 black opacity overlay from the design in Figma. Might be nice to add the opacity as a prop also :)

I'm not sure I follow - instead of having a black background and then making the image less opaque you'd like an overlay on the image? What would be the difference/advantage? 

But I actually on purpose made the image darker than you did it on the landing page since I found it hard to read the white text otherwise.

gustaveen:
I think I'm just used to working with overlays since that maps to how designers adds overlays :) But as you say, it looks the same if you have always have a black background, and for the web-onboarding that's the case 👍

So it's just that you have to keep in mind to invert the value then, so for example if it's 0.3 opacity overlay in the design, you will have 0.7 opacity on the image to mimic the same look

styrelius:
@gustaveen Ah, yes I thought since we do have the black background and then load the image it was easier so set the opacity on the image. But we can of course add an overlay instead.

styrelius:
@gustaveen  And again, setting the `opacity` of the image to `0.6` and not `0.7` was intentional since I had problems reading the white text with a more opaque (i.e. lighter) image. But maybe I should have checked that with Mange first 😅

PR #542, opened 2021-05-14
Fix edit quote mutation in SE

robinandeer:
## What?

Also: Skip passing `offerData` to `editQuote` function. Data is already in scope.

## Why?

Currently, editing quote details doesn't work in Sweden.

Reference: https://sentry.io/organizations/hedvig/issues/2396702298/?project=1308938&query=is%3Aunresolved

### DIFF
--- a/src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/index.tsx
+++ b/src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/index.tsx
@@ -9,7 +9,7 @@ import { EditQuoteInput, useEditQuoteMutation } from 'data/graphql'
 import { LocaleLabel, locales } from 'l10n/locales'
 import { OfferData } from 'pages/OfferNew/types'
 import { useTextKeys } from 'utils/textKeys'
-import { getMainQuote } from '../../../utils'
+import { getMainQuote, isSwedish } from '../../../utils'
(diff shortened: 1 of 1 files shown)


styrelius (approved):
Good quick fix! 👌

Another solution is to fix the `getFormData` function by adding the Swedish offer/quote types, since that's what seems to be causing the issue:
```javascript
export const getFormData = (form: EditQuoteInput, offerData: OfferData) => {
  if (isSwedishApartment(offerData.quotes[0].quoteDetails))
    return form.swedishApartment
  if (isSwedishHouse(offerData.quotes[0].quoteDetails)) return form.swedishHouse
  if (isDanish(offerData)) return form.danishHomeContents
  if (isNorwegian(offerData)) form.norwegianHomeContents
  throw new Error('Unknown market/quote type')
}
```

But hey, this should probably be deeply refactored so I don't know if that's better, probably not 🤷‍♀️

styrelius:
And again, setting the `opacity` of the image to `0.6` and not `0.7` was intentional since I had problems reading the white text with a more opaque (i.e. lighter) image. But maybe I should have checked that with Mange first 😅

robinandeer:
Yeah I agree, will probably have to rethink some of this when we move to even more markets so perhaps this solution is okey for now!

PR #543, opened 2021-05-14
Correctly check if there's a discount

robinandeer:
## What?

-

## Why?

Bundles in Denmark indicated there would be a discount even if the original price was the same as the "discounted" price. Only bundles in Norway come directly with a discount.

### Denmark

Screenshot 2021-05-14 at 10 33 37

### Norway

Screenshot 2021-05-14 at 10 35 30


gustaveen (approved):
🙌

PR #546, opened 2021-05-17
GRW-293 - track DK bundles

gustaveen:
## What?

Make sure we track DK bundles. A pretty naive solution but checking the actual contents of the danish bundles got a bit messy so checking for the length does the trick IMO :) The naming of the different util functions makes it a bit confusing since some takes `offerData` as an argument and some takes a `quote` or `quoteDetails`

## Why?

Must have for going live 

#### Accident bundle student
Screenshot 2021-05-17 at 16 08 18

#### Accident bundle
Screenshot 2021-05-17 at 16 08 53

#### Home contents
Screenshot 2021-05-17 at 16 09 34


styrelius (approved):
🥷 (closest thing to a spy emoji?)

robinandeer (approved):
Nice work! I agree this works for now but super important we don't forget to refactor when these assumptions are no longer true (e.g. danish bundle with two quotes is always "travel bundle")

PR #547, opened 2021-05-18
Fix / Connect Payment page copy plus styling

styrelius:
## What?

- New copy on `/connect-payment` page that hopefully makes it more clear that you are now a Hedvig member, but you still need to connect payment and download the app.
- Some styling fixes to adjust to the new text (many of the sentences are longer now), and to make it look better in more viewports.

## Why?

We've gotten feedback from the IEX team that it's not clear to people after they've signed that they actually are Hedvig members. These copy changes aim to address that issue.

_Referenced ticket(s): [GRW-276]_

## Screenshots

**/se-en:**
2021-05-18-connect-payment-mobile-se-en

**/no:**
2021-05-18-connect-payment-mobile-no

**/se:**
2021-05-18-connect-payment-desktop-se

**/no:**
2021-05-18-connect-payment-desktop-no

**/dk:**
2021-05-18-connect-payment-desktop-dk

**/dk-en**
2021-05-18-connect-payment-desktop-dk-en

[GRW-276]: https://hedvig.atlassian.net/browse/GRW-276


robinandeer (approved):
Nice work! A lot of changes to try to keep track of but I can't really find something that needs changing.

I like that you are cleaning up the paddings to be more like "8rem" rather than "7.25rem"!

styrelius:
> Nice work! A lot of changes to try to keep track of but I can't really find something that needs changing.
> 
> I like that you are cleaning up the paddings to be more like "8rem" rather than "7.25rem"!

Haha I know, I'm sorry - I moved too much things around 😳 

Regarding the `rem` thing - I think we should scrap that altogether since it makes no sense to use it like we do it. IMO we should just have fixed pixel values for all measurements centralized. But since that requires more of a "helhetsgrepp" and it hasn't been popular before when I've replaced `rem` with `px` I tried to make the best of it 😅

PR #548, opened 2021-05-19
Update translations for DK bundle

robinandeer:
## What?

Change wording about when it's available.

## Why?

We will not ship in May.

Screenshot 2021-05-19 at 14 20 32
Screenshot 2021-05-19 at 14 20 19


styrelius (approved):
GREAT WORK 😀

PR #549, opened 2021-05-20
GRW-326 / Fix / Extra building on Swedish house quote

styrelius:
## What?

- New types etc. from updated GraphQL schema
- New `type` property added when querying for `ExtraBuildingValue` in `QuoteBundle` and `Quote` query (`ExtraBuildingValue` is replacing `ExtraBuildingCore` since for some reason Apollo client removed the `extraBuilding` data from the response when using `ExtraBuildingCore` 🤷‍♀️) .
- This `extraBuildings` `type` property is used as an initial form value for the Details Modal, instead of broken mapping between `__typename` and the `ExtraBuildingType`

## Why?

After updates to GraphQL schema(s) the implementation we had for dealing with the `extraBuildings` property, which exists on the Swedish house quotes.

_Referenced ticket(s): [GRW-326]_

<hr />

**Review app:** https://web-onboardi-grw-236-fi-9aqzci.herokuapp.com/se/new-member

## Screen recording

https://user-images.githubusercontent.com/42962286/118983507-ce1ca880-b97c-11eb-8fef-ec3b1b6d17f0.mp4

<hr />

_Additional note:_ I realised that you have a lot more options for extra buildings in the Details Modal than in the Embark flow. Probably something we should look into since they of course shouldn't differ.

[GRW-326]: https://hedvig.atlassian.net/browse/GRW-326


src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
```diff
@@ -133,7 +133,7 @@ const getHouseExtraBuildingsMaybe = (
       key: 'buildingSize',
       label: textKeys.CHECKOUT_DETAILS_EXTRA_BUILDINGS_SIZE(),
       value: textKeys.CHECKOUT_DETAILS_SQM_VALUE({
-        VALUE: extraBuilding.area ?? 'blah',
```
robinandeer, line 136: Nice catch!

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/utils.ts
```diff
@@ -575,12 +555,14 @@ export const getInitialSwedishHouseValues = (
...
-        hasWaterConnected: b.hasWaterConnected,
-      })),
+      .filter((building) => Boolean(building.__typename))
+      .map(({ type, area, hasWaterConnected }) => {
+        return {
+          type,
+          area,
+          hasWaterConnected,
+        }
+      }),
```
robinandeer, line 565: Do we even need to do any mapping here?

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/utils.ts
```diff
@@ -575,12 +555,14 @@ export const getInitialSwedishHouseValues = (
     yearOfConstruction: details.yearOfConstruction,
     isSubleted: details.isSubleted,
     extraBuildings: details.extraBuildings
-      .filter((b) => !!b.__typename)
-      .map((b) => ({
-        type: extraBuildingTypes[b.__typename!],
-        area: b.area,
-        hasWaterConnected: b.hasWaterConnected,
-      })),
+      .filter((building) => Boolean(building.__typename))
```
robinandeer, line 558: I wonder where this comes from. The previous commit message says:

> "Get debugger and norwegian inbo quote to work kida"

😅 I can't figure out why it's needed..

robinandeer (dismissed):
Cool stuff! I tried out the house flow in the review app and everything worked

There's a reference to `ExtraBuildingCore` in "MemberOffer.graphql" as well I noticed. Don't know if we want to replace that one also?

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/utils.ts
styrelius: No you're probably right that we don't need that anymore.

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/utils.ts
styrelius: Yeah I don't know. So much weird and overly complex stuff going on in these files IMO 😭 But I don't know if it's worth digging in now? Probably not right?

styrelius:
> There's a reference to ExtraBuildingCore in "MemberOffer.graphql" as well I noticed. Don't know if we want to replace that one also?

Oh, yes I think I saw that. But we don't use that query right? Maybe we should delete it?

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/utils.ts
styrelius: It feels like it's just some kind of check to see that the `building` exists? 🤔

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/utils.ts
styrelius: OK so this could be removed 😀

src/client/pages/OfferNew/Introduction/Sidebar/DetailsModal/utils.ts
styrelius: And this could also be removed 😀 

Tested Swedish house flow with and without extra buildings, and also created a Norwegian quote through the debugger after removing it and all of this worked fine.

gustaveen: Yes, amazing! That was so confusing :D

gustaveen (approved):
Great job with all this Ellen! 🏡 Worked perfect for me as well🙌

PR #555, opened 2021-05-27
Redirect users that fail to get quote

robinandeer:
## What?

Redirect users to start page if create quote returns "UnderwritingLimitsHit".
Capture error to Sentry.

## Why?

Currently, we throw an exception that we never catch so users end up with a blank page.

I didn't find any generic error page/component to show the user. I noticed a similar case where we simply redirect the user so I copied that solution.

This "error" should really be handled in Embark (or better yet, the GraphQL API should return a proper error 😄). I do believe that this is an improvement on the existing solution. If we fix things in Embark/Giraffe - no users need to end up in this if-block.

_Referenced ticket(s): []_

### DIFF
--- a/src/client/pages/OfferNew/index.tsx
+++ b/src/client/pages/OfferNew/index.tsx
@@ -22,6 +22,7 @@ import { getOfferData } from 'pages/OfferNew/utils'
 import { useVariation, Variation } from 'utils/hooks/useVariation'
 import { trackOfferGTM } from 'utils/tracking/gtm'
 import { getUtmParamsFromCookie, TrackAction } from 'utils/tracking/tracking'
+import { captureSentryError } from 'utils/sentry-client'
 import { useQuoteIds } from '../../utils/hooks/useQuoteIds'
 import { LanguagePicker } from '../Embark/LanguagePicker'
 import { Checkout } from './Checkout'
@@ -72,9 +73,10 @@ export const OfferNew: React.FC = () => {
   }
 
   if (!loadingQuoteBundle && !data?.quoteBundle) {
-    throw new Error(
+    captureSentryError(
       `No quote returned to show offer with (quoteIds=${quoteIds}).`,
     )
+    return <Redirect to={`/${currentLocale}/new-member`} />
   }
 
   const handleCheckoutToggle = (open: boolean) => {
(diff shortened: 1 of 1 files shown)


src/client/pages/OfferNew/index.tsx
```diff
@@ -72,9 +73,10 @@ export const OfferNew: React.FC = () => {
   }
 
   if (!loadingQuoteBundle && !data?.quoteBundle) {
-    throw new Error(
+    captureSentryError(
       `No quote returned to show offer with (quoteIds=${quoteIds}).`,
```
styrelius, line 77: This message is a bit strange? Maybe we could write something like 
```suggestion
    captureSentryError(
      `No quote data could be returned for quoteIds: ${quoteIds}.`,
```

Also, could we be helped if we got the `error` from the query on line 52 and logged that to Sentry as well?

styrelius (commented):
Oh,  I had forgotten we did this thing here 😳 But I'm thinking if it's just as strange to suddenly be back at the landing page when you've completed the flow as getting the white screen? Meaning that maybe we shouldn't change anything until we've built some error page to show on these occasions 🤔

And two other thoughts that you can dismiss if you want to 👇

robinandeer:
@styrelius: I totally agree that it's not great UX to redirect back to the Landing page. I would say that it's preferred to letting the app simply crash though 🤔  I think that hurts the brand more, no?

And like I mentioned, hopefully, we can avoid letting any user end up here by making changes in Embark/Giraffe.

The redirect was the first fix I thought of. Perhaps together we can think of a better solution? Also having a generic error component we can fall back to would be good I think - perhaps we can ask Mange to look into that?

src/client/pages/OfferNew/index.tsx
robinandeer: @styrelius the error I got is quite strange - something like "Currency Code may not be null". Currently, we log all GraphQL request errors to Sentry so we will see it there.

styrelius: Oh, yes of course 🧠

styrelius:
@robinandeer Hm, I don't know but maybe you're right that this is better. So let's go with this. And yes, I think it's a good idea to ask Mange to look at an error component/page, or we could make a simple one ourselves. But it's probably faster to ask Mange to make one 👍

styrelius (approved):
🥇

robinandeer:
We handle UW hits so let's not continue with this.

PR #556, opened 2021-05-27
Feature/adtraction dk

gustaveen:
## What?

- Add all possible Adtraction values to `adtractionProductMap` for better overview
- Load script with danish `ap` and don't render the script for non existing markets

## Why?

Track affiliates in Denmark 

https://user-images.githubusercontent.com/6661511/119842242-cd4cbf00-bf06-11eb-9f36-5a4d1b96bda6.mov

### DIFF
--- a/src/client/utils/tracking/tracking.ts
+++ b/src/client/utils/tracking/tracking.ts
@@ -12,7 +12,6 @@ import { OfferData } from 'pages/OfferNew/types'
 import {
   isBundle,
   isYouth,
-  isDanish,
   isNorwegian,
   isDanishAccidentBundle,
   isDanishTravelBundle,
@@ -101,15 +100,8 @@ export const { TrackAction, IdentifyAction } = setupTrackers<
 )
 
 type TypeOfContractExcludedUnused = Exclude<
-  TypeOfContract,
-  | 'DK_HOME_CONTENT_OWN'
-  | 'DK_HOME_CONTENT_RENT'
-  | 'DK_HOME_CONTENT_STUDENT_OWN'
-  | 'DK_HOME_CONTENT_STUDENT_RENT'
-  | 'DK_ACCIDENT'
-  | 'DK_ACCIDENT_STUDENT'
-  | 'DK_TRAVEL'
-  | 'DK_TRAVEL_STUDENT'
+  TypeOfContract | NoComboTypes | DkBundleTypes,
+  'DK_ACCIDENT' | 'DK_ACCIDENT_STUDENT' | 'DK_TRAVEL' | 'DK_TRAVEL_STUDENT'
 >
 
 const adtractionProductMap: Record<TypeOfContractExcludedUnused, number> = {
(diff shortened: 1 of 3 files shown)


src/server/page.ts
```diff
@@ -147,14 +148,22 @@ export const getPage = (
...
   const ADTRACTION_SE = 'https://cdn.adt387.com/jsTag?ap=1412531808'
+  const ADTRACTION_DK = 'https://cdn.adt387.com/jsTag?ap=1589794294'
+  const adtractionTag = match([
+    ['se', ADTRACTION_SE],
+    ['se-en', ADTRACTION_SE],
+    ['no', ADTRACTION_NO],
+    ['no-en', ADTRACTION_NO],
+    ['dk', ADTRACTION_DK],
+    ['dk-en', ADTRACTION_DK],
+  ])(ctx.params.locale)!
```
robinandeer, line 161: Much easier to understand!

src/client/utils/tracking/tracking.ts
```diff
@@ -101,15 +100,8 @@ export const { TrackAction, IdentifyAction } = setupTrackers<
...
-  | 'DK_HOME_CONTENT_OWN'
-  | 'DK_HOME_CONTENT_RENT'
-  | 'DK_HOME_CONTENT_STUDENT_OWN'
-  | 'DK_HOME_CONTENT_STUDENT_RENT'
-  | 'DK_ACCIDENT'
-  | 'DK_ACCIDENT_STUDENT'
-  | 'DK_TRAVEL'
-  | 'DK_TRAVEL_STUDENT'
+  TypeOfContract | NoComboTypes | DkBundleTypes,
+  'DK_ACCIDENT' | 'DK_ACCIDENT_STUDENT' | 'DK_TRAVEL' | 'DK_TRAVEL_STUDENT'
```
robinandeer, line 104: So basically this type includes all the "promise stacks" that you can sign up for? I don't know any other word for it 😅 The excluded contracts can be added in addition to the core home contents product, yes?

robinandeer (dismissed):
Much improved! Still not 100% sure what the script is trying to do but I can follow the logic in the code at least :D

src/server/page.ts
styrelius, line 161: Couldn't this just be a plain object? Then we could also easily type the `ctx.params.locale`s with the `LocaleLabel` type which is exported from `/client/l10n/locales.ts`? Maybe 😃

styrelius: Like 
```typescript
const adtractionTag: Record<LocaleLabel, string> = {
  'se': ADTRACTION_SE,
  'se-en': ADTRACTION_SE,
  'no': ADTRACTION_NO,
  'no-en': ADTRACTION_NO,
  'dk': ADTRACTION_DK,
  'dk-en': ADTRACTION_DK
```
or something like that 😀

styrelius: Or these values (the url's) could even be added to the `locales` object?

gustaveen: Sure, we can do that! My only concern is that the `locales` object grows to contain a lot of different things, but I guess we can split it up later if we feel that way.

src/client/utils/tracking/tracking.ts
gustaveen: Yes! This is one of the confusing things with the existing setup. Since `'DK_ACCIDENT' | 'DK_ACCIDENT_STUDENT' | 'DK_TRAVEL' | 'DK_TRAVEL_STUDENT'` are contract types but not sold as standalone products and only in bundles, we don't want to track them.

gustaveen:
> Much improved! Still not 100% sure what the script is trying to do but I can follow the logic in the code at least :D

🙌  Hehe I hear you, Adtraction will validate that it works once it's live

src/server/page.ts
styrelius: Yeah my idea was to split it up in like on file for each locale object or something if they become to big and just import them and put them together in `localse.ts`. To me it makes sense to collect _all_ locale specific data in one place but you already know that since it was me who started building that structure haha 😄

gustaveen: Sounds good! 🙌 Or have slice it on specific to features like ssn.

robinandeer (approved):

src/server/page.ts
```diff
@@ -40,7 +41,7 @@ const clientConfig: ClientConfig = {
 const template = (
   route: ServerSideRoute,
   locale: string,
-  adtractionTag: string,
+  adtractionTag: string | null,
```
styrelius, line 44: What's the benefit from typing this as `string | null` instead of making it optional with a `?` (which of course means that the type is really `string | undefined`)? If we would just have it optional there wouldn't be the need to use the nullish coalescing operator on line 152?

styrelius (approved):
Nice! 😀 Just a question below that can totally be dismissed since it's very minor and I'm mostly curious

src/server/page.ts
gustaveen: I'm with you, that's what I did at first :D But TS got mad at me for doing that 🤷‍♂️

PR #558, opened 2021-05-31
Update app screenshots

robinandeer:
## What?

Correct spelling of addresses.

Screenshot 2021-05-31 at 13 28 41


styrelius (changes requested):
Something strange seems to have happened with the left image maybe when it was exported, it has a weird... pixelated(is that the right word?) area here:
Screenshot 2021-05-31 at 15 27 28

Maybe we should just ask Mange for images? If the issue was that we copied them from one Figma space to another 🤔

styrelius:
For reference - this is the image in staging:
image

gustaveen (approved):
🖼 👍

styrelius (approved):
Beautiful!! ⭐

PR #562, opened 2021-06-16
Adress autocomplete: Remove possibility to edit address from Danish Details Modal

styrelius:
### Please note that this should be merged and deployed at the same as the address autocomplete feature in Embark
Hence we'll keep it as a draft until then.

## What?

- `'street'` and `'zipCode'` are removed from the `DanishDetails` component (meaning the Danish details modal form)
-  Same properties are also removed from the Danish forms schema in `/DetailsModal/utils.ts`, and omitted from the `PartialEditDanishHomeContentsInput` type.

## Why?

For the first iteration of the Address Autocomplete feature for Denmark we've decided to simply remove the option to edit your address in the Details Modal, since we only want addresses in the correct format and with a `bbrId` attached.

The plan for next iteration(/-s) is to implement autocomplete in the Details Modal as well.

_Referenced ticket(s): [GRW-228]_

[GRW-228]: https://hedvig.atlassian.net/browse/GRW-228


robinandeer (approved):
Snyggt!

PR #567, opened 2021-07-05
Danish Checkout Summary

styrelius:
## What?
- `floor` and `apartment` are added to `QuoteBundle` query on `DanishHomeContentsDetails` in order to render them as part of the checkout summary
- Layout of `InsuranceSummaryDetails` component is tweaked to make space for longer address lines. 
- A few unit tests are added to make sure the address string put together from `street`, `floor` and `apartment` is correct (and unchanged for other markets than Denmark). 
  _Note: I've realised that the more reasonable thing of course would be to  write component/UI tests for the entire `InsuranceSummaryDetails` component instead, so I might do that later but this is probably OK for now_
- Danish Debugger form is updated with floor and apartment

**Also (not strictly related to the updated Checkout Summary):**
- Debugger component currently called `Offer` is renamed `QuoteData` since it's too easy to mix it up with the actual offer page component.
- `Quote` query is removed since it's not used anywhere and probably won't ever be used again :)

## Why?

- We should display the entire address in the summary section in Denmark in order to not feel.. broken 🙃 

_Referenced ticket(s): [GRW-356]_

## Screenshots

2021-07-05-dk-checkout-summary-desktop

2021-07-05-dk-checkout-summary-mobile

[GRW-356]: https://hedvig.atlassian.net/browse/GRW-356


src/client/pages/Debugger/components/QuoteFormDenmark.tsx
```diff
@@ -13,8 +13,10 @@ export const initialDkHomeValues = {
   danishHomeContents: {
     coInsured: 0,
     livingSpace: 34,
-    street: 'Nørrebrogade 50',
-    zipCode: '1234',
+    street: 'Theodore Roosevelts Vej 1',
+    floor: '2',
+    apartment: 'tv',
```
robinandeer, line 18: I guess these can still be reset/erased, right?

styrelius: Yes, since they're not required to create a quote the values can be deleted on the `/debugger` page if you want to create a quote with an address that isn't an apartment.

styrelius: And all values in that `debugger` form can of course be edited 😀

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    })
+
+    it('returns null when quote is Norwegian Travel quote', () => {
+      const norwegianTravelQuote = mockedOfferDataNoCombo.quotes[0]
+
+      const norwegianTravelQuoteDetails = norwegianTravelQuote.quoteDetails
+
+      const address = getAddress(norwegianTravelQuoteDetails)
+      expect(address).toBe(null)
+    })
```
robinandeer, line 37: Do we call this function even for travel quote? 😅

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
```diff
@@ -16,14 +16,15 @@ import { useCurrentLocale } from 'components/utils/CurrentLocale'
...
-  width: 50%;
   color: ${colorsV3.gray500};
 `
 const Value = styled.div`
-  width: 50%;
   color: ${colorsV3.gray700};
   text-align: right;
 `
+const HorizontalSpacer = styled.div`
+  width: 48px;
```
robinandeer, line 26: Don't we use rems in this project?

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
```diff
@@ -200,6 +203,29 @@ const getQuoteDetails = (
...
+
+    if (floor || apartment) {
+      const formattedFloor = floor ? `${floor}.` : ''
+      const apartmentString = apartment ? ` ${apartment}` : ''
+
+      return `${street}, ${formattedFloor}${apartmentString}`
+    }
+
+    return street
+  }
```
robinandeer, line 224: Wouldn't this be the same?

```suggestion
  const { floor, apartment } = quoteDetails 
  if (floor || apartment) {
    const formattedFloor = floor ? `${floor}.` : ''
    const apartmentString = apartment ? ` ${apartment}` : ''

    return `${street}, ${formattedFloor}${apartmentString}`
  }
```

robinandeer: I would also add some comment here that this is Danish-specific logic since it lives outside the localisation module.

robinandeer (approved):
Looks really good! Great that you took the time to add the tests :D

I wrote down some comments but we could also merge first to setup for the test session.

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.test.ts
styrelius: No we don't but since TS doesn't understand that I added a check that it just returns `null` it the quote lacks address and thought it was best to add a check for that as well so I didn't mess up 🙈

robinandeer: Right, I understand. So the other way would be to place an "assert" statement at the top of the function. But this is probably "safer".

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
styrelius: Since it doesn't really make sense the way we use `rem` I thought I could use `px` in this file where there are no other measurements. Also, we do have `px` in a bunch of other places. Isn't the important thing that we keep to the 4px scale?

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
```diff
@@ -200,6 +203,29 @@ const getQuoteDetails = (
   return detailsGroups
 }
 
+export const getAddress = (quoteDetails: QuoteDetails) => {
+  if (!quoteDetailsHasAddress(quoteDetails)) {
+    return null
+  }
```
robinandeer, line 209: If we are really sure we wouldn't call this function with a quote lacking address:

```suggestion
  if (!quoteDetailsHasAddress(quoteDetails)) {
    throw new Error("Quote details need to include address fields")
  }
```

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.test.ts
styrelius: Aha, don't know about the `assert` - will check that out. But yes maybe in this case keep it? But I agree it looks weird.

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
styrelius: Wait, I feel blind - you just removed space, right?

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.test.ts
robinandeer: Haha I was thinking about Python 🙃. It's strange that JS doesn't have assertions -- but I wrote a comment in the `getAddress` function about doing the same thing in the JS way :)

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
styrelius: You can't remove the `in` check I think since those properties don't exist on all type of `quoteDetails` but only the Danish version.

styrelius: And I feel like this is an edge case if it should live in the localisation module or not - maybe we want to display apartment number for other markets as well?

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.test.ts
styrelius: Haha OK 😄 But sure, I'll throw an `Error` instead, that's probably better

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
robinandeer: Right, I just mean that other markets most likely format their addresses differently ;)

And regarding the `in` checks; won't the values just be `undefined` for non-Danish quotes?

styrelius: No, TypeScripts yells if you remove the `in` checks since those properties don't even exist on e.g. `SwedishApartmentQuoteDetails`.

But yes, you are of course correct that other markets might format their addresses differently. Maybe somewhat of a pre-optimisation to prepare for that now though? Or maybe not 🤔

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
gustaveen: I still think we should continue to use `rem` for the new code we write 🙂

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
```diff
@@ -200,6 +203,28 @@ const getQuoteDetails = (
...
+
+  const { street } = quoteDetails
+
+  if ('floor' in quoteDetails && 'apartment' in quoteDetails) {
+    const { floor, apartment } = quoteDetails
+
+    if (floor || apartment) {
+      const formattedFloor = floor ? `${floor}.` : ''
+      const apartmentString = apartment ? ` ${apartment}` : ''
```
gustaveen, line 218: Minor detail, just thought of the naming, should they use the same strucutre?
```suggestion
      const floorString = floor ? `${floor}.` : ''
      const apartmentString = apartment ? ` ${apartment}` : ''
```

gustaveen (commented):
Great job Ellen! 🙌  Awesome that you added tests as well ⭐️

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
styrelius: Hehe, thought about this but they actually aren't the same. The first one adds a `.` - meaning it's formatted differently. The other one just makes sure it's a string and not `null` or `undefined`. But maybe I'm overthinking lol 😅

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
styrelius: I think we should "fasa ut det" since we use it weirdly 🙂

src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx
gustaveen: That sounds reasonable! 🙌

PR #569, opened 2021-07-07
Fix / Offer page hero address line

styrelius:
## What?

- Display floor and apartment as well as street name + number in hero section of */offer* page (reuse logic from `InsuranceSummaryDetails` component)
- Tweak styling slightly, most importantly change font size of address text to be smaller in order for it to look better on more screen sizes

## Why?

Danish people (of course) want to see their entire address 🙃 

_Referenced ticket(s): [GRW-370]_

## Screenshots

**Desktop Chrome, DK**
2021-07-08-hero-address-dk-desktop-chrome

**iPhone 11 Safari, DK**
2021-07-08-hero-address-dk-iPhone-safari

**iPad 8 Safari, SE**
2021-07-08-hero-address-se-iPad-safari

**Google Pixel 4 Chrome, SE**
image

[GRW-370]: https://hedvig.atlassian.net/browse/GRW-370


src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
```diff
@@ -1,75 +1,84 @@
...
-
-const Heading = styled.h1`
+const Headline = styled.h1`
   margin: 0;
   text-transform: uppercase;
-  font-size: 0.75rem;
+  font-size: 1rem;
+`
+const OfferInfoWrapper = styled.div`
+  padding: 1rem 0;
```
styrelius, line 31: Hm actually I think it might look nicer with 
```suggestion
  padding-bottom: 2rem;
```
image

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
```diff
@@ -1,75 +1,84 @@
...
+
+  ${MEDIUM_SCREEN_MEDIA_QUERY} {
+    font-size: 2.5rem;
+  }
+`
 
 export const HeroOfferDetails: React.FC<Props> = ({ offerData }) => {
   const [numberCoInsured, setNumberCoInsured] = useState<number | null>(null)
-  const [street, setStreet] = useState<Address['street'] | null>(null)
+  const [address, setAddress] = useState<string | null>(null)
```
robinandeer, line 54: Do we really need state here? Can't we just derive it? Something like:

```
const { person, quotes } = offerData

const numberCoInsured = useMemo(() => {
  const householdSize = getHouseholdSize(quotes[0].quoteDetails)
  return householdSize - 1
}, [quotes[0].quoteDetails])

const address = useMemo(() => {
  const quoteWithAddress = quotes.find((quote) => quoteDetailsHasAddress(quote.quoteDetails))
  return quoteWithAddress ? getAddress(quoteWithAddress.quoteDetails) : null
}, [quotes])
```

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
```diff
@@ -1,75 +1,84 @@
...
-        {!!numberCoInsured && ` +${numberCoInsured}`}
-      </OfferInfo>
-      {street && <OfferInfo>{street}</OfferInfo>}
-    </Container>
+    <Wrapper>
+      <Headline>{textKeys.HERO_OFFER_DETAILS_HEADER()}</Headline>
+      <OfferInfoWrapper>
+        <NameAndCoInsured>
+          {person.firstName}
+          {Boolean(numberCoInsured) && ` +${numberCoInsured}`}
```
robinandeer, line 78: With the suggestion above we can simplify this:
```suggestion
          {numberCoInsured > 0 && ` +${numberCoInsured}`}
```

robinandeer (dismissed):
Looks good to me!

Do consider my comment about deriving rather than managing state -- I think it will simplify the code quite a bit!

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
```diff
@@ -1,75 +1,84 @@
...
 import { OfferData } from 'pages/OfferNew/types'
 import { useTextKeys } from 'utils/textKeys'
-import { LARGE_SCREEN_MEDIA_QUERY } from 'utils/mediaQueries'
-import { getHouseholdSize } from '../utils'
+import {
+  LARGE_SCREEN_MEDIA_QUERY,
+  MEDIUM_SCREEN_MEDIA_QUERY,
+} from 'utils/mediaQueries'
+import { getHouseholdSize, quoteDetailsHasAddress } from '../utils'
+import { getAddress } from '../Checkout/InsuranceSummaryDetails'
```
robinandeer, line 11: Should this function be moved to the more central utils module? Just to make it clear that it is not just used by `InsuranceSummaryDetails` anymore

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
styrelius: It might be smarter to do it this way but I disagree that this would simplify the code - to me this feels a lot more complicated actually.

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
styrelius: Yep that's more reasonable 👍

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
styrelius: Probably, I'm not sure what approach we should have though. I'm really not a fan of the enormous `utils` files we have in a bunch of places in this repo. Would probably rather have a `/utils` folder or something, with more files with less functions in them. Would make it easier to navigate I think. And also easier to move things around maybe? What do you think?

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
styrelius: Oh, now I realised why I did it the way I did it - `numberCoInsured` can also be `null`

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
styrelius: But it's probably unnecessary to have two states since they both will always change at the same time. Could do one state and an object instead.

styrelius: Or rather (sorry for this stream of consciousness lol) you're probably right about the state being unnecessary, but the answer, I believe, is not doing `useMemo`'s but just getting the values 🙃   
```typescript
const { person, quotes } = offerData
  const numberCoInsured = getHouseholdSize(quotes[0].quoteDetails) - 1

  const quoteWithAddress = quotes.find((quote) => {
    return quoteDetailsHasAddress(quote.quoteDetails)
  })
  const address = quoteWithAddress
    ? getAddress(quoteWithAddress.quoteDetails)
    : null
```

robinandeer: Yeah, I don't think `useMemo` is necessary either :-)

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
gustaveen: Sounds good to me to split up large utils files to more specific files in a folder. 🙌

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
gustaveen: Aha! I'm good with using the double `!!` then. Or do we feel that it's too unclear to use that?

robinandeer: Right, but only if we are setting it up with some hooks state, no?

If we derive state directly from the offer data, it will always be defined I think

gustaveen:
Looks really good! ⭐️

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
robinandeer: Yes totally agree with the folder + multiple files approach!

Makes refactoring/deleting things easier.

I just read this article that suggested the same thing:https://programmingisterrible.com/post/139222674273/how-to-write-disposable-code-in-large-systems

I was thinking it would be a good discussion point for a tech weekly or similar actually!

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
styrelius: Yep, this will soon be gone.

But when comparing `!!` and explicit type coercion with `Boolean()` I think the latter is more clear.

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
styrelius: Fixed it now!

robinandeer (approved):
I like it - great job!

PR #570, opened 2021-07-08
Override style reset for Embark content

robinandeer:
## What?

Increase CSS specificity to make sure link padding takes precedence.

## Why?

Fix this issue:

Screenshot 2021-07-08 at 10 12 06

We are applying some general styles to all descendant elements inside Embark that overrides component-specific styling.

```js
// src/client/pages/Embark/index.tsx

const EmbarkStyling = styled.div`
  * {
    padding: 0;
  }
`
```

It seems like this doesn't cause issues consistently but rather based on e.g. order of how styles are injected into the page 🤔 

In either case I think we should refrain from using these semi-global styles in the future. I'm not too happy with the solution since it doesn't address the real problem but it will fix this specific issue at least.

## Demo

Screenshot 2021-07-08 at 11 03 13

Screenshot 2021-07-08 at 11 19 01

### DIFF
--- a/src/client/pages/Embark/LanguagePicker.tsx
+++ b/src/client/pages/Embark/LanguagePicker.tsx
@@ -28,9 +28,13 @@ const StyledLink = styled(Link)`
   outline: 0;
   color: ${colorsV2.black};
   text-decoration: none;
-  padding: 10px 20px;
   transition: all 250ms;
   font-size: 0.9rem;
+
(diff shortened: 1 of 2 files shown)


gustaveen (approved):
Ah good find! 🙏 Hm, not optimal to have these embark specific resets where the specificity is too high. Definitely, we should move away from this semi-global setup :D

PR #572, opened 2021-07-09
Proper T&C links again

styrelius:
## What?

Reintroduce the T&C URL's we get on the `quotes` (from back-end) for the links on `/offer` page and in checkout. Currently, due to license reasons that are no longer valid, we only use the URL's on the Danish `quotes`.

This means the temporary hard coded URL's that have been used for `'TERMS_AND_CONDITIONS'` links in Sweden and Norway are finally removed  😀

## Why?

We now get URL's for the correct T&C's documents from back-end. Or at least as correct as it gets - we still don't have English translations of these documents, hence the links on `/se-en` and `/-no-en` will lead to documents in Swedish and Norwegian respectively 🤷‍♀️ 

_Referenced ticket(s): [GRW-295]_

## Screenshot

2021-07-09-tc-link

[GRW-295]: https://hedvig.atlassian.net/browse/GRW-295


joacimastrom (approved):

robinandeer (approved):
So nice and clean ✨

gustaveen (approved):
Yaay! 🙌

PR #575, opened 2021-07-13
GRW-378: Depend on official @sentry/react package

robinandeer:
## What?

- Upgrade to the latest official Sentry NPM dependencies.

- Remove existing sentry setup (CDN-based script).

- Add dotenv setup in webpack config. Bundle sentry env variables.

## Why?

We want to setup Sentry in the standard way. As far as possible, we want to rely on the automated exception logging that Sentry provides.


gustaveen (approved):
Great job setting this up! 🙌

gustaveen (approved):
Great job setting this up! 🙌

PR #576, opened 2021-07-13
Point VSCode to local TypeScript lib

robinandeer:
## What?

Tell VSCode to use the local installation of TS.

## Why?

@styrelius and I have run into issues with types etc. since VSCode doesn't reliably use the local TS lib installation but falls back to a more recent version of TS.

This seems pretty safe to include as a default for whoever is developing this project, don't you think?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+  "typescript.tsdk": "node_modules/typescript/lib"
+}


gustaveen (approved):
Aha, I thought it always defaulted to the version in the workspace but doesn't hurt to be explicit 🙌

robinandeer:
Yes as long as it doesn't cause conflicts with our environments!

PR #578, opened 2021-07-13
GRW-377 Add link component that will determine when to open a link in a new tab

gustaveen:
## What?

Add link component that will determine when to open a link in a new tab

## Why?

When using the web-onboarding in the apps, in a web view, you can't open links with `target=_blank`, in other words, links that would open in a new tab. So when `android` or `ios` variation is active, render a link without `target=_blank`

Things to consider: 

- What should we name the component? 
- Should we rather use the `target=_blank` prop for the component instead of adding a custom one `external`? Might actually prefer that since we would keep the same API as for a regular link.

### Demo
https://user-images.githubusercontent.com/6661511/125460271-f82f60a2-18ea-4479-9669-0f418e7a6223.mov

### DIFF
--- a/src/client/components/RawLink.tsx
+++ b/src/client/components/RawLink.tsx
@@ -3,20 +3,20 @@ import { useVariation, Variation } from 'utils/hooks/useVariation'
 
 type RawLinkProps = {
   className?: string
-  external?: boolean
+  openInNewTab?: boolean
   href: string
 }
 
 export const RawLink: React.FC<RawLinkProps> = ({
   className,
   children,
-  external,
+  openInNewTab,
   href,
 }) => {
   const variation = useVariation()
-  const externalCondition =
-    external && ![Variation.IOS, Variation.ANDROID].includes(variation!)
-  const linkAttributes = externalCondition
+  const linkCondition =
+    openInNewTab && ![Variation.IOS, Variation.ANDROID].includes(variation!)
+  const linkAttributes = linkCondition
     ? { target: '_blank', rel: 'noreferrer noopener' }
     : {}
 
--- a/src/client/components/RawLink.tsx
+++ b/src/client/components/RawLink.tsx
@@ -1,0 +1,28 @@
+import React from 'react'
+import { useVariation, Variation } from 'utils/hooks/useVariation'
+
+type RawLinkProps = {
+  className?: string
+  external?: boolean
+  href: string
+}
+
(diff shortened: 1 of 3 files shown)


src/client/pages/OfferNew/Perils/InsuranceValues/index.tsx
```diff
@@ -88,14 +89,14 @@ export const InsuranceValues: React.FC<Props> = ({ offerQuote }) => {
         {[...offerQuote.insuranceTerms.entries()].map(
           ([insuranceTermType, insuranceTerm]) => {
             return (
-              <Link
+              <TermsLink
                 key={insuranceTermType}
                 href={insuranceTerm.url}
-                target="_blank"
+                external
```
gustaveen, line 95: Maybe use target instead for a more familiar API 
```suggestion
                target="_blank"
```

src/client/components/RawLink.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export const RawLink: React.FC<RawLinkProps> = ({
+  className,
+  children,
+  external,
+  href,
+}) => {
+  const variation = useVariation()
+  const externalCondition =
+    external && ![Variation.IOS, Variation.ANDROID].includes(variation!)
```
robinandeer, line 18: Why do we need the non-null assertion here for `variation`? Always makes me worried 😅

robinandeer (dismissed):
Nice work! I agree that it's a good idea to centralise this kind of logic

I guess it's okey to do it in steps though, starting with these few links.

src/client/pages/OfferNew/Perils/InsuranceValues/index.tsx
robinandeer: Yeah I like that better!

Another approach would be to determine it directly form the URL (starts with http...) but perhaps it's better to be explicit.

src/client/components/RawLink.tsx
gustaveen: Hehe, yeah not the best feeling :D

src/client/pages/OfferNew/Perils/InsuranceValues/index.tsx
gustaveen: Cool, let's do that! I think we shouldn't default all external links to have `target="_blank"` and only use it for edge cases.

PR #579, opened 2021-07-14
Fix/blocked tracking pixels

gustaveen:
## What?

Get rid of some of the blocked tracking pixels

## Why?

- For the blocked Linkedin pixel I couldn't really see anything weird. Added protocol to see if it does the trick.
- For google tracking pixels it's a bit of a mess since it would be a mess to implement all of the top level domains for goole and we can't have a wildcard for it. So for now I added .dk so we at least allow it for all markets we're live in.


robinandeer (approved):
Looks good for me :P

PR #580, opened 2021-07-16
GRW-371 Update brand repo

gustaveen:
## What?

- Adjust components for new brand colors. The main color, `purple500`, should be used for buttons etc with black text.
- Add Tooltip to Storybook.
- Set tooltip color to black, since we should never use white text on purple.
- Set text links to `purple900` for better contrast. This is the one that might be a bit low contrast but I think we're good with using this for now and then maybe could ask design to come up with a slightly darker swatch.

## Why?

Align with new brand guidelines https://www.notion.so/hedviginsurance/Brand-Colors-4c36fa0fddb0495bb332fd77abcf13a0

## Demo
Screenshot 2021-07-16 at 11 28 09
Screenshot 2021-07-16 at 15 42 54
Screenshot 2021-07-16 at 15 43 11
Screenshot 2021-07-16 at 15 56 51

Screenshot 2021-07-16 at 16 09 26
Screenshot 2021-07-16 at 16 09 43


src/client/pages/ConnectPayment/components/TrustlyFailPage.tsx
```diff
@@ -1,22 +1,22 @@
...
-  flexFlow: 'column',
-})
+const Wrapper = styled.div`
+  width: 100%;
+  min-height: 100vh;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  flex-flow: column;
+`
```
robinandeer, line 15: Nice!

robinandeer (dismissed):
Great update :-)

styrelius (commented):
Very pretty! 💅 My only concern is the discounted price in the CTA box/checkout, and the text links on light backgrounds - these are pretty bad in contrast and the small fonts are really hard to read even if you have OK/good vision. And since this is the darkest of the purple tints we have I guess we have to talk to our designer/brand friends and see if they can give us another one? Or we have to replace the instances where we use purple for text with black/grey?

gustaveen:
> Very pretty! 💅 My only concern is the discounted price in the CTA box/checkout, and the text links on light backgrounds - these are pretty bad in contrast and the small fonts are really hard to read even if you have OK/good vision. And since this is the darkest of the purple tints we have I guess we have to talk to our designer/brand friends and see if they can give us another one? Or we have to replace the instances where we use purple for text with black/grey?

🙌 I agree with that. The reason I went with this is that it's marginally worse than the color we had before if you look at the contrast values 😄 So maybe we could live with that small difference until we have a new swatch or WDYT? But I will check with design, if we want to have purple text on white background we should have a darker swatch as well 

**Old color:**
Screenshot 2021-08-11 at 13 52 45

**Updated color:**
Screenshot 2021-08-11 at 13 52 05

styrelius:
> 🙌 I agree with that. The reason I went with this is that it's marginally worse than the color we had before if you look at the contrast values 😄 So maybe we could live with that small difference until we have a new swatch or WDYT? But I will check with design, if we want to have purple text on white background we should have a darker swatch for sure

Hm, it still looks to me like the difference in contrast is a lot bigger but it might be we just passed some sort of bad-contrast-threshold with this new tint 🙈 
I'm thinking if we should just change the instances of purple text on white background to be dark grey or something until we have a darker purple? Or do you think it's better to just go with this purple for the time being?

styrelius:
@gustaveen What should we do about this? What do you think about replacing purple text with grey/black?

gustaveen:
@styrelius UPDATE: New purple tint from brand repo, let's get this out 😄
Screenshot 2021-09-06 at 11 35 40

styrelius (approved):
🥳 🎨

PR #582, opened 2021-07-19
Stop logging GraphQL errors to Sentry

robinandeer:
## What?

Remove handler that logs all GraphQL errors to Sentry.

## Why?

Most (if not all) errors are handled in the client code and doesn't need to get logged to Sentry. 

Instead we should follow up on these errors either in the micro services or Apollo Studio which has a lot more information about why the error happens.


styrelius (approved):
👍 👍

PR #583, opened 2021-07-19
Upload source maps to Sentry from webpack

robinandeer:
## What?

Add sentry webpack plugin.
Remove thread loader (doesn't work with sentry webpack plugin).
Conditionally include Sentry webpack plugin if sentry auth token is defined.

## Why?

We want more informative exception logging in Sentry.

Webpack thread-loader isn't compatible with the Sentry webpack plugin.
Ref: https://github.com/getsentry/sentry-webpack-plugin/issues/272

_Referenced ticket(s): [GRW-379]_

[GRW-379]: https://hedvig.atlassian.net/browse/GRW-379


webpack/webpack.config.server.js
```diff
@@ -52,5 +52,16 @@ module.exports = webpackConfig({
         {},
       ),
     }),
+    ...(process.env.SENTRY_AUTH_TOKEN
+      ? [
+          new SentryWebpackPlugin({
+            release: process.env.HEROKU_SLUG_COMMIT,
```
styrelius, line 58: Is this env variable `HEROKU_SLUG_COMMIT` needed to test the integration locally? Do we need to able to test it locally after we've merged? And if so, should `HEROKU_SLUG_COMMIT` be added to `.env.example`? 🤔 

And maybe the same questions goes for the `SENTRY_AUTH_TOKEN` env variable?

styrelius (commented):
Nice! 🙌 Just some questions about the env variables

webpack/webpack.config.server.js
robinandeer: Ahh yeah I could add those there and comment that they are optional :-)

styrelius: Awesome!! 🙌 😀

styrelius (dismissed):
🗺️ 🤸‍♀️ 🤸‍♀️ 🤸‍♀️

.env.example
```diff
@@ -12,3 +12,7 @@ ADYEN_ORIGIN_KEY=pub.v2.something-kinda-secret
 ADYEN_ENVIRONMENT=<test|live>
 
 APP_ENVIRONMENT=<development|staging|production>
+
+# Optional - if you want to test uploading source maps to Sentry
+HEROKU_SLUG_COMMIT=<commit hash>
+SENTRY_AUTH_TOKEN=<sentry auth token>
```
styrelius, line 18: Maybe the `SENTRY_AUTH_TOKEN` should also be added to this file?

robinandeer: What do you mean? It's in the file, no?

styrelius: Hahaha what?? I'm blind apparently? That's kind of scary actually

robinandeer:
Just had ro rebase to solve the conflict in `yarn.lock`

styrelius (approved):

PR #585, opened 2021-07-28
Migrate from Enzyme to React Testing Library

robinandeer:
## What?

Switch to using XML based snapshots.
Remove Enzyme setup scripts.
Extend expect with JSDOM assertions.
Refactor test utils to use built-in async helpers.
Remove tests that focus on implementation details (specific class names etc.)

## Why?

Enzyme is no longer the recommended library to test React components.
React Testing Library is focused on testing components based on how users are actually interfacing with your app. This pushes us to write tests that stay away from implementation details. This makes tests more robust.

Ref: https://testing-library.com/docs/react-testing-library/migrate-from-enzyme/


src/client/utils/textKeys.test.tsx
```diff
@@ -1,8 +1,25 @@
-import { mount } from 'enzyme'
 import React from 'react'
-import { act } from 'react-dom/test-utils'
-import { nextTickAsync } from './misc'
-import { makeTextKeyResolver, TextKeyProvider, useTextKeys } from './textKeys'
+import { renderHook } from '@testing-library/react-hooks'
```
styrelius, line 2: Why not use the custom `renderHook` function from the `/test/utils.tsx` file here?

src/client/utils/StorageContainer.test.tsx
```diff
@@ -1,27 +1,29 @@
-import { mount } from 'enzyme'
 import React from 'react'
+import userEvent from '@testing-library/user-event'
```
styrelius, line 2: Since we export all things `'@testing-library/react'` from the `/test/utils.tsx` file shouldn't `userEvent` rather be imported from there instead? Or have I completely misunderstood? 🙈

styrelius (commented):
Really great job with this!! 🥇 👑 🎉  Sorry about my thousand comments, it's mostly things that I don't understand and want to learn 🙃

src/client/containers/SessionTokenGuard.test.tsx
```diff
@@ -13,48 +12,51 @@ jest.mock('../apolloClient', () => ({
...
+        <SessionTokenGuard>
+          <div />
+        </SessionTokenGuard>
+      </Provider>
+    </MockedProvider>,
+    { wrapperProps: { location: { pathname: '/se/new-member/blah' } } },
   )
 
-  expect(wrapper.find(Redirect).prop('to')).toBe('/se/new-member')
+  expect(getByText('Redirected to /se/new-member')).toBeInTheDocument()
```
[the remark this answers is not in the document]
robinandeer: This comes from the `react-router-dom` mock that I setup earlier in the module.

The mock just makes it so that some text is printed when a `<Redirect />` component is rendered. I think it would otherwise be hard to test that such a component is rendered when expected.

I agree that this is kind of testing implementation details but it does cover the current use. Perhaps it's okey for this PR and then we can continue improving it!

src/client/pages/Embark/LanguagePicker.test.tsx
```diff
@@ -1,84 +1,63 @@
...
 import { MemoryRouter } from 'react-router-dom'
+import userEvent from '@testing-library/user-event'
+import { renderComponent, waitFor } from 'test/utils'
 import { LanguagePicker } from './LanguagePicker'
 
 jest.mock('../../apolloClient', () => ({}))
 
-describe('Swedish language picker renders without 💥', () => {
-  const wrapper = mount(
+test('Swedish language picker renders without 💥', () => {
```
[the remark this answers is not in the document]
robinandeer: I was inspired by this article: https://kentcdodds.com/blog/avoid-nesting-when-youre-testing

Perhaps we can break out some of the expect statements though, I'm not sure if that would make it more clear.

src/client/test/utils.tsx
```diff
@@ -0,0 +1,34 @@
...
+    ),
+    ...options,
+  })
+
+const customRenderHook = (
+  callback: (props: any) => any,
+  options?: Omit<RenderHookOptions<AllProvidersProps>, 'wrapper'>,
+) => renderHook(callback, { wrapper: AllProviders, ...options })
+
+export * from '@testing-library/react'
```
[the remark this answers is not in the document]
robinandeer: I actually haven't given it much thought. I don't really love this approach actually. However, I think there's a point to setting up things following the official guides. That's why I ended up with this approach.

src/client/utils/ExchangeTokenRetrieval.test.tsx
```diff
@@ -10,44 +10,44 @@ jest.useFakeTimers()
...
+import { render, act, waitFor } from '@testing-library/react'
 import { ExchangeTokenDocument } from 'data/graphql'
-import { nextTickAsync } from 'utils/misc'
 import { MockStorage } from 'utils/storage/MockStorage'
 import { StorageContext } from 'utils/StorageContainer'
 import possibleTypes from '../../../possibleGraphqlTypes.json'
 import { createSession, Session } from '../../shared/sessionStorage'
 import { ExchangeTokenRetrieval } from './ExchangeTokenRetrieval'
 
 it("moves to taking too long when location doesn't have exchange-token", async () => {
```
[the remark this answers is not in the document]
robinandeer: haha it's actually supposed to be double quotes since it includes a single quote: `doesn't`

src/client/test/utils.tsx
```diff
@@ -0,0 +1,34 @@
...
+}
+
+const AllProviders: React.FC<AllProvidersProps> = ({ children, location }) => {
+  return <StaticRouter location={location}>{children}</StaticRouter>
+}
+
+const customRender = (
+  ui: React.ReactElement,
+  options?: Omit<RenderOptions, 'wrapper'> & {
+    wrapperProps?: Record<string, any>
```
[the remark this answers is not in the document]
robinandeer: Yes, that's a good point -- I will type those out. Not sure why I left it like that 😅

src/client/test/utils.tsx
```diff
@@ -0,0 +1,34 @@
...
+  },
+) =>
+  render(ui, {
+    wrapper: (props: any) => (
+      <AllProviders {...props} {...options?.wrapperProps} />
```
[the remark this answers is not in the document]
robinandeer: Right, so the `wrapperProps` are provided by the user when calling `customRender`. The other `props` are provided by testing-library itself when it eventually calls the `wrapper` function. It will pass things like `children`. That's why we need both!

Code reference here: https://github.com/testing-library/react-testing-library/issues/780

src/client/utils/hooks/useVariation.test.tsx
```diff
@@ -1,44 +1,30 @@
...
 import { useVariation } from 'utils/hooks/useVariation'
-
-const VariationShower: React.FC = () => {
-  const variation = useVariation()
-
-  return <>{variation}</>
-}
+import { renderHook } from 'test/utils'
 
 it('returns null on no variation', () => {
```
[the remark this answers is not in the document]
robinandeer: I agree with you - will update

src/client/utils/textKeys.test.tsx
robinandeer: It's just because we needed a different provider to wrap these tests. I wasn't sure if it's worth it to include the `TextKeyProvider` in the main `customRenderHook` function.

Just to not create abstractions too soon.

src/client/utils/StorageContainer.test.tsx
robinandeer: Hmm should we? It comes from a different package though. Not `@testing-library/react` I mean.

Like I wrote above, I don't love that convention but it's what is recommended in the official guide :P

src/client/pages/Embark/LanguagePicker.test.tsx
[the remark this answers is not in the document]
styrelius: Hm, I understand the point but I'm not sure I agree it really applies here - the nesting wasn't _that_ bad and I wouldn't say there were really unnecessary abstractions either 🙈 But I'm very pro following some sort of principle/guideline so maybe this is better!

src/client/utils/ExchangeTokenRetrieval.test.tsx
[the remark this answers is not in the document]
styrelius: Aah, yes of course! Missed that one 🤦‍♀️

src/client/test/utils.tsx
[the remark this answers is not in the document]
styrelius: OK yes, that makes sense 👍

src/client/test/utils.tsx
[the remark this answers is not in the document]
styrelius: 🆒 👍

src/client/utils/StorageContainer.test.tsx
styrelius: Oh, sorry I just read it as `'@testing-library/react'`, never mind! 😳 Or should we export everything related to testing packages from that `utils` file just to be consistent? Or what's least confusing?

src/client/utils/textKeys.test.tsx
styrelius: I see! In a way I think it would be nice to always use the same `renderHook` function and have all the provider thingies there (to make it more consistent I guess, and be clear about that you're generally not supposed to import the `renderHook` directly from `'@testing-library/react-hooks'` and add the `StaticRouter` etc.), but maybe that's not a very AHA way to think 🙃

robinandeer: I'll have a look into it and see if including `StaticRouter` across all tests will cause any issues!

src/client/utils/StorageContainer.test.tsx
robinandeer: This is most explicit so that's why it wins for me 🙃

src/client/utils/textKeys.test.tsx
robinandeer: So the `TextKeyProvider` actually performs state updates in a `useEffect` -- so I think `StaticTextKeyProvider` is a better option for tests - I added that to the providers.

To test this particular logic, I've updated the test utils for that you can override the wrapper component on a per-test basis! What do you think of that solution?

src/client/utils/StorageContainer.test.tsx
styrelius: It's good, let's keep it like this! 👍

src/client/utils/textKeys.test.tsx
styrelius: I like it! 👍 😀

src/client/pages/ConnectPayment/components/TrustlyCheckout.test.tsx
```diff
@@ -33,46 +32,42 @@ it('opens trustly modal and renders correct iframe url', async () => {
...
-      >
-        <StaticTextKeyProvider>
-          <TrustlyCheckout />
-        </StaticTextKeyProvider>
-      </MockedProvider>
-    </StaticRouter>,
+        },
+      ]}
+    >
+      <StaticTextKeyProvider>
```
styrelius, line 60: So maybe the `StaticTextKeyProvider` is redundant here now that the `StaticTextKeyProvider` is included in the `renderComponent` function? Same on line 65 in `TrustlyModal.test.tsx`. Or am I missing something?

robinandeer: Good catch! Removed it :)

styrelius (approved):
🥳 🌤️ 🌴

PR #586, opened 2021-07-29
Add explicit width and height to img tags

robinandeer:
## What and Why?

Help browsers figure out dimensions before images are fully downloaded. Avoid layout shifts.
Add empty alt tags to signify decorative img elements for screen readers.

No visual updates expected.

_Referenced ticket(s): [TSEO-21]_

### Peril collection

Screenshot 2021-07-29 at 10 32 54

### Peril collection - desktop
Screenshot 2021-07-29 at 10 33 00

### Peril modal
Screenshot 2021-07-29 at 10 33 12

### Connect payment screen
Screenshot 2021-07-29 at 10 37 39

### Download screen
Screenshot 2021-07-29 at 10 59 10

### Download screen - desktop
Screenshot 2021-07-29 at 10 59 18

[TSEO-21]: https://hedvig.atlassian.net/browse/TSEO-21


src/client/pages/Download/components/AppImage.tsx
```diff
@@ -6,33 +6,50 @@ type Props = {
   width: number
 }
 
-const Image = styled.img<Props>`
-  width: ${({ width }) => width}px;
+const Image = styled.img<{ imageWidth: number }>`
+  width: ${({ imageWidth }) => imageWidth}px;
```
styrelius, line 10: Question since I truly don't know this: will setting the dimensions directly as attributes on the `img` have any effect on layout shift (or anything else) when the dimensions are also, already, set with css like this?

src/client/pages/OfferNew/Perils/PerilCollection/PerilItem.tsx
```diff
@@ -113,7 +113,7 @@ export const PerilItem: React.FC<PerilItemProps> = ({
   return (
     <OuterContainer>
       <Container onClick={onClick}>
-        {iconUrl && <PerilIcon src={iconUrl} />}
+        {iconUrl && <PerilIcon src={iconUrl} alt="" width={24} height={24} />}
```
styrelius, line 116: Since these icons are set to be `3 rem` in `width` and `height` on desktop with CSS won't this just be confusing to the poor browser and everyone else? Also, I don't really understand why we use `img` elements for the SVG's but that's definitely not something you decided 🙃

src/client/pages/Download/components/AppImage.tsx
```diff
@@ -6,33 +6,50 @@ type Props = {
...
-      return `${basePath}/hedvig-app-no.png`
+      return {
+        src: `${basePath}/hedvig-app-no.png`,
+        width: 1660,
+        height: 1400,
+      }
     case 'dk':
-      return `${basePath}/hedvig-connect-payment-dk.png`
+      return {
+        src: `${basePath}/hedvig-connect-payment-dk.png`,
```
styrelius, line 33: Maybe since we're in here anyway we should take the opportunity to change names of this and `hedvig-connect-payment-dk-en.png` to `hedvig-app-dk` and `hedvig-app-dk-en`..?

styrelius (commented):
🖼️

src/client/pages/ConnectPayment/sections/ConnectPayment.tsx
```diff
@@ -151,6 +151,9 @@ export const ConnectPaymentPage: React.FC = () => {
             src={
               '/new-member-assets/connect-payment/connect-payments-illustration-mono.svg'
             }
+            alt=""
+            width={637}
+            height={600}
```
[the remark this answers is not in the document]
robinandeer: Nice catch! I forgot to add `height: auto;` :)

Now it works:

Screenshot 2021-08-03 at 16 24 42

Screenshot 2021-08-03 at 16 24 47

src/client/pages/Download/components/AppImage.tsx
robinandeer: We are only setting the width of the image so there's not enough information to calculate the aspect ratio.

I think it's necessary to set width/height directly on the img element to enable automatic calculation of `aspect-ratio` - see here: https://web.dev/optimize-cls/

src/client/pages/OfferNew/Perils/PerilCollection/PerilItem.tsx
robinandeer: Yeah I think in this case it will not make a difference. Perhaps the only reason is to be consistent and always add height/width to img-elements like Google recommends.

Agree that we can rework how we deal with SVG icons!

src/client/pages/ConnectPayment/sections/ConnectPayment.tsx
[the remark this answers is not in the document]
styrelius: Awesome!! 👑

src/client/pages/OfferNew/Perils/PerilCollection/PerilItem.tsx
styrelius: Let's keep it like this "for now" then 😄 If there won't be an issue that the browser calculates the image size incorrectly here

src/client/pages/Download/components/AppImage.tsx
styrelius: That's very true 🧠  What I guess I'm worried about then is that we're saying that these images will be **x** size but then when the CSS does its thing the size is actually **y** - aren't we still at the risk of causing a layout shift then?

robinandeer: CSS is evaluated before anything is render on screen though (https://css-tricks.com/browser-painting-and-considerations-for-web-performance/).

So the browser will have both the HTML and CSS information available. The only thing is doesn't know on first render/paint is the image data which is not yet loaded.

So browser will use both HTML and CSS to calculate size of the image and avoid layout shift. At least I'm fairly sure this is how it works.

styrelius: Ah, OK cool! Thank you for your explanation (and patience) 👍

styrelius (approved):
👏 👏 👏

PR #588, opened 2021-08-02
Disallow indexing of forever pages

robinandeer:
## What?

Block crawlers by adding noindex meta tag.

## Why?

We don't want people to be able to google for discount codes.
Since this is a single page, it's better use a meta tag instead of including it as part of the robots.txt.

https://developers.google.com/search/docs/advanced/crawling/block-indexing


styrelius (approved):
🤖 🚫 👌

PR #592, opened 2021-08-11
Add CODEOWNERS file

robinandeer:
## What?

Add CODEOWNERS file.

## Why?

Automatically request review when someone opens a pull request.

Ref: https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/about-code-owners#about-code-owners

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+*       @HedvigInsurance/www


styrelius (approved):
Nice! 😀

robinandeer:
Or should it be @HedvigInsurance/clients instead? Is it too narrow with only "www"?

styrelius:
> Or should it be @HedvigInsurance/clients instead? Is it too narrow with only "www"?

No I think **WWW** is enough - no need _always_ ask the app team for review on the web-onboarding. I think. (But we should probably be better at requesting reviews from them when we do something that potentially affects them though)

robinandeer:
Right, that's true. Especially until they migrate to their own Embark-clients

PR #593, opened 2021-08-12
Integrate datadog buildpack

robinandeer:
## What?

Integrate Datadog Buildpack with Heroku.

I've updated Heroku staging + prod with the needed setup and environment variables already.

## Why?

Collect all logs from services in a single place.


styrelius (approved):
🦮

PR #595, opened 2021-08-16
Fix / Quote title/headline/name text keys

styrelius:
## What?

- `displayName` is added to `QuoteBundle` query and used instead of `insuranceTypeTextKeys` properties in `PerilRow` and  for the split start dates in `StartDate` components.
- New text keys for Quote Titles are added, to be used anywhere the type of insurance/quote is written out, like e.g. in the CTA box/`Sidebar` component.  These are accessed as before through the `getInsuranceTitle` function in `/OfferNew/utils.ts`.
- `SIDEBAR_INSURANCE_LABEL_SE` is removed from both `CheckoutContent` and `Sidebar` components since the type of insurance is now included in all Quote Title text keys.
- Generic `h3` tag in `PerilRow` component is replaced with `InsuranceTypeHeadline` styled component so we don't just use default browser styling.

## Why?

- Current text keys for the quote/insurance type/title aren't consistent with what we call the insurance products on the landing page.
- They also don't specify if the quote is regarding a Youth/Student insurance contract, which is good to make really clear (although that information can of course be found in the checkout summary as well).
- It makes more sense to use the `displayName` that we get form back-end instead of doing another mapping between Lokalise text keys and type of contract, for the places where we need a label for a specific insurance (like in the perils section)

_Referenced ticket(s): [GRW-324]_

## Screenshots

2021-08-13-new-quote-titles-no

2021-08-16-new-quote-titles-dk

2021-08-16-new-quote-titles-se-student

[GRW-324]: https://hedvig.atlassian.net/browse/GRW-324


gustaveen:
GJ! 🙌 Great with some better consistency between titles 👌
We could also remove `SIDEBAR_INSURANCE_TYPE_*` textkeys (in `InsuranceSummaryDetails.tsx`) and replace them with `CHECKOUT_DETAILS_RESIDENCE_TYPE_*` right? :)

styrelius:
> We could also remove `SIDEBAR_INSURANCE_TYPE_*` textkeys (in `InsuranceSummaryDetails.tsx`) and replace them with `CHECKOUT_DETAILS_RESIDENCE_TYPE_*` right? :)

Hm, I'm not following 🤔 Can only find the `SIDEBAR_INSURANCE_TYPE_*` text keys in `/Sidebar/DetailsModal/utils.ts`. And we should probably replace them there, yes, but maybe in another PR? Since we should probably rename them then, and not use the word `CHECKOUT`.

gustaveen:
Haha sorry, wrong file, yes it was the details modal file :D But yes, let's do that in another one 👍

gustaveen (approved):
⭐️

PR #596, opened 2021-08-20
Split up tracking

gustaveen:
## What?

Split up different tracking services we use 

## Why?

Easier to edit and delete tracking in the future

### DIFF
--- a/src/client/utils/tracking/adTraction.ts
+++ b/src/client/utils/tracking/adTraction.ts
@@ -1,0 +1,93 @@
+import md5 from 'md5'
+import { TypeOfContract } from 'src/client/data/graphql'
+import { OfferData } from 'src/client/pages/OfferNew/types'
+import {
+  isBundle,
+  isNorwegian,
+  isYouth,
+  isDanishAccidentBundle,
+  isStudentOffer,
+  isDanishTravelBundle,
+} from 'src/client/pages/OfferNew/utils'
+import { NoComboTypes, DkBundleTypes } from './tracking'
+
+type TypeOfContractExcludedUnused = Exclude<
+  TypeOfContract | NoComboTypes | DkBundleTypes,
+  'DK_ACCIDENT' | 'DK_ACCIDENT_STUDENT' | 'DK_TRAVEL' | 'DK_TRAVEL_STUDENT'
+>
+
+const adtractionProductMap: Record<TypeOfContractExcludedUnused, number> = {
+  SE_HOUSE: 1477448913,
+  SE_APARTMENT_BRF: 1417356498,
+  SE_APARTMENT_STUDENT_BRF: 1423041022,
+  SE_APARTMENT_RENT: 1417356528,
+  SE_APARTMENT_STUDENT_RENT: 1412601108,
+  NO_HOME_CONTENT_OWN: 1492623645,
+  NO_HOME_CONTENT_RENT: 1492623645,
+  NO_HOME_CONTENT_YOUTH_OWN: 1492623719,
+  NO_HOME_CONTENT_YOUTH_RENT: 1492623719,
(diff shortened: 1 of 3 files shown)


styrelius (approved):
👌

src/client/utils/tracking/adTraction.ts
```diff
@@ -0,0 +1,93 @@
...
+  isDanishAccidentBundle,
+  isStudentOffer,
+  isDanishTravelBundle,
+} from 'pages/OfferNew/utils'
+import { NoComboTypes, DkBundleTypes } from './tracking'
+
+type TypeOfContractExcludedUnused = Exclude<
+  TypeOfContract | NoComboTypes | DkBundleTypes,
+  'DK_ACCIDENT' | 'DK_ACCIDENT_STUDENT' | 'DK_TRAVEL' | 'DK_TRAVEL_STUDENT'
+>
```
robinandeer, line 17: Is this a place where a comment "why" these are unused should be added?

That's a bit unclear at least to me reading this.

src/client/utils/tracking/adTraction.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const adtractionProductMap: Record<TypeOfContractExcludedUnused, number> = {
```
robinandeer, line 19: Is product the same as contract in this case? Should we agree on using only one term?

src/client/utils/tracking/adTraction.ts
```diff
@@ -0,0 +1,93 @@
...
+
+    if (isDanishTravelBundle(offerData)) {
+      return isStudentOffer(offerData)
+        ? adtractionProductMap[DkBundleTypes.DkTravelBundleStudent]
+        : adtractionProductMap[DkBundleTypes.DkTravelBundle]
+    }
+  }
+  return adtractionProductMap[
+    offerData.quotes[0].contractType as TypeOfContractExcludedUnused
+  ]
```
robinandeer, line 65: I know we do this in other place but it's also something I react to when reading. So perhaps a comment why taking the contractType from the "first item" in the list makes sense, no?

src/client/utils/tracking/adTraction.ts
```diff
@@ -0,0 +1,93 @@
...
+}
+
+export const adtraction = (
+  orderValue: number,
+  orderId: string,
+  emailAddress: string,
+  couponCode: string | null,
+  offerData: OfferData,
+) => {
```
robinandeer, line 74: Agree with what @styrelius usually says that so many positional arguments is less clear than accepting a single object like:

```suggestion
export const adtraction = ({
  orderValue,
  orderId,
  emailAddress,
  couponCode,
  offerData,
}: {
  orderValue: number,
  orderId: string,
  emailAddress: string,
  couponCode: string | null,
  offerData: OfferData,
}) => {
```

src/client/utils/tracking/tracking.ts
```diff
@@ -18,6 +16,8 @@ import {
   isStudentOffer,
 } from 'pages/OfferNew/utils'
 import { trackOfferGTM } from './gtm'
+import { adtraction } from './adTraction'
```
robinandeer, line 19: Unify how we spell this word. Either camelcase or all lowercase :)

robinandeer (commented):
Nice updates, just had some improvement suggestions

src/client/utils/tracking/adTraction.ts
gustaveen: Good point 🙌

src/client/utils/tracking/adTraction.ts
styrelius: We have a `getMainQuote` function in `/OfferNew/utils.ts` - maybe that could be used?

PR #597, opened 2021-08-23
Fix / Quote title translations for House Insurance

styrelius:
## What?

- Translations of `QUOTE_TITLE_HOUSE` are updated, both in Swedish and English, to match the names/titles on the market-web.
- `download-translations` script is updated to include line breaks from Lokalise (this means that if a line break is added to a translation through the Lokalise UI this will be added as a newline character `\n` in the translation JSON files, which will enable adding line breaks to our UI with CSS).
- `white-space: pre-line;` rule is added to the headlines in `CheckoutContent` and `Sidebar` components in order for the the headlines to wrap on line breaks like newline character
- Smaller font on the peril collection insurance type headline

## Why?

- Consistency in our communication
- It's nice to be able to include line breaks in translation strings since that gives us more control over white space wrapping.

## Screenshots

Screenshot 2021-08-23 at 15 12 26

Screenshot 2021-08-23 at 15 17 21

Screenshot 2021-08-23 at 15 19 28

<hr />

## Edit

- Font size of peril section headline is changed back to it's current size since no-one liked the change to the smaller font 🙃 

2021-08-25-resetted-peril-section-headline


src/translations/nb_NO.json
```diff
@@ -123,7 +123,7 @@
   "SIDEBAR_REQUEST_CANCELLATION_GENERIC_INSURANCE": "Kanseller min gamle forsikring",
   "RETRY": "Prøv igjen",
   "CONNECT_PAYMENT_DIRECT_ERROR_TAKING_TOO_LONG": "Det tok for lang tid å laste siden, vennligst prøv igjen.",
-  "CONNECT_PAYMENT_DIRECT_ERROR_INVALID_TOKEN": "Koblingen du skrev inn er en ugyldig lenke.\\n",
+  "CONNECT_PAYMENT_DIRECT_ERROR_INVALID_TOKEN": "Koblingen du skrev inn er en ugyldig lenke.\n",
```
gustaveen, line 126: Not related to the PR itself, but just curious why there's a line break at the end of the sentence?

src/client/pages/OfferNew/Perils/PerilRow.tsx
```diff
@@ -4,15 +4,10 @@ import { InsuranceValues } from 'pages/OfferNew/Perils/InsuranceValues'
...
 
 const InsuranceTypeHeadline = styled.div`
-  padding: 16px 0;
-  font-size: 1.375rem;
-
-  ${LARGE_SCREEN_MEDIA_QUERY} {
-    font-size: 1.5rem;
-  }
+  padding: 1rem 0;
+  font-size: 1rem;
```
gustaveen, line 10: Should we keep some hierarchy between the title of the section and the title of the perils? It feels a bit small, in desktop, maybe have it as 1.25rem? :)

src/translations/nb_NO.json
styrelius: I have no idea 😄 But apparently it's been there for a while (only it's been written out as `\n` I suppose)

src/client/pages/OfferNew/Perils/PerilRow.tsx
styrelius: Hm, I think it's nice that it's small. But I actually looked in the **Web UI Kit** file Figma and there's no font size between 16px and 24px there. But maybe we don't follow that at all? 🤔

styrelius: What I liked about it being smaller is that everything on the page is currently so big compared to the CTA-box, which looks weird IMO.

styrelius: But also, maybe it doesn't need to be a huge headline for the perils section, since it says in the CTA box what insurance we're talking about? It's more for the bundles, to differentiate between the contracts?

gustaveen: I think it's good to be clear about that in the bundles as you say. I would vote for larger titles :) But maybe check with Mange and Zak, we will be moving towards having tab navigation and large headlines in the updates to the offer page in the new designs. 
https://www.figma.com/proto/xsUGMTqfaBHgmYE8YX8O2e/Web-Accident-Insurance-PROTOTYPES?page-id=0%3A1&node-id=2%3A893&viewport=388%2C48%2C0.32&scaling=scale-down&starting-point-node-id=2%3A678&show-proto-sidebar=1&hide-ui=1

robinandeer (commented):
Nice update!

The only thing is the perils heading text is very small now. Smaller than the "more information" heading that's technically nested under it, no? It might be confusing in terms of the page hierarchy.

On the other hand, this will change when we introduce the tab layout like @gustaveen mentioned.

Screenshot 2021-08-24 at 14 46 56

src/client/pages/OfferNew/Perils/PerilRow.tsx
styrelius: OK changed it back to the larger font

gustaveen (approved):
🏡👌

PR #598, opened 2021-08-24
Track virtual page views

gustaveen:
## What?
Send page views to Google Tag Manager with custom dimensions specified by Sonny:
Screenshot 2021-08-24 at 14 33 50

The hook sends paga data to GTM on each route change. Added a new type for `GTMPageData`. It might be a bit overkill to have typed datalayer object since we can send basically anything. But we agreed to have it as this for now, if it will grow we can change it.

## Why?
To be able to remove Segment, who currently picks up the virtual page views, we need to send it ourselves. 

https://user-images.githubusercontent.com/6661511/130618481-e99626d4-9c5a-42ef-894a-1b5e46ec8bb9.mov

### DIFF
--- a/src/client/App.tsx
+++ b/src/client/App.tsx
@@ -13,10 +13,12 @@ import {
   StorageState,
   WithStorageProps,
 } from './utils/StorageContainer'
+import { usePageview } from './utils/tracking/gtm'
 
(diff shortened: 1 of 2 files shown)


robinandeer (approved):
Looks good, just consider naming the hook as an "effect"!

PR #599, opened 2021-08-24
Minor refactoring to tracking

gustaveen:
## What?
As suggested in previous PR https://github.com/HedvigInsurance/web-onboarding/pull/596


src/client/utils/tracking/adtraction.ts
```diff
@@ -8,15 +8,17 @@ import {
   isDanishAccidentBundle,
   isStudentOffer,
   isDanishTravelBundle,
+  getMainQuote,
 } from 'pages/OfferNew/utils'
 import { NoComboTypes, DkBundleTypes } from './tracking'
 
+// Exclude single contract types since they are only sold as part of bundles
```
styrelius, line 15: Miiiinor change that might make it even clearer? 🙈 
```suggestion
// Exclude single contract types that are only sold as part of bundles
```

src/client/utils/tracking/adtraction.ts
```diff
@@ -8,15 +8,17 @@ import {
...
 type TypeOfContractExcludedUnused = Exclude<
   TypeOfContract | NoComboTypes | DkBundleTypes,
   'DK_ACCIDENT' | 'DK_ACCIDENT_STUDENT' | 'DK_TRAVEL' | 'DK_TRAVEL_STUDENT'
 >
 
-const adtractionProductMap: Record<TypeOfContractExcludedUnused, number> = {
+const adtractionContractMap: Record<TypeOfContractExcludedUnused, number> = {
```
styrelius, line 21: Hm, no that I think about it - why do we call this `Map`? 🤔 Isn't it more like `adtractionContractValues` or something?

styrelius (dismissed):
Nice! Two very minor things, you can just skip them if you don't agree 😄

styrelius (approved):
👑

PR #602, opened 2021-08-31
GRW-388 / Feature / Add support for Swedish accident to debugger

robinandeer:
## What?

Add new graphql mutation for Swedish home + accident quote.
Add support for Swedish home + accident bundle in debugger interface.
Update codegen.

> I followed the same approach as for the Danish bundles. It does add some boilerplate but it pretty easy to follow. Backend will work on making generic quotes/contract types -- that would be a good point to revisit this setup and make it work without need to update every time there's a new contract type.

## Why?

We want to be able to test and develop the flow for signing up for accident insurance in SE.

_Referenced ticket(s): [GRW-388]_

Screenshot 2021-08-31 at 11 26 49

[GRW-388]: https://hedvig.atlassian.net/browse/GRW-388


src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
```diff
@@ -47,8 +47,10 @@ const Address = styled.div`
 
 export const HeroOfferDetails: React.FC<Props> = ({ offerData }) => {
   const { person, quotes } = offerData
-  const numberCoInsured = getHouseholdSize(quotes[0].quoteDetails) - 1
+  const numberCoInsured = person.householdSize - 1
```
styrelius, line 50: Nice! 🌟 I think we do this unnecessary thing in more places actually, so that's something that can be changed when one runs into that

src/client/pages/OfferNew/utils.ts
```diff
@@ -61,6 +61,17 @@ export const getOfferData = (quoteBundle: QuoteBundle): OfferData => {
   }
 }
 
+const getHouseholdSizeFromBundledQuotes = (quotes: BundledQuote[]): number => {
+  const quoteDetails = quotes.map((x) => x.quoteDetails)
```
styrelius, line 65: We usually don't use variable names like `x`, or `i` but try to generally be more explicit.
```suggestion
  const quoteDetails = quotes.map((quote) => quote.quoteDetails)
```

src/client/utils/tracking/adtraction.ts
```diff
@@ -15,7 +15,12 @@ import { NoComboTypes, DkBundleTypes } from './tracking'
 // Exclude single contract types that are only sold as part of bundles
 type TypeOfContractExcludedUnused = Exclude<
   TypeOfContract | NoComboTypes | DkBundleTypes,
-  'DK_ACCIDENT' | 'DK_ACCIDENT_STUDENT' | 'DK_TRAVEL' | 'DK_TRAVEL_STUDENT'
+  | 'DK_ACCIDENT'
+  | 'DK_ACCIDENT_STUDENT'
+  | 'DK_TRAVEL'
+  | 'DK_TRAVEL_STUDENT'
+  | TypeOfContract.SeAccident
+  | TypeOfContract.SeAccidentStudent
```
styrelius, line 23: Maybe be consistent here and either use strings for `TypeOfContract.SeAccident` and `TypeOfContract.SeAccidentStudent` as well, or replace the strings `'DK_ACCIDENT'` and so on with that syntax which explicitly references the enum 🤔

src/client/pages/OfferNew/utils.ts
```diff
@@ -93,33 +104,28 @@ const getAddressFromBundledQuotes = (
...
+    isHomeQuote(quoteDetails),
+  )
 
-  const mainQuote = isBundle(offerData)
-    ? mainQuoteInBundle[0]
-    : offerData.quotes[0]
+  const firstQuote = offerData.quotes[0]
 
-  return mainQuote
+  return (isBundle(offerData) && mainQuoteInBundle) || firstQuote
```
styrelius, line 128: Actually it might be bad to do it this way that if we have a bundle but for some reason we can't find the Home Quote we just return the first quote, since there should never ever be a case where we have a bundle without a home quote. So I'm leaning towards throwing an error or something instead if that's the case. WDYT?

Something like: 
```suggestion
const mainQuote = isBundle(offerData) ? mainQuoteInBundle : firstQuote

if (!mainQuote) {
  throw Error('There must be a main quote')
}
return mainQuote
```

or

```suggestion
if (isBundle(offerData)) {
    const mainQuoteInBundle = offerData.quotes.find(({ quoteDetails }) =>
    isHomeQuote(quoteDetails),
  )
  if (!mainQuoteInBundle) {
    throw Error('A bundle must include a home quote')
  }
  return mainQuoteInBundle
}
return offerData.quotes[0]
```

src/client/pages/OfferNew/utils.ts
```diff
@@ -61,6 +61,17 @@ export const getOfferData = (quoteBundle: QuoteBundle): OfferData => {
...
 
+const getHouseholdSizeFromBundledQuotes = (quotes: BundledQuote[]): number => {
+  const quoteDetails = quotes.map((x) => x.quoteDetails)
+
+  for (const details of quoteDetails) {
+    if ('householdSize' in details) return details.householdSize
+    if ('coInsured' in details) return details.coInsured + 1
+  }
+
+  return 0
```
styrelius, line 72: I think it might not be a great idea to return `0` as some sort of default. If we neither have `householdSize` nor `coInsured` on the `quoteDetails` something is seriously wrong and that should probably be communicated.

styrelius (commented):
I don't know who wrote what here so I don't expect answers on all of this right now 🙃

src/client/pages/OfferNew/utils.ts
mehdielket: I see your point and I agree that we should check for this case. I just see that this is not generic / scalable enough for the following reasons:
* I think that the offer page should be generic and be able display whatever combination of quotes we throw at it as long as they are valid. We are bound at some point in some market to support buying an insurance (or a bundle) that does not include home (Travel + Accident for example). 
* I think the responsibility to enforce this business rule should be (and is already) in other places such as the backend and UI for bundle selection.
Would be happy to discuss this more in person ☺️

src/client/pages/OfferNew/utils.ts
mehdielket: Same idea as the previous comment. Let's discuss this further.

src/client/pages/OfferNew/utils.ts
styrelius: Yeah let't talk about it in person 👍

styrelius: @gustaveen We'd like to know your view on this so let's talk about it all three of us.

gustaveen: Good point, it would be nice if backend could have those limitations and we will just output the results. I'm not sure that's the case right now though. Happy to talk more about it 🙌

gustaveen (approved):
GJ everyone! 🙌

PR #605, opened 2021-09-17
GRW-414 / Refactored l10n solution - part 1

styrelius:
## What?

New solution to how we work with our different locales.

- New `useCurrentLocale.ts` hook which returns an entire `locales` object, i.e. all the locale specific data needed. This makes all other functions to get locale specific data redundant, and it will hopefully make it possible to remove all checks for different locale specific values that we currently do in a lot of components.
- A type safe dynamic way to get the pattern for possible `:locale` URL parameters, which ensures they matches the keys and `path` properties of the `locales` object.

**What's left?**
- Replacing all instances of the current/old `useCurrentLocale` hook with the new one
- Removing redundant checks for locales and calls to e.g. function that returns the ISO format.
- Cleaning up (like deleting files and functions that are redundant once all replacing etc. is finished)

## Why?

- When launching in a new market there should be very clear where data that's specific to that market needs to be added, and all of that stuff should be centrailized.
- Components should not have to do a lot of checks based on strings like `'se'` or `'dk-en'` etc.

_Referenced ticket(s): [GRW-414]_

## Review app [here](https://web-onboardi-feat-l10n--gprqpl.herokuapp.com/se/new-member)

[GRW-414]: https://hedvig.atlassian.net/browse/GRW-414


mehdielket (commented):
Nice work and really good improvement!

src/client/l10n/birthDateAndSsnFormats.ts
```diff
@@ -0,0 +1,23 @@
...
+  SE: 12,
+  NO: 11,
+  DK: 10,
+}
+
+const yearMonthDayFormat = /^(19[0-9]{2}|20[0-9]{2})-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])$/
+const dayMonthYearFormat = /^(0[1-9]|[1-2][0-9]|3[0-1])-(0[1-9]|1[0-2])-(19[0-9]{2}|20[0-9]{2})$/
+
+export const birthdateFormats: Record<MarketLabel | 'default', RegExp> = {
+  default: yearMonthDayFormat,
```
robinandeer, line 19: How is default intended to be used? Don't we always know the market?

robinandeer (commented):
Really great work! There were a few things I need to look closer on to really understand -- so I'll re-review soon!

src/client/l10n/birthDateAndSsnFormats.ts
```diff
@@ -0,0 +1,23 @@
...
+export const ssnLengths: Record<MarketLabel, number> = {
+  SE: 12,
+  NO: 11,
+  DK: 10,
+}
+
+const yearMonthDayFormat = /^(19[0-9]{2}|20[0-9]{2})-(0[1-9]|1[0-2])-(0[1-9]|[1-2][0-9]|3[0-1])$/
+const dayMonthYearFormat = /^(0[1-9]|[1-2][0-9]|3[0-1])-(0[1-9]|1[0-2])-(19[0-9]{2}|20[0-9]{2})$/
+
+export const birthdateFormats: Record<MarketLabel | 'default', RegExp> = {
```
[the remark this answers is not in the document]
styrelius: Ah, yes! I've been pretty inconsistent with the spelling since I've been confused about what's actually correct. But I'll fix it so it's consistent 👍

src/client/l10n/birthDateAndSsnFormats.ts
styrelius: Really good question, don't remember since I wrote that a long time ago (only moved it to a new file now) - will look into it 🙃

src/client/l10n/localePathPattern.ts
```diff
@@ -0,0 +1,27 @@
...
+  const labelsString = permittedLocaleUrlParams.reduce(
+    (accumulated, currentLabel) => {
+      return accumulated.length
+        ? `${accumulated}|${currentLabel}`
+        : currentLabel
+    },
+    '',
+  )
+
+  return `/:locale(${labelsString})`
```
gustaveen, line 24: Could this be simplified with an array join?
```suggestion
  return `/:locale(${permittedLocaleUrlParams.join('|')})`
```

src/client/l10n/useCurrentLocale.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+const FALLBACK_LOCALE: LocaleLabel = 'se-en'
+
+const getLocaleFromPath = (path: string): LocaleLabel => {
+  const localeMatch = matchPath<LocaleUrlParam>(path, {
+    path: localePathPattern + '/*',
+  })
+
+  const localeFromPath = localeMatch?.params?.locale
```
gustaveen, line 16: Do we need to make it lowercase like we had it before or is that redundant? 
```suggestion
  const localeFromPath = localeMatch?.params?.locale?.toLowerCase()
```

src/client/l10n/useCurrentLocale.ts
```diff
@@ -0,0 +1,26 @@
...
+
+  return localeFromPath ? localeFromPath : FALLBACK_LOCALE
+}
+
+export const useCurrentLocale = () => {
+  const location = useLocation()
+  const locale = getLocaleFromPath(location.pathname)
+  const currentLocale = locales[locale]
+  return currentLocale
+}
```
gustaveen, line 26: Could we clarify what we get from this hook, maybe rename it? For me the concept of a locale is just a string, but we return an object with all settings for a specific locale. Something like `currentLocaleSettings` or `currentLocaleObject` :)

src/client/pages/OfferNew/Checkout/Checkout.stories.tsx
```diff
@@ -24,14 +24,14 @@ export default {
 
 type StoryProps = {
   initialEntries: LocationDescriptor
-  locale: Locale
+  translationsLocale: TranslationsLocale
```
gustaveen, line 27: IMO we could keep `locale` here :)

gustaveen (commented):
Awesome work Ellen! 🙌

src/client/l10n/localePathPattern.ts
styrelius: Haha YES! 🧠 🤦‍♀️

src/client/l10n/useCurrentLocale.ts
styrelius: We don't since it's typesafe now and we can know it will only be exactly the same strings as the `LocaleLabel` union type 😇

src/client/l10n/useCurrentLocale.ts
styrelius: I wanted it to be called the same as the hook we currently have since it made it a tiny bit easier to replace. And according to [Wikipedia](https://en.wikipedia.org/wiki/Locale_(computer_software)) at least a *locale* is not only a string representing language/region but all "special variant preferences":
> In computing, a locale is a set of parameters that defines the user's language, region and any special variant preferences that the user wants to see in their user interface.

src/client/pages/OfferNew/Checkout/Checkout.stories.tsx
styrelius: I think we have too many things names `locale` and it might be confusing?

src/client/l10n/localePathPattern.ts
```diff
@@ -0,0 +1,27 @@
+export const LOCALE_URL_PARAMS = [
+  'se',
+  'se-en',
+  'no',
+  'no-en',
+  'dk',
+  'dk-en',
+] as const
```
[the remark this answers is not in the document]
styrelius: The reason for using `as const` is to be able to derive a type from this so we can be sure that the `:locale` URL parameters will always be the same as the keys in the `locales` object. But maybe this array should just switch files to the `/l10n/locales.ts`?

src/client/l10n/localePathPattern.ts
styrelius: (It's rarely a good sign when you feel like you've done something super smart)

src/client/l10n/birthDateAndSsnFormats.ts
styrelius: OK so it seems to mostly be pretty poor naming form my part - it's the English format, which will be used for all English locales and I guess that could be thought of as some sort of "default". But maybe it's better to just call it `EN`? Or WDYT?

src/client/l10n/localePathPattern.ts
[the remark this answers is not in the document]
mehdielket: Yes moving it to the other file could be a good. :)

src/client/l10n/birthDateAndSsnFormats.ts
robinandeer: Aha, yeah either that or add a "why" comment about it :)

src/client/l10n/useCurrentLocale.ts
styrelius: Actually, seems like we should do a `.toLowerCase()` in order to allow `/Se-EN` and other crazy versions of the locale params (which I think we _should_ allow and we also do today). But it has to be done earlier because at this stage there can only be lower case strings (because of the type awesomeness).

So I added so we're making  entire `location.pathname` lower case before passing it as an argument to `getLocaleFromPath`


image

src/client/l10n/birthDateAndSsnFormats.ts
styrelius: OK so actually it turned out it's the format we expect from back-end, and it's used for form validation in the details modal and to check that we have the expected format before we reverse in a function called `getFormattedBirthdate`.
So maybe a comment is enough, or call it something like `backEndDefault`? 🤔

src/client/pages/OfferNew/Checkout/Checkout.stories.tsx
styrelius: Also, it's clearer that this is the locale for the translations and not a `locale` in the sense it's now used in a bunch of other files?

src/client/l10n/birthDateAndSsnFormats.ts
robinandeer: Yeah, anything that helps us remember!

src/client/l10n/useCurrentLocale.ts
gustaveen: Awesome, good catch! 🙏

src/client/l10n/useCurrentLocale.ts
gustaveen: My perception is a bit off it seems 😄 So I think it's when we mix the naming like: `const locale = getLocaleFromPath(location.pathname)` since that is just a string and not an object. Just looking at the names you might expect `locale` and `currentLocale` to be of the same type. But I don't have a better for it so maybe we shouldn't overthink this 😄

src/client/l10n/useCurrentLocale.ts
styrelius: Well it's all thanks to you 😊

src/client/l10n/useCurrentLocale.ts
styrelius: Ah, I see. Hm, maybe I should just rename this `locale` on line 23.

styrelius: Maybe like this?
image

gustaveen: That works IMO! 🙌

mehdielket (approved):
Looks all good to me!

PR #607, opened 2021-09-22
Add credit check info for NO

joacimastrom:
[IPC-264](https://hedvig.atlassian.net/browse/IPC-264)

## What?
* Add info box with information about credit checks for new customers
* Add a modal with info for when sign fails

## Why?
* Legal probably?

image

image


src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -337,6 +340,10 @@ export const Checkout: React.FC<CheckoutProps> = ({
       {signStatus?.signState === SignState.Completed && (
         <CheckoutSuccessRedirect offerData={offerData} />
       )}
+      <SignFailModal
```
joacimastrom, line 343: I think this feedback is good for all markets, but story was only for Norway. Should I add a check for NO market or leave it for all?

src/client/pages/OfferNew/Checkout/CreditCheckInfo.tsx
```diff
@@ -0,0 +1,50 @@
...
+  color: ${colorsV3.gray900};
+  margin: 0;
+  line-height: 24px;
+`
+const InfoContent = styled.p`
+  color: ${colorsV3.gray700};
+  margin: 0;
+  margin-top: 0.25rem;
+  font-size: 0.875rem;
+  line-height: 20px;
```
robinandeer, line 32: We try to use rems for all values so they scale with user zoom level :)

```suggestion
  line-height: 1,25rem;
```

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -263,6 +271,10 @@ export const Checkout: React.FC<CheckoutProps> = ({
     })
       .then(({ data }) => {
         if (data?.signQuotes?.__typename === 'FailedToStartSign') {
+          if (data?.signQuotes.errorCode === 'MANUAL_REVIEW_REQUIRED') {
+            setShowFailModal(true)
+            return
+          }
```
robinandeer, line 277: Does this code only apply to NO users?

Otherwise, it seems like we don't give the other users any indication something went wrong. Since the `SignFailModal` is only rendered in NO.

However, if this error code is NO-specific we would not need to conditionally render the `SignFailModal`, right? That would remove some market specific logic which I think is the way forward!

src/client/components/ModalNew.tsx
```diff
@@ -7,6 +7,8 @@ import { Cross } from './icons/Cross'
 export interface ModalProps {
   isVisible: boolean
   dynamicHeight?: boolean
+  maxWidth?: string
+  overflowX?: string
```
robinandeer, line 11: I would prefer passing `className` to the modal and styling with:

```
const MyCustomModal = styled(Modal)`
  maxWidth: 23rem;
`
```

That is unless these values are meant to be truly dynamic :)

robinandeer (commented):
I think this looks good overall!

One thing I think about is how we could avoid market specific logic in multiple places in the code.

How about one single place that determines whether the "credit check" feature is enabled? (if `market === "NO"`). That way we would centralise the market specific stuff in one place and it would be easier to enable it for a second market :)

I realise this is not how we've implemented things before but I think it's a better way forward.

styrelius (commented):
Nice! Just one thing that I'd like to talk about before looking at the code - we already have an error message for _all markets_ when the signing fails. So I'm wondering if it's really necessary to add a specific one for the Norwegian market?

image

styrelius:
Here's the sign error message in Norwegian:
image

src/client/pages/OfferNew/Checkout/index.tsx
joacimastrom: Yeah I think this error-code is specifically for if credit-checks fail in NO, and if so I can remove the market-check for the SignFailModal. Will double check with backend!

src/client/components/ModalNew.tsx
joacimastrom: Yeah this seems a bit hacky I agree, but correct me if I'm wrong: Wouldn't your suggestion just add "maxWidth" to the root-element of `Modal`, which is the `<Wrapper>` and not the actual modal dialog?

joacimastrom:
> Nice! Just one thing that I'd like to talk about before looking at the code - we already have an error message for _all markets_ when the signing fails. So I'm wondering if it's really necessary to add a specific one for the Norwegian market?
> 
> image

Yeah I'm aware of this but I think they wanted something a bit more verbose and explicit for when credit-checks fail, so they don't just try again

styrelius:
> Yeah I'm aware of this but I think they wanted something a bit more verbose and explicit for when credit-checks fail, so they don't just try again

OK, I guess that kind of makes sense.

src/client/components/ModalNew.tsx
joacimastrom: Imo this is the real culprit
```const ModalInnerContainer = styled('div')`
  width: 100%;
  height: 100%;
`
```

This essentially makes previous max-width a fixed width. Shouldn't we to some extent let the content decide the size of the modal?

robinandeer: Aha, yeah that seems a little unnecessary I agree :/

You could technically pass a `className` prop to any of the child components of `<Modal>` -- whichever one makes most sense to style. And we we need more flexibility I would export multiple styled components from the module.

src/client/pages/OfferNew/Checkout/CreditCheckInfo.tsx
```diff
@@ -0,0 +1,50 @@
...
+import React from 'react'
+import { FadeIn } from 'components/animations/appearings'
+import { InfoIcon } from 'components/icons/Info'
+import { useTextKeys } from 'utils/textKeys'
+
+const InfoBox = styled.div`
+  padding: 1rem;
+  padding-left: 3rem;
+  margin-bottom: 1.5rem;
+  background-color: ${colorsV2.violet100};
```
gustaveen, line 12: We only use `colorsV3` :) 
```suggestion
  background-color: ${colorsV3.purple100};
```

src/client/components/utils/MarketFeatures.ts
```diff
@@ -0,0 +1,7 @@
+import { Market, useMarket } from './CurrentLocale'
+
+export const useCreditCheckInfo = () => {
+  const market = useMarket()
+
+  return [Market.No].includes(market)
+}
```
robinandeer, line 7: I really like this approach!

I would even name the module "features.ts" -- market is in the end just one way we differentiate features :)

joacimastrom: Maybe `featureToggles.ts`?

src/client/pages/OfferNew/Checkout/CreditCheckInfo.tsx
```diff
@@ -0,0 +1,50 @@
...
+  margin: 0;
+  line-height: 1.5rem;
+`
+const InfoContent = styled.p`
+  color: ${colorsV3.gray700};
+  margin: 0;
+  margin-top: 0.25rem;
+  font-size: 0.875rem;
+  line-height: 1.25rem;
+`
```
gustaveen, line 33: We try to follow the pattern of content being "pushed down" and using `margin-bottom`  :) 
```suggestion
const InfoHeader = styled.p`
  color: ${colorsV3.gray900};
  margin-top: 0;
  margin-bottom: 0.25rem;
  line-height: 1.5rem;
`
const InfoContent = styled.p`
  color: ${colorsV3.gray700};
  margin: 0;
  font-size: 0.875rem;
  line-height: 1.25rem;
`
```

src/client/pages/OfferNew/Checkout/SignFailModal.tsx
```diff
@@ -0,0 +1,55 @@
...
+    <Modal
+      isVisible={show}
+      onClose={onClose}
+      dynamicHeight
+      overflowX="auto"
+      maxWidth="36rem"
+    >
+      <Container>
+        <WarningTriangle />
+        <Header> {textKeys.CHECKOUT_SIGN_FAIL_HEADER()}</Header>
```
gustaveen, line 50: Guessing this space is accidental :) 
```suggestion
        <Header>{textKeys.CHECKOUT_SIGN_FAIL_HEADER()}</Header>
```

gustaveen (commented):
Good job @joacimastrom 🙌 Is there any of the norwegian SSNs that trigger this error in order to test the modal?

src/client/pages/OfferNew/Checkout/CreditCheckInfo.tsx
joacimastrom: But it's not the same 😰  🙈   In the design, it was `#efe8ff`. But I see your point, I'll let Zak know

src/client/pages/OfferNew/Checkout/SignFailModal.tsx
joacimastrom: Yes :)

src/client/pages/OfferNew/Checkout/CreditCheckInfo.tsx
joacimastrom: Okay, I'm used to the other way around. What's nice when the following component owns the spacing is that you don't have to worry about a leftover spacing at the end

src/client/components/ModalNew.tsx
joacimastrom: Ah cool! Wasn't aware of this :)

src/client/pages/OfferNew/Checkout/CreditCheckInfo.tsx
styrelius: Haha, do we? I'd say we shouldn't use `margin` at all but go for padding and spacer components but this is a way bigger discussion 🙈

joacimastrom:
> Good job @joacimastrom 🙌 Is there any of the norwegian SSNs that trigger this error in order to test the modal?

Not yet, backend isn't finished yet but I figure it's harmless to merge this since the error code doesn't exist yet anyway

src/client/components/icons/Info.tsx
```diff
@@ -1,16 +1,16 @@
 import React from 'react'
 
-export const InfoIcon = () => (
+export const InfoIcon = ({ height = 16, width = 16 }) => (
```
styrelius, line 3: I think this should always be square so maybe one prop, like `size`, is enough?

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -170,6 +171,7 @@ export const Checkout: React.FC<CheckoutProps> = ({
   const [signUiState, setSignUiState] = useState<SignUiState>('NOT_STARTED')
   const [emailUpdateLoading, setEmailUpdateLoading] = useState(false)
   const [ssnUpdateLoading, setSsnUpdateLoading] = useState(false)
+  const [showFailModal, setShowFailModal] = useState(false)
```
styrelius, line 174: I think we should stick to naming boolean variables with a prefix of `is` or `has` (or in special cases `should` or `can`). So this could be `isShowingFailModal` or `isFailModalVisible` for example 🙏

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -337,6 +343,10 @@ export const Checkout: React.FC<CheckoutProps> = ({
       {signStatus?.signState === SignState.Completed && (
         <CheckoutSuccessRedirect offerData={offerData} />
       )}
+      <SignFailModal
+        show={showFailModal}
```
styrelius, line 347: Same as above - `show` should IMO be `isVisible` :)

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
```diff
@@ -44,15 +46,14 @@ export const UserDetailsForm: React.FC<Props> = ({
...
-  const [emailError, setEmailError] = React.useState<boolean>(false)
-  const [ssn, reallySetSsn] = React.useState(() => initialSsn)
-  const [emailChangeTimout, setEmailChangeTimout] = React.useState<
-    number | null
-  >(null)
-  const [ssnChangeTimout, setSsnChangeTimout] = React.useState<number | null>(
+  const [email, reallySetEmail] = useState(() => initialEmail)
+  const [emailError, setEmailError] = useState<boolean>(false)
+  const [ssn, reallySetSsn] = useState(() => initialSsn)
+  const [showCreditCheckInfo, setShowCreditCheckInfo] = useState(false)
```
styrelius, line 52: Sorry, third comment (temporally, not on the page) about naming of booleans 🙃  See comment below for details

styrelius: And I know some other variables in this file don't follow this (looking at you old `emailError`!) but I guess that could be changed now as well, you know code scouting etc.

src/client/components/utils/featureToggles.ts
```diff
@@ -0,0 +1,7 @@
+import { Market, useMarket } from './CurrentLocale'
+
+export const useCreditCheckInfo = () => {
```
styrelius, line 3: This is really not important but maybe this doesn't need to be a hook but it could take the market as a parameter instead and just be a regular ol' function? Then it could be named something like `hasCreditCheckInfo`.

styrelius (commented):
Great job!! 🥳 Just a few comments, almost all of them related to naming of boolean parameters 🙃

src/client/components/utils/featureToggles.ts
joacimastrom: I considered that as well, even have it as a plain object as 
```
export const hasCreditCheckInfo = {
   [Market.No]: true
}
```
but that forces the `useMarket`-usage onto the parent component,  imo it's better to have it like this, to be used without dependencies 🤷‍♂️

styrelius: Sure, I don't feel super strongly about this 😊  (the `useMarket` function is on its way out though, but changing this function to the new ways will be easy so that really doesn't matter). 

But one other thing is that I think this file should probably be in the `/src/client/utils` folder instead of under `/components` 🤔

src/client/pages/OfferNew/Checkout/SignFailModal.tsx
```diff
@@ -34,22 +34,21 @@ const Content = styled.p`
   line-height: 1.25rem;
 `
 
+const CustomModal = styled(Modal)`
+  overflow-x: auto;
+  max-width: 36rem;
+`
```
robinandeer, line 40: Nice, I like it!

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -171,7 +171,7 @@ export const Checkout: React.FC<CheckoutProps> = ({
   const [signUiState, setSignUiState] = useState<SignUiState>('NOT_STARTED')
   const [emailUpdateLoading, setEmailUpdateLoading] = useState(false)
   const [ssnUpdateLoading, setSsnUpdateLoading] = useState(false)
-  const [showFailModal, setShowFailModal] = useState(false)
+  const [showFailModal, setShowFailModal] = useState(true)
```
robinandeer: Is this a mistake from local testing?

robinandeer (dismissed):
Looks good - like the changes to override styles rather than adding more props 😌

src/client/pages/OfferNew/Checkout/index.tsx
joacimastrom: Yep :) Has been fixed since

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
```diff
@@ -142,11 +145,13 @@ export const UserDetailsForm: React.FC<Props> = ({
         inputMode="numeric"
         pattern="[0-9]*"
         maxLength={ssnMaxLength}
+        onFocus={() => setIsShowingCreditCheckInfo(true)}
         value={ssn}
         errors={ssnBackendError ? textKeys[ssnBackendError]() : undefined}
         onChange={handleSsnChange}
         onBlur={handleSsnBlur}
       />
+      {useCreditCheckInfo() && isShowingCreditCheckInfo && <CreditCheckInfo />}
```
robinandeer, line 154: I just a little confused to see a hook called in the return statement. Maybe move it up and assign it to a variable first?


```
const hasEnabledCreditCheckInfo = useCreditCheckInfo()
```

robinandeer (dismissed):
Cool! Just have a look at the hook-comment I wrote. Perhaps we should move it out of the JSX-block

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
joacimastrom: Done

joacimastrom:
> Cool! Just have a look at the hook-comment I wrote. Perhaps we should move it out of the JSX-block

@robinandeer Done :) 
It's a bit of a strange hook but it does the job as a first iteration. Later on we would want to have the feature toggles dynamic and not hard-coded, but that's a bit out of scope 😅

robinandeer (approved):
Thanks for updating the hook-call!

I agree that we should keep iterating on the feature toggles -- I think @mehdielket is already thinking about that ;)

PR #612, opened 2021-10-06
 Fix/ add storybook to readme

dendoesit:
## What?

Added storybook to readme

## Why?

Because it was missing.


gustaveen (approved):
Great!

PR #615, opened 2021-10-08
Chore / General package updates

styrelius:
## What?

The good ol' `yarn upgrade`

## Why?

We should keep our dependencies fresh.

<hr />

We also need to upgrade certain packages to newer major versions (see [this Notion page]() for more information) but I think it makes sense to split it up, so we don't risk things becoming messy :)

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3,23 +3,22 @@
 
 
 "@apollo/client@^3.0.0-beta.23", "@apollo/client@^3.2.3", "@apollo/client@latest":
-  version "3.3.11"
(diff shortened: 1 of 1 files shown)


robinandeer:
Do things usually look like this in Storybook?

Screenshot 2021-10-13 at 13 44 14

styrelius:
@robinandeer Good catches 🎣 Will try to fix this :) Don't think it usually looks like that in Storyblok but I rarely use Storyblok so forgot to check that 🙈

styrelius:
I'm closing this. Me or someone else can start over with it from our current `master` branch 🧹

PR #617, opened 2021-10-13
Disable credit checks temporarily

gustaveen:
## What?
Disable frontend for credit checks in prod

## Why?
Backend is not ready and we need to deploy to prod


robinandeer (approved):
Looks good :)

PR #619, opened 2021-10-13
GRW-443 / Feature / Styling updated to OfferPage address

dendoesit:
GRW-443 / Feature / Styling updated to OfferPage address
## What?

Updated font size of Address in Offer Page to match Name . 

## Why?

Design updates

image


src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
```diff
@@ -68,7 +68,8 @@ const Address = styled.div`
   display: inline-block;
 
   ${MEDIUM_SCREEN_MEDIA_QUERY} {
-    font-size: 2.5rem;
+    font-size: 3rem;
+    line-height: 3.5rem;
```
robinandeer, line 72: Should we prefer unit-less line heights?

```suggestion
    line-height: 1.17;
```

> In most cases, this is the preferred way to set line-height and avoid unexpected results due to inheritance.

https://developer.mozilla.org/en-US/docs/Web/CSS/line-height

I also find it easier to read these values. They should also apply even across breakpoints so we can set it ones per component.

src/client/pages/OfferNew/Perils/PerilCollection/PerilItem.tsx
```diff
@@ -13,11 +14,12 @@ const OuterContainer = styled.div`
   position: relative;
   display: flex;
   color: ${colorsV3.gray900};
+  box-shadow: 0px 1px 2px rgba(0, 0, 0, 0.1);
 
   &:before {
     content: '';
 
-    @media (min-width: 800px) {
+    ${MEDIUM_SCREEN_MEDIA_QUERY} {
```
robinandeer, line 22: Nice code scouting!

robinandeer (commented):
Nice update! Don't forget to add some screenshots for PRs with visual updates -- then you'll have my "approve" ;)

src/client/pages/OfferNew/Introduction/HeroOfferDetails.tsx
dendoesit: @robinandeer changed the entire file to be according.

robinandeer (approved):
Cool, looking good!

PR #621, opened 2021-10-14
Tracking - Send passage data to Google Tag Manager

gustaveen:
## What?
- Send Embark passage data to Google Tag Manager datalayer
- Filter out undefined values since payload passes down all properties we want to track, but all are undefined in the start of the flow

## Why?
We want to get rid of Segment and pass data directly to GTM. Sonny told me to still keep Segment data push since Mixpanel is still being used for some dashboards. Aim is to get rid of it later

Screenshot 2021-10-14 at 10 14 41

### DIFF
--- a/src/client/pages/Embark/index.tsx
+++ b/src/client/pages/Embark/index.tsx
@@ -223,7 +223,7 @@ export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
     },
   ) => {
     const filteredArray = Object.entries(payload).filter(
-      ([key, value]) => value !== undefined,
+      (property) => property[1] !== undefined,
     )
 
     const payloadObject = filteredArray.reduce((prev, item) => {
--- a/src/client/pages/Embark/index.tsx
+++ b/src/client/pages/Embark/index.tsx
@@ -216,6 +216,39 @@ export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
 
   const textKeys = useTextKeys()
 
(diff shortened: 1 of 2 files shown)


src/client/pages/Embark/index.tsx
```diff
@@ -215,6 +216,39 @@ export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
 
   const textKeys = useTextKeys()
 
+  const trackPassageData = (
+    eventName: string,
+    payload: {
+      [key: string]: any
+    },
```
robinandeer, line 223: I prefer:

```suggestion
    payload: Record<string, any>,
```

src/client/pages/Embark/index.tsx
```diff
@@ -215,6 +216,39 @@ export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
...
+  ) => {
+    const filteredArray = Object.entries(payload).filter(
+      (property) => property[1] !== undefined,
+    )
+
+    const payloadObject = filteredArray.reduce((prev, item) => {
+      const [key, value] = item
+      return { ...prev, [key]: value }
+    }, {})
```
robinandeer, line 232: If you wanna get fancy:

```suggestion
		const payloadObject = { ...payload }
    Object.keys(payloadObject).forEach(
    	(key) => payloadObject[key] === undefined && delete payloadObject[key],
  	)
```

robinandeer (dismissed):
Looks good to me! Feel free to discard my suggestions 😉

PR #623, opened 2021-10-14
Remove description field from edit quote mutation

robinandeer:
## What?

Remove description field from mutation.

## Why?

Remove field not part of schema anymore.
Broken in production.

_Referenced ticket(s): [GRW-458]_

Co-Authored-By: Mehdi <89857278+mehdielket@users.noreply.github.com>

[GRW-458]: https://hedvig.atlassian.net/browse/GRW-458?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #624, opened 2021-10-14
GRW-458 / Update codegen client to fix `editQuote` mutation

robinandeer:
## What?

Update codegen client.

## Why?

Forgot to update it after updating the `editQuote` mutation in #623.


styrelius (approved):
👑

PR #630, opened 2021-10-17
GRW-448 / Fix / Separate date for home accident Sweden

mehdielket:
## What?

When displaying the Swedish Home + Accident insurance bundle we should allow users to choose between two start dates, one for each insurance.

## Why?

Allows users to get more configurability

_Referenced ticket(s): [GRW-448]_

## Screenshots
Screenshot 2021-10-17 at 20 19 02
Screenshot 2021-10-17 at 20 19 12

[GRW-448]: https://hedvig.atlassian.net/browse/GRW-448?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robinandeer (changes requested):
Somethings is still not right here 😅 Now the switch safety section is back even if we don't use it anymore - just delete that file before we merge.

The start date update looks good tho!

PR #610, opened 2021-10-04
GRW-391 / Feature / Toggle Quotes on Offer Page (logic)

robinandeer:
Note: Copy on top of the insurance selector is not finalised

## What?

- Integrate insurance selector component on offer page.
- Setup logic needed to toggle possible quotes.
- Use the new "possibleVariations" field from the API.
- Add feature toggle for market / environment

## Why?

- So we can work with Home Contents and Accident insurance in SE.
- The changes will apply globally but can be controlled via the API.

## Demo

https://user-images.githubusercontent.com/89857278/136381732-87c26258-c712-445a-93e2-33f1d37b12de.mov

PR #631, opened 2021-10-18
Implement insurance selector/toggle on offer page

robinandeer:
> Note: opened in favour of #610 

## What?

- Integrate insurance selector component on offer page.
- Setup logic needed to toggle possible quotes.
- Use the new "possibleVariations" field from the API.
- Add feature toggle for market / environment

## Why?

- So we can work with Home Contents and Accident insurance in SE.

## Demo

Screenshot 2021-10-18 at 15 36 40

localhost_8040_se_new-member_offer(iPad Pro)

https://user-images.githubusercontent.com/89857278/136381732-87c26258-c712-445a-93e2-33f1d37b12de.mov

_Referenced ticket(s): [GRW-391]_

Co-Authored-By: Mehdi <89857278+mehdielket@users.noreply.github.com>

[GRW-391]: https://hedvig.atlassian.net/browse/GRW-391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/utils/hooks/useFeature.ts
```diff
@@ -0,0 +1,59 @@
...
+
+const Config: readonly FeatureConfig[] = [
+  {
+    name: Features.OFFER_PAGE_INSURANCE_TOGGLE,
+    envs: ['development', 'staging'],
+    markets: [Market.Se],
+  },
+  {
+    name: Features.CHECKOUT_CREDIT_CHECK,
+    envs: ['development', 'staging', 'production'],
```
gustaveen, line 24: I'm thinking if we have a feature enabled in prod, maybe we could remove the envs field? So if you don't specify envs it's enabled in all envs. Also would be nice if you wouldn't need to specify both dev and staging since they should always be enabled in at the same time. All in all I think this is a great 🙌

src/client/utils/hooks/useQuoteIds.ts
```diff
@@ -7,28 +7,53 @@ import { StorageState, useStorage } from '../StorageContainer'
...
+  selectedQuoteIds: ReadonlyArray<string>
+} => {
   const storedIds = storage.session.getSession()?.quoteIds
-  if (storedIds) {
-    return storedIds
+  const selectedQuoteIds = storage.session.getSession()?.selectedQuoteIds
+  if (storedIds?.length) {
+    return {
+      quoteIds: storedIds,
+      selectedQuoteIds: selectedQuoteIds?.length ? selectedQuoteIds : storedIds,
```
gustaveen, line 46: 👍

src/client/utils/hooks/useQuoteIds.test.ts
```diff
@@ -0,0 +1,52 @@
...
+// Needs investigation
+const mockUrlParametersQuery = 'quoteIds=[1,2,3]'
+jest.mock('react-router', () => ({
+  ...jest.requireActual('react-router'),
+  useLocation: () => ({
+    search: `?${mockUrlParametersQuery}`,
+  }),
+}))
+
+it('should return correct quoteIds from storage', () => {
```
gustaveen, line 28: Awesome you added tests for this 👍

gustaveen (dismissed):
Solid work guys, I think this is good to go! 🙌 I had some comments on the feature flags but that can be iterated on later of course.

dendoesit (approved):

PR #632, opened 2021-10-19
Update chevron down icon (make it thinner)

robinandeer:
## What?

... and rename "DownArrow" to align with icon library.

## Why?

Feedback from @zakkjellin. We need to continue to align with the updated designs.

## Before and after

<table>
<tr>
	<td>
	<td>
</table>


src/client/pages/OfferNew/Compare/CompareTable.tsx
```diff
@@ -288,7 +288,7 @@ const MobileDropdownWrapper = styled('div')`
   max-width: 85px;
   margin-bottom: 3.25rem;
 `
-const TinyDownArrow = styled(DownArrow)`
+const TinyDownArrow = styled(ChevronDown)`
```
gustaveen, line 291: Not crucial for this PR but this styled TinyDownArrow could be replaced then with `<ChevronDown size="7" />` :)

gustaveen (approved):
🔛🔝

PR #633, opened 2021-10-19
Update language picker

robinandeer:
## What?

Remove dropdown + old icon.
Reuse the same language picker on forever page.
Remove forever-specific language picker.
Update tests and snapshots.

## Why?

We noticed that the language picker is not usable on mobile (offer page).

Zak has also made updates to it recently so would be good to implement that design.

_Referenced ticket(s): []_

## Demo

Screenshot 2021-10-19 at 15 47 13
 
Screenshot 2021-10-19 at 15 49 51

Screenshot 2021-10-19 at 15 44 22


gustaveen (approved):
🌎

PR #634, opened 2021-10-21
GRW-477 / Fix / `isLoading ` check in `useQuoteIds` hook

robinandeer:
## What?

Fix the loading quote ids check.

## Why?

The value is never null. Strange that TS didn't complain before.
We are sending an extra query before the ids have loaded.

_Referenced ticket(s): [GRW-477]_

[GRW-477]: https://hedvig.atlassian.net/browse/GRW-477?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gustaveen (approved):
🏆

PR #637, opened 2021-10-22
GRW-456 / Fix / Improve-loading-time-on-offer-page

mehdielket:
## What?

Improves the performance of the offer page by doing the following:
* Requesting only the possibleVariations in the QuoteBundle query
* Requesting only the root bundle in the QuoteBundle query while signing
* Fixing bug with requesting QuoteBundle without quoteIds

## Why?

The performance of the QuoteBundle query has significantly deteriorated when adding bundle variations. We use it often especially after changing the date or adding discounts, so it's performance it important. 

_Referenced ticket(s): [GRW-456]_

[GRW-456]: https://hedvig.atlassian.net/browse/GRW-456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robinandeer (approved):
Okey fair enough, looks good from my POV!

I'm just curious how other people think about removing the feature flag already. Don't want to block staging if something unexpected comes up.

robinandeer (approved):
Cool stuff! I would suggest we do some "stress testing" in a review app across markets but otherwise I think it looks good :)

PR #639, opened 2021-10-22
GRW-468 / Fix / Override iOS 15 button text color in calendar

robinandeer:
## What?

☝️ 

## Why?

iOS 15 thinks buttons should be blue.

_Referenced ticket(s): []_

Screenshot 2021-10-22 at 16 41 20


dendoesit (approved):
great !

PR #640, opened 2021-10-23
GRW-469 / Fix / Prevent sidebar overlapping FAQs

robinandeer:
## What?

Clear text so sidebar doesn't overlap text in FAQ section.
Using a fluid typography technique: https://css-tricks.com/snippets/css/fluid-typography/

## Why?

It's a bit of a complicated solution. However, we don't wrap the FAQ-section in the flexbox container, so it's not that easy to make sure the FAQ text content doesn't flow under the sidebar.

Still I'm kind of happy with how well it actually works ;)

## Demo

https://user-images.githubusercontent.com/1220232/138550380-68bdb1ac-6db3-4513-ac2e-e124ed555c10.mov


styrelius (approved):
👌

PR #641, opened 2021-10-25
Add route for switcher without accident

gustaveen:
## What?
Add route for switcher without accident

## Why?
So we can go live before we have fixed cross sell for switchers


robinandeer (commented):
Nice, see comments :)

robinandeer (approved):
Nice solution! Now I understand much easier how to read the code 👏

PR #644, opened 2021-10-25
GRW-470 / Copy / Update placeholder for discount code to "Your code"

mehdielket:
## What?

Updates the placeholder for discount code to "Your code"

## Why?

The "XXXXX" copy was confusing to users since it assumed a number of characters that was not accurate

_Referenced ticket(s): [GRW-470]_

Screenshot 2021-10-25 at 16 43 11

[GRW-470]: https://hedvig.atlassian.net/browse/GRW-470?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robinandeer (approved):
Wow lot's of nice updates here!

I thought that placeholder usually should be an example of the input. Not sure how possible that is to come up with for the discount code tho...

Anyways, this is better than before!

PR #645, opened 2021-10-26
Bugfix horizontal scroll

dendoesit:
## What?

On the offer page you can horizontally scroll making the bottom scrollbar appear. 

## Why?

Horizontal scrolling should not be happening. 
This is caused because the image has width:100vw which is different from width: 100% as viewport width doesn't take into consideration the scrollbar width. 
More can be found here : https://destroytoday.com/blog/100vw-and-the-horizontal-overflow-you-probably-didnt-know-about

_Referenced ticket(s): []_

Before
image

After 
image


robinandeer (approved):
Looks good in Desktop Safari :)

I like that you reference the blog post

PR #647, opened 2021-10-28
IPC-330 / feature / Group terms by offer in checkout

joacimastrom:
## What?

* Remove strange conversion on terms from array to Map
* Remove logic and tests for filtering and ordering of terms in checkout (Will move ordering to PCMS)
* Render terms grouped by quote

## Why?

* Current sorting/ordering is not scaleable with new types of insurance documents, mainly "Pre-purchase info" which is also contract-specific

_Referenced ticket(s): [IPC-330]_

image

[IPC-330]: https://hedvig.atlassian.net/browse/IPC-330?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


styrelius (commented):
Great! 🙌 👑  Just one minor thing 👇 

But also, just so I understand correctly - you will add sorting through PromiseCMS or something so that we get the T&C links in a reasonable order? Like e.g. the "Villkor and förköpsinformation" (i.e. `type` `'TERMS_AND_CONDITIONS'`) always being first, and the order being the same for all the insurance products? So it won't look like this:
image

styrelius (dismissed):
Awesome! 🥳

robinandeer (dismissed):
Very nice! I think we can merge this first and I will build on top of these changes in #646 

https://media.giphy.com/media/etGPvG1Pqj9ks/giphy.gif

styrelius (approved):

PR #649, opened 2021-10-29
Fix layout issue caused by tooltip overflow

robinandeer:
## What?

Display tooltip popover to the right (instead of above) for smaller screens. Same solution as before for larger screens.

I took inspiration from this guide: https://www.w3schools.com/css/css_tooltip.asp

## Why?

Currently the tooltip causes layout issues on smaller screens:

Screenshot 2021-10-29 at 09 27 23

## Demo

@zakkjellin let us know if you have any input how this is working from you POV!

Mobile

Screenshot 2021-10-29 at 09 48 02

Desktop

Screenshot 2021-10-29 at 09 48 24


mehdielket (approved):
Looks good! Tests great! 🚀

PR #650, opened 2021-11-01
GRW-494: Fix check icon color on onboarding page

guilhermespopolin:
## What?

- `defaultProps` was removed from`IconRoot` component
- `InfoIcon` `color` setting is now being done on places where we render it (by passing in a `color` prop) rather on its implementation

## Why?

`defaultProps` will eventually be deprecated so instead of fixing it by replacing the `IconRoot.defaultProps` to `InfoIcon.defaultProps` I decided to remove it

Referenced ticket(s): [GRW-494](https://hedvig.atlassian.net/browse/GRW-494)

### Before
before

### After
Screenshot 2021-11-01 at 15 38 00


gustaveen:
Aha, good find @guilhermespopolin 🙌 We shouldn't have a default prop for IconRoot. Would this still make the Info icon to be able to have white color?

src/client/components/icons/Info.tsx
```diff
@@ -3,7 +3,12 @@ import { colorsV3 } from '@hedviginsurance/brand'
 import { IconRoot, IconRootProps } from './IconRoot'
 
 export const InfoIcon: React.FC<IconRootProps> = (props) => (
-  <IconRoot viewBox="0 0 16 16" xmlns="http://www.w3.org/2000/svg" {...props}>
+  <IconRoot
+    viewBox="0 0 16 16"
+    xmlns="http://www.w3.org/2000/svg"
+    color={colorsV3.gray900}
+    {...props}
+  >
```
robinandeer, line 11: What do you think about this solution? It would set a default value that can still be overridden if we use the icon on a dark background for example.

```
export const InfoIcon: React.FC<IconRootProps> = ({
  color = colorsV3.gray900,
  ...props
}) => (
  <IconRoot
    viewBox="0 0 16 16"
    xmlns="http://www.w3.org/2000/svg"
    color={color}
    {...props}
  >
```

robinandeer (commented):
I didn't know that `defaultProps` were deprecated - good catch!

guilhermespopolin:
> Aha, good find @guilhermespopolin 🙌 We shouldn't have a default prop for IconRoot. Would this still make the Info icon to be able to have white color?

Actually @robinandeer was the one that found it so big kudos to him :)
With this, `InfoIcon` will keep its default color but in this case its a gray color :)

src/client/components/icons/Info.tsx
robinandeer: Hmm you are probably right that "currentColor" should be the default (IconRoot takes care of that).

I just suggest this to keep the behavior from before. But it's probably better to pass the color in the context where we use the icon :)

guilhermespopolin: Yeah, at first I tried this one as it's kind of the standard way do deal with `defaultProps` with Typescript. Also works, I guess I didn't move on with it because I still would need to explicitly set `color` props to `IconRoot` so I thought it would be simpler just added it and still spread `props`. This solution shines when we have scenarios like the following, where we guarantee default values but don't intent to spread the props:

```
type LabelProps = { text?: string }

function Label({ text = 'default' }: LabelProps) {
  return <span>{text}</span>
}
```

However, I'm totally fine to change it to the proposed way :)

guilhermespopolin: Gonna remove `color` definition from `InfoIcon` then :)

guilhermespopolin: One thing tho, is that if we remove `color={colorsV3.gray900}` fron `InfoIcon` we would need to check for places where we're using `InfoIcon` without setting a `color` and end up setting it. Like [here](https://github.com/HedvigInsurance/web-onboarding/blob/master/src/client/pages/OfferNew/Checkout/CreditCheckInfo.tsx#L41) for example, right?!

robinandeer: Oh yeah, I think that's the way to go!

Explicitly pass color to `InfoIcon` wherever it's implemented :)

gustaveen: Sounds good, it's never bad to set the color explicitly :) But my thought with using `currentColor` is that you don't have to since it inherits from it's parent that has color. Or was ther an issue with the color in any other browser? 

https://user-images.githubusercontent.com/6661511/139806626-9b518c2b-6d9a-4cd1-9fef-9275841b78e8.mov

.

robinandeer: I agree. We should only override the and give an explicit color if it differs from the color of the text :)

src/client/pages/OfferNew/Perils/AmountCollection/AmountItem.tsx
```diff
@@ -58,21 +57,6 @@ const Value = styled.p`
   }
 `
 
-const StyledInfoIcon = styled(InfoIcon)`
```
guilhermespopolin, line 61: While checking for references to `InfoIcon` I found this one that it seems not to be used anywhere else, that's why the removal. Can you guys double check that?

gustaveen: Good catch, that can be removed 🙌

gustaveen (approved):
Great job! @guilhermespopolin 🙌

mehdielket (approved):
Tested it locally! Nicely done 🙂

PR #655, opened 2021-11-03
Tracking - add initial offer and discounted premium to GTM event

gustaveen:
## What?
Add `initial_offer` and `discounted_premium` to GTM event

## Why?
Update based on this tracking document from Sonny https://docs.google.com/spreadsheets/d/1IHswc17yIOPkXTGA5aaLsgLRX1lhRvrKaVayzx7RAUo/edit#gid=0

Screenshot 2021-11-02 at 17 11 38

### DIFF
--- a/src/client/utils/tracking/tracking.ts
+++ b/src/client/utils/tracking/tracking.ts
@@ -15,6 +15,7 @@ import {
   isDanishTravelBundle,
   isStudentOffer,
   getMainQuote,
+  isSwedish,
 } from 'pages/OfferNew/utils'
 import { trackOfferGTM } from './gtm'
 import { adtraction } from './adtraction'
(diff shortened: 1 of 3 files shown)


src/client/utils/tracking/tracking.ts
```diff
@@ -44,6 +47,11 @@ export const getUtmParamsFromCookie = (): UtmParams | undefined => {
   }
 }
 
+export enum SeBundleTypes {
+  SeAccidentBundle = 'SE_ACCIDENT_BUNDLE',
+  SeAccidentBundleStudent = 'SE_ACCIDENT_BUNDLE_STUDENT',
+}
```
gustaveen, line 53: I see that we both added this types @mehdielket :) Let's see if I can use yours, I'm not sure I need to know about brf, house or rent though.

mehdielket: Yes. I confirmed this with Sonny and Fred so we need to granularity

gustaveen: For sure, using your type instead 🙌

mehdielket (changes requested):

mehdielket (approved):

PR #656, opened 2021-11-03
GRW-502: updated translation files

guilhermespopolin:
## What?

- Translation files were updated;

## Why?

- So you can have access to newer translations;

_Referenced ticket(s): [GRW-502](https://hedvig.atlassian.net/browse/GRW-502)_


src/translations/da_DK.json
```diff
@@ -304,5 +304,6 @@
   "OFFER_BUNDLE_SELECTOR_DESCRIPTION": "",
   "OFFER_APP_PROMOTION_HEADING": "En forsikringsapp, der løser dine problemer",
   "OFFER_APP_PROMOTION_PARAGRAPH": "",
-  "WEB_OFFER_DOCUMENTS_SECTION_TITLE": "Vigtige dokumenter"
+  "WEB_OFFER_DOCUMENTS_SECTION_TITLE": "Vigtige dokumenter",
+  "STARTPAGE_UNINSURED_BODY_WITHOUT_ACCIDENT": "Beregn prisen på en ny indboforsikring hos Hedvig"
```
styrelius, line 308: This text key is actually only used for Sweden so there's no need for a Danish translation. That said, I still think we can merge this since this text key is only temporary and should be removed when we launch accident SE right?

styrelius (approved):
🇩🇰 🇳🇴 🇸🇪 🇫🇷

src/translations/da_DK.json
robinandeer: Yes that's how I also understand it :)

src/translations/en.json
```diff
@@ -301,8 +301,9 @@
...
+  "OFFER_BUNDLE_SELECTOR_DESCRIPTION": "Pick the insurance plan that fits you best. The insurances cover you and all co-insured. You can always add the accident insurance or change the activation date for your insurances in the app later on.",
   "OFFER_APP_PROMOTION_HEADING": "Take care of everything in the Hedvig app",
-  "OFFER_APP_PROMOTION_PARAGRAPH": "If you've had an accident, you can easily report the injury directly in the app, no matter where you are. In some cases, you won't even have to wait 5 minutes for your compensation. We are available between 08-22 on weekdays and 9-18 on weekends.",
-  "WEB_OFFER_DOCUMENTS_SECTION_TITLE": "Important documents"
+  "OFFER_APP_PROMOTION_PARAGRAPH": "Make a claim, manage your insurances or ask questions. With the app you’ll have full control over your insurances and get instant help if something has happened.",
+  "WEB_OFFER_DOCUMENTS_SECTION_TITLE": "Important documents",
+  "STARTPAGE_UNINSURED_BODY_WITHOUT_ACCIDENT": "Get a price quote on our home insurance and the new complementary accident insurance."
```
robinandeer, line 308: This one is just a fallback, right? Won't show up in production - otherwise we leak that we are launching accident...

robinandeer (commented):
Think it looks good - just double check that "en" translation doesn't show up in production (see code comment)

src/translations/en.json
styrelius: We're also kind of leaking _here_ that we're launching accident 🤪

robinandeer: Haha but only if you look into the source -- and I think in that case we have plenty of GH issues that reveals our plans anyways 😅

robinandeer:
Wait no changed files??? @guilhermespopolin

guilhermespopolin:
> Wait no changed files??? @guilhermespopolin

Wat?! That's weird 🤔
Ok, maybe that's because the translations were already on master and since master was merged into my branch its makes sense having no file changed. I believe I can close this one then

PR #658, opened 2021-11-03
Chore / Change node version 12.19.0 -> 14.17

styrelius:
### Maintenance day chore
Node version changed from `12.19.0` to `14.17`.

I figured there was no need for a fixed patch version but if you disagree we change it to a specific patch version :)

## Why?

We've agreed it's time. Also fresh 🥝 🍍


robinandeer (approved):
Oh nice! And no issues with Codefresh even :D

Perhaps it's enough to specify:

```
"engines": {
  "node": "14.x"
}
```

At least that's suggested by Heroku:
https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version

styrelius:
@robinandeer Hm, yeah maybe? But if we change the `"node"` engine in `package.json` we can't specify the minor version anywhere, right? Or? That might be fine, though 🤔 Not sure what the benefits are with specifying a minor version. Or can we do it like `14.17.x`?

robinandeer:
Yeah it's confusing to have three files that all specify node version 😂

So probably most clear to keep them in sync (same version everywhere).

Definitely think we can do "14.7.x" though

guilhermespopolin (approved):

PR #659, opened 2021-11-03
chore: workspace settings

guilhermespopolin:
## What?

1. Adds webpack folder into `.eslintignore`
2. Adds extension recommendations list into `vscode` configuration

## Why?

1. It's not a big deal. `Eslint` red underlines triggers me hahaha
2. Again, just another suggestions. I saw that `vscode` provides a cool way to share essential extensions between developers. By clicking into the button shown bellow, all recommended  extensions would be installed:
extensions


.vscode/extensions.json
```diff
@@ -0,0 +1,7 @@
+{
+  "recommendations": [
```
robinandeer, line 2: Love this! Much better than documenting in README 😌

.vscode/settings.json
```diff
@@ -1,3 +1,5 @@
 {
-  "typescript.tsdk": "node_modules/typescript/lib"
+  "typescript.tsdk": "node_modules/typescript/lib",
+  "editor.defaultFormatter": "esbenp.prettier-vscode",
+  "editor.formatOnSave": true
```
robinandeer, line 4: This we should perhaps discuss or align - we've discussed weather or not to share vscode settings but forget what we agreed on.

The typescript setting I think would be great to share tho!

src/client/components/buttons.tsx
```diff
@@ -13,7 +13,9 @@ interface ButtonProps {
 }
 
 export const Button = styled.button<ButtonProps>`
-  display: inline-block;
+  display: inline-flex;
+  align-items: center;
+  gap: 0.5rem;
```
robinandeer, line 18: Have you checked if flex + gap has sufficient browser support?

src/client/pages/Debugger/ActualDebugger.tsx
```diff
@@ -27,7 +30,9 @@ export const ActualDebugger: React.FC = () => {
           {(storage) => (
             <>
               <Button
+                disabled={isCleaningQuoteState}
                 onClick={() => {
+                  setIsClearningQuoteState(true)
```
robinandeer, line 35: Is anything resetting this to false? 🤔

robinandeer (commented):
Very nice updates! Sorry for a lot of comments. I was reviewing on my phone so might have missed something obvious 🙃

One suggestion: this PR does three things in one as you clearly explain. However, I think it would be much easier merge them separately. A PR can't be too small IMO as long as it's atomic ;)

guilhermespopolin:
> Very nice updates! Sorry for a lot of comments. I was reviewing on my phone so might have missed something obvious 🙃
> 
> One suggestion: this PR does three things in one as you clearly explain. However, I think it would be much easier merge them separately. A PR can't be too small IMO as long as it's atomic ;)

You're right. At least I could have created two PR:

- One for tooling related things: changes on `vscode` settings, `eslint` and so on;
- And another one that adds the loading state to debugger page

src/client/components/buttons.tsx
guilhermespopolin: Yes. I mean, it seems to be well supported now. It's being used in other places in the codebase as well :)

robinandeer: Hmm are you sure 😉 or are we using grid + gap because I have know that has better browser support

88% users covered it seems: https://caniuse.com/flexbox-gap
But still a Working Draft...

src/client/pages/Debugger/ActualDebugger.tsx
guilhermespopolin: 🤦🏻‍♂️ Good catch. I mean, it's working because at the end of this function we force a reload. But definitely I should have reset it to `false`

src/client/components/buttons.tsx
guilhermespopolin: 🤔 Yeah, you're right. Gonna refactor it.

robinandeer: I can't wait for that to be officially part of CSS though -- we should start using it as soon as it lands 😅

src/client/pages/Debugger/ActualDebugger.tsx
robinandeer: Ahh right, understand. But just because it's confusing when reading the code it might be worth implementing it hehe

src/client/components/buttons.tsx
guilhermespopolin: Me too!

src/client/pages/Debugger/ActualDebugger.tsx
```diff
@@ -18,6 +19,8 @@ const Row = styled.div`
 `
 
 export const ActualDebugger: React.FC = () => {
+  const [isCleaningQuoteState, setIsClearningQuoteState] = useState(false)
```
styrelius, line 22: Typo 🙈 
```suggestion
  const [isCleaningQuoteState, setIsCleaningQuoteState] = useState(false)
```
Also, I think a better name might be
```suggestion
  const [isResettingAllState, setIsResettingAllState] = useState(false)
```
or
```suggestion
  const [isResettingQuoteState, setIsResettingQuoteState] = useState(false)
```

src/client/pages/Debugger/components/QuoteData.tsx
```diff
@@ -377,11 +378,15 @@ export const QuoteData: React.FC<OfferProps> = ({ sessionToken }) => {
...
 
                       <Button
+                        type="submit"
                         background={colorsV3.purple500}
                         foreground={colorsV3.gray900}
-                        type="submit"
+                        disabled={props.isSubmitting}
                       >
                         Create quote
+                        {props.isSubmitting && (
```
styrelius, line 387: I would suggest that in addition to having the loading indicator on the button also not render the form when `isSubmitting` (or just as much of the form as is visible when we have an error). Since you can submit the form by selecting an input field and pressing return key it makes sense that a loading indicator would be visible even if you don't scroll to the bottom of the page.

.vscode/settings.json
guilhermespopolin: I understand. I'm not aware of a way to set it up as a recommendation. So I think it would be better to revert this and keep these settings about the editor as a personal preference :)

.vscode/extensions.json
```diff
@@ -0,0 +1,7 @@
...
+    "dbaeumer.vscode-eslint",
```
styrelius, line 3: Nice! I started writing a list with extension tips/recommendations (after you asked about this) that I'll put in Notion. Since we have a bunch of repos that _might_ be a more scalable approach? Or the two can complement each-other. Anyway, great initiative! ✨

styrelius (changes requested):
Nice job! Agree with Robin though, that this should be split up into one PR for each... thing :)

.vscode/extensions.json
guilhermespopolin: I see. Perhaps the two complement each other. I just thought would be more convenient having a vscode extensions under a vscode configuration.

guilhermespopolin:
Just updated the PR so this only has workspace settings stuff. Gonna open a new one for _Loading State on Debugger_ page feature.

gustaveen (approved):
Great initiative @guilhermespopolin 🙌

.vscode/extensions.json
styrelius: No sure, let's do both 😊

styrelius (approved):
🙌 👌

PR #661, opened 2021-11-04
feature: Loading on debugger page

guilhermespopolin:
## What?

- Adds a loading state to _Debugger_ page that will be displayed when there is a ongoing _create quote_ operation. Additionally, I'm also removing part of the form from the page. 
Screenshot 2021-11-04 at 11 55 48

## Why?

- It's nice to now that something is happening when  _Create Quote_ button is clicked

### DIFF
--- a/src/client/pages/Debugger/components/QuoteData.tsx
+++ b/src/client/pages/Debugger/components/QuoteData.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useState } from 'react'
+import styled from '@emotion/styled'
 import { Form, Formik, FormikProps } from 'formik'
 import { v4 as uuid } from 'uuid'
 import { colorsV3 } from '@hedviginsurance/brand'
@@ -21,6 +22,7 @@ import {
   useMarket,
   Market,
 } from 'components/utils/CurrentLocale'
+import { LoadingDots } from '../../../components/LoadingDots/LoadingDots'
 import { initialSeApartmentValues, SwedishApartment } from './QuoteFormSweden'
 import {
   initialNoHomeValues,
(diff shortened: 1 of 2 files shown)


src/client/pages/Debugger/components/QuoteData.tsx
```diff
@@ -325,63 +332,72 @@ export const QuoteData: React.FC<OfferProps> = ({ sessionToken }) => {
...
-                        label="Start Date (optional)"
-                        placeholder="2020-03-13"
-                        {...props.getFieldProps('startDate')}
-                      />
-                      <InputField
-                        label="Email"
-                        placeholder=""
-                        {...props.getFieldProps('email')}
-                      />
+                      {!props.isSubmitting && (
```
robinandeer, line 335: What is happening here? Are we hiding the input fields while we are submitting? If so, what's the rationale behind that?

robinandeer (commented):
Cool stuff - much easier to review :D

src/client/pages/Debugger/components/QuoteData.tsx
guilhermespopolin: Yes. It comes from a suggestions made by Ellen [here](https://github.com/HedvigInsurance/web-onboarding/pull/659#discussion_r742668805).

> Since you can submit the form by selecting an input field and pressing return key it makes sense that a loading indicator would be visible even if you don't scroll to the bottom of the page

robinandeer: Hmm okey I see the point 👍

robinandeer (approved):
Looks good to me!

PR #663, opened 2021-11-04
GRW-552 / Fix / Add missing "/month" in insurance selector

mehdielket:
## What?

Add missing "/month" missing next to the price on the insurance selector cards

_Referenced ticket(s): [GRW-552]_

[GRW-552]: https://hedvig.atlassian.net/browse/GRW-552?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Screenshot 2021-11-04 at 14 56 49


dendoesit:
We have a problem on screen size 1000px, the text gets too big and goes out of the box. 

image

robinandeer (commented):
Not to be that guy but shouldn't it be "kr/mån"?

Screenshot 2021-11-04 at 17 24 17

That would require more internationalisation work but would be nice to get it figured out :)

mehdielket:
> Not to be that guy but shouldn't it be "kr/mån"?
> 
> Screenshot 2021-11-04 at 17 24 17
> 
> That would require more internationalisation work but would be nice to get it figured out :)

Yeah, the SEK comes from the backend. So I think if we want apps to show kr, we have to do it there.

mehdielket:
> We have a problem on screen size 1000px, the text gets too big and goes out of the box. image

Yeah, we have issues with text coming out of the boxes on portrait iPad size.

I think adding the month thingy is required for launch, I created a ticket to separate ticket to fix iPad issues. I think we don't have a lot of users that will face this issue, so it should be ok to delay it a bit. 

https://hedvig.atlassian.net/browse/GRW-554

robinandeer (approved):
Not sure this is the time to discuss approaches but perhaps we should use built-in internationalization like:

```
const price = 43.4
price.toLocaleString("en", {style: "currency", currency: "USD"})
// "$43.40"
```

Of course we might need some fallback solution when browser support fails.

PR #665, opened 2021-11-05
 fix offer page sidebar

dendoesit:
https://hedvig.atlassian.net/browse/GRW-551

## What?

Removed hasMultipleQuotes conditional as it displayed only when you have 2 quotes.

## Why?

Price on single offer was not displayed.

Currently on dev.hedvigit
image

Now:
image
Multiple quotes 
image


styrelius (dismissed):
Seems like we suddenly have a `package-lock.json` that shouldn't be here 😅 

Also, I assume you're looking into the type checking that's failing.

styrelius:
@dendoesit Would also be great if you could provide a screenshot of what it looks like with your changes 😊

dendoesit:
hmm. not sure what happened to package.lock . remove and added images with after.

robinandeer (dismissed):
Nice catch! 🧖‍♂️

gustaveen:
Hm, wasn't this what @zakkjellin wanted us to not show, the price when it's only one quote? But I might be wrong. Does this fix the logo issue and cropping issue? @dendoesit

dendoesit:
@gustaveen hmm I wasn't aware of that. I thought it's a bug as the design still displays the text with single offer 😿.  will talk to @zakkjellin  about it.

dendoesit:
@gustaveen  asked Zak
Zak  5:16 PM
Hey Dan! We agreed upon that before, but I'm starting to change my mind about that... Would it be a hassle to "revert" back to showing single contracts too? :thinking_face:

src/client/pages/OfferNew/Introduction/Sidebar/index.tsx
```diff
@@ -231,7 +230,7 @@ export const Sidebar: React.FC<Props> = ({
                 />
               </Header>
               <Body>
-                {hasMultipleQuotes && <PriceBreakdown offerData={offerData} />}
+                {<PriceBreakdown offerData={offerData} />}
```
guilhermespopolin, line 233: I think we can remove the curly brackets now as it should always be displayed :)
```suggestion
                <PriceBreakdown offerData={offerData} />
```

guilhermespopolin (dismissed):

robinandeer (approved):
Cool! This is the design in Figma also :)

PR #666, opened 2021-11-05
Tracking - Add current_offer to GTM tracking

gustaveen:
## What?
Update offer events with `current_offer`
Spec here: https://docs.google.com/spreadsheets/d/1IHswc17yIOPkXTGA5aaLsgLRX1lhRvrKaVayzx7RAUo/edit#gid=0

## Why?
Sonny needs to map what offer the user took and then match to what they signed

Screenshot 2021-11-05 at 11 54 53

### DIFF
--- a/src/client/utils/tracking/gtm.ts
+++ b/src/client/utils/tracking/gtm.ts
@@ -34,6 +34,7 @@ type GTMOfferData = {
   has_accident: boolean
   has_travel: boolean
   initial_offer: string
+  current_offer: string
   member_id?: string
 }
 
@@ -121,9 +122,7 @@ export const trackOfferGTM = (
         has_accident: hasAccidentQuote(offerData),
         has_travel: hasTravelQuote(offerData),
         initial_offer: initialOffer,
-        ...(eventName === EventName.SignedCustomer && {
-          signed_offer: contractCategory,
-        }),
+        current_offer: contractCategory,
         ...(switchedFrom && {
           switched_from: getTrackableContractCategory(
             getContractType(switchedFrom),
--- a/src/client/utils/tracking/gtm.ts
+++ b/src/client/utils/tracking/gtm.ts
@@ -103,7 +103,7 @@ export const trackOfferGTM = (
 ) => {
   const contractType = getContractType(offerData)
   const contractCategory = getTrackableContractCategory(contractType)
-  const initialOffer = getInitialOfferFromSessionStorage(contractType)
+  const initialOffer = getInitialOfferFromSessionStorage(contractCategory)
   const grossPrice = Math.round(Number(offerData.cost.monthlyGross.amount))
   const netPrice = Math.round(Number(offerData.cost.monthlyNet.amount))
   try {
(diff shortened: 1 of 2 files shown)


src/client/utils/tracking/tracking.ts
```diff
@@ -152,15 +152,14 @@ export const getTrackableContractCategory = (
...
   const initialtOffer = sessionStorage.getItem('initial_offer')
 
   if (initialtOffer) {
     return initialtOffer
   }
 
-  const contractCategory = getTrackableContractCategory(contractType)
   sessionStorage.setItem('initial_offer', contractCategory)
   return contractCategory
 }
```
robinandeer, line 165: Wait, you pass in `contractCategory` and then return it at the end of the function? 😅
Seems like you could just skip passing it in/returning it altogether.

I would also break it into two functions; `getInitialOfferFromSessionStorage` + `updateInitialOfferInSessionStorage` for example to avoid causing side effects and getting things in the same function.

gustaveen: True you are, a bit lazy of me 😄

src/client/utils/tracking/gtm.ts
```diff
@@ -102,9 +104,15 @@ export const trackOfferGTM = (
...
 ) => {
   const contractType = getContractType(offerData)
-  const initialOffer = getInitialOfferFromSessionStorage(contractType)
+  const contractCategory = getTrackableContractCategory(contractType)
   const grossPrice = Math.round(Number(offerData.cost.monthlyGross.amount))
   const netPrice = Math.round(Number(offerData.cost.monthlyNet.amount))
+
+  let initialOffer = getInitialOfferFromSessionStorage()
+  if (!initialOffer) {
+    setInitialOfferToSessionStorage(contractCategory)
```
robinandeer, line 113: I think this is a perfect case for `useEffect` since it's causing a "side effect". Now we don't really know how many times this function will be called. Not a big deal but I think it would feel more react-y.

robinandeer (commented):
Looks good, I would only consider updating session storage in a `useEffect` hook to make the code more react-y :)

src/client/utils/tracking/gtm.ts
```diff
@@ -120,13 +128,14 @@ export const trackOfferGTM = (
         has_accident: hasAccidentQuote(offerData),
         has_travel: hasTravelQuote(offerData),
         initial_offer: initialOffer,
```
styrelius, line 130: Instead of reassigning `initialOffer` on line 114 - couldn't you just use different variables depending on whether we have `initialOffer` or not? Like
```suggestion
        initial_offer: initialOffer ? initialOffer : contractCategory,
```
```suggestion
        initial_offer: initialOffer || contractCategory,
```
Or am I missing something? Probably (lol) 🙈

gustaveen: I'm open for suggestions!  :) Yes, we can do it like that for sure if it feels clearer  🙌

src/client/utils/tracking/gtm.ts
robinandeer: Whoops, I didn't realise that from the diff :P nvm then!

robinandeer (approved):
Looks good to me now! 💯

PR #667, opened 2021-11-08
Fix / Closing date picker

amandaflod:
## What?

- Create and implement CloseButton component
- Move DateInput and Calendar to separate files
- Add CloseButton and BackgroundClick to DateInput
- New svgs for Arrows and Cross

## Why?

 Be able to close the date picker without choosing a date. 

_Referenced ticket(s): [GRW-48]_

## Screen Recording

https://user-images.githubusercontent.com/59652803/140719653-456b2c94-f898-440a-9098-0cfa283fc445.mov

[GRW-48]: https://hedvig.atlassian.net/browse/GRW-48?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/components/CloseButton/CloseButton.tsx
```diff
@@ -0,0 +1,44 @@
...
+    fill: ${colorsV3.gray700};
+  }
+
+  path {
+    fill: ${colorsV3.gray900};
+  }
+`
+
+type Props = {
+  onClose: () => void
```
robinandeer, line 35: Doesn't it make more sense to call this `onClick`?

I would prefer not creating custom props at all and just pass all standard button props so it can be used like any other button. Like this:

```
type Props = React.HTMLProps<HTMLButtonElement>

export const CloseButton = (props: Props) => {
   return (
     <Button {...props}>
       <Cross />
     </Button>
   )
 }
```

src/client/components/DateInput/Calendar.tsx
```diff
@@ -0,0 +1,244 @@
...
+
+export const Calendar: React.FC<DayzedCalendarProps> = ({
+  calendars,
+  getBackProps,
+  getForwardProps,
+  getDateProps,
+}) => {
+  const [weekdayNamesShort, setWeekdayNamesShort] = React.useState<string[]>(
+    [] as string[],
+  )
```
robinandeer, line 139: I don't think you need these type assertions:

```suggestion
  const [weekdayNamesShort, setWeekdayNamesShort] = React.useState<string[]>([])
```

src/client/components/icons/Cross.tsx
```diff
@@ -2,16 +2,15 @@ import React from 'react'
 
 export const Cross = () => (
```
robinandeer, line 3: nitpick: while we are at it you could consider wrapping these in the `IconRoot` component to standardise how we use icons.

src/client/components/icons/BackArrow.tsx
```diff
@@ -1,7 +1,18 @@
...
+    height="24"
+    viewBox="0 0 24 24"
+    fill="none"
+    xmlns="http://www.w3.org/2000/svg"
+  >
+    <path
+      fillRule="evenodd"
+      clipRule="evenodd"
+      d="M9.06066 12L16.5303 4.53033L15.4697 3.46967L6.93934 12L15.4697 20.5303L16.5303 19.4697L9.06066 12Z"
+      fill="#777777"
```
robinandeer, line 15: This value can be imported from `@hedviginsurance/brand` package :)

robinandeer (commented):
Great work @amandaflod! Certainly not the easiest code to get into 😅 

Solution seems to work great when I tested it -- just have a look through my comments

guilhermespopolin (commented):
Good stuff :)

src/client/components/CloseButton/CloseButton.tsx
styrelius: I don't think changing `className` is a pattern we should use for changing styles on a component though, but that's a discussion we could have outside of this forum maybe 😄

src/client/components/icons/BackArrow.tsx
amandaflod: I don't think we have a color in the brand package that is the same, should I just go with the closest one or is there another way around that?

src/client/components/CloseButton/CloseButton.tsx
styrelius: And not surprisingly @robinandeer I don't agree with you that it would be better to just pass all standard props 🙈 This is a component that does a very specific thing, which is nice IMO since then there are no question marks around when/where/how to use it. But therefore there shouldn't be a need to us it like "any other button"? Or what would be gained from having that possibility?

src/client/components/icons/BackArrow.tsx
styrelius: Maybe ask Zak if we should update brand colors or if you can use a color that's in the brand package? Because he's the one who gave you these icons right?

src/client/components/CloseButton/CloseButton.tsx
robinandeer: haha not surprised @styrelius ;)
I think it will be easier to reason about these things as we move to a design system. So it's fine to keep this very specific and just pass a single prop if that's all we need.

I do like to reuse existing patterns/conventions when possible though. So I would still suggest renaming the prop to `onClick`. We want something to happen when a user clicks the `CloseButton`, right? Not something to happen when a user closes the `CloseButton`... But that's a nitpick!

styrelius: Oh, I definitely think the prop should be called `onClick` 🤝

styrelius: Sorry that I was unclear - it was more the part of having access to setting all the `button` props/attributes from the parent component that I'm opposed :)

src/client/components/icons/Cross.tsx
amandaflod: Fixed that now on the three icons I used :)

src/client/components/icons/BackArrow.tsx
amandaflod: I checked with Zak and he said it was just a placeholder for us to override with whatever color is supposed to be there. I checked how we have done with other Icons and for many of them we have just left fill out completely or set it to "none" in the icon components themselves so I did that now. The color is set then where the icon is used... Or should we rethink that?

src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
```diff
@@ -325,15 +316,57 @@ const DateForm: React.FC<{
...
+    <StyledDateInput
+      open={datePickerOpen}
+      setOpen={setDatePickerOpen}
+      date={dateValue || new Date()}
+      setDate={setDate}
+      hasCurrentInsurer={hasCurrentInsurer(quote)}
+      modal={Boolean(modal)}
+    />
+  )
+}
```
robinandeer, line 368: nit: This seems so simple we don't really need the abstraction -- could just use `StyledDateInput` directly inside `DateForm`

src/client/components/CloseButton/CloseButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+  :hover path {
+    fill: ${colorsV3.gray700};
+  }
+
+  path {
+    fill: ${colorsV3.gray900};
+  }
+`
+
+type Props = React.HTMLProps<HTMLButtonElement>
```
robinandeer, line 34: Haha I won you over with this one ;) 

Like I said in the other thread - I'm happy with either solution as long as we move to a design system in the near future.

robinandeer (dismissed):
I think the code look good now from my end 🎉

src/client/components/icons/BackArrow.tsx
gustaveen: Exactly @amandaflod ,colors should be set where they are used 🙌 In the case of this color, @robinandeer added `gray600: '#777777'` a few days ago since Zak wanted another tint. So now we have the full scale of grays (in Brand repo). 
```
  gray100: '#fafafa',
  gray200: '#f2f2f2',
  gray300: '#eaeaea',
  gray500: '#aaaaaa',
  gray600: '#777777',
  gray700: '#505050',
  gray800: '#272727',
  gray900: '#121212',
```

src/client/components/CloseButton/CloseButton.tsx
```diff
@@ -0,0 +1,42 @@
...
+  top: 1rem;
+  right: 1rem;
+  padding: 0;
+  display: flex;
+  align-items: center;
+  text-align: center;
+  border-radius: 50%;
+  border: none;
+  cursor: pointer;
+  background: ${colorsV3.white};
```
gustaveen, line 19: super nit: consider using `background-color:` if we just want to change the background color and not any other properties like `background-image, background-position` etc

gustaveen (dismissed):
Splendid work Amanda! As said, not the easiest thing to jump on 🙌 Very nice improvement!

src/client/components/CloseButton/CloseButton.tsx
amandaflod: Hahah yes Ellen and I came up with a solution that would make both you @robinandeer and @styrelius happy ;)

robinandeer: Oh right, since you throw away all the other props 😅 very diplomatic

styrelius (approved):
Great job Amanda!! 🥳 🤸‍♀️

src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
```diff
@@ -325,15 +319,58 @@ const DateForm: React.FC<{
...
+//     <StyledDateInput
+//       open={datePickerOpen}
+//       setOpen={setDatePickerOpen}
+//       date={dateValue || new Date()}
+//       setDate={setDate}
+//       hasCurrentInsurer={hasCurrentInsurer(quote)}
+//       modal={Boolean(modal)}
+//     />
+//   )
+// }
```
robinandeer, line 372: Oh no! But we can clean this up later @amandaflod

PR #673, opened 2021-11-09
Fix hidden Hedvig logo on Safari

gustaveen:
## What?
The problem is that with `display: flex` on the parent of a responsive SVG, the SVG does not take up any height. Resorting to good ol' `display: block` does the trick. The `font-size: 0` hack as made popular by @robinandeer takes care of unwanted white-space below the SVG 🙌

## Why?
Logo was not visible in Safari

### Before
Screenshot 2021-11-09 at 17 23 20

### After
Screenshot 2021-11-09 at 17 24 39

### DIFF
--- a/src/client/components/TopBar.tsx
+++ b/src/client/components/TopBar.tsx
@@ -60,8 +60,10 @@ const CenteredContainer = styled(Container)`
 `
 
 const LogoLink = styled.a`
+  display: block;
   color: inherit;
-  display: flex;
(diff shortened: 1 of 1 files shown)


robinandeer (approved):
Nice catch! 😌

I wonder how we can avoid adding the same hack-comment everywhere 🤔. Setting global styles goes against emotion conventions... But maybe it should be considered as a style reset so it's okey.

gustaveen:
Good point @robinandeer This issue can be resolved with setting a height on the svg as well. Let's sleep on that 🙌

PR #674, opened 2021-11-09
GRW-563: removed horizontal scroll from modal

guilhermespopolin:
## What?

1. Move _lock scroll on body_ logic to a custom hook (`useLockBodyScroll`)
2. Remove weird horizontal scroll from modal:

### Before
image

### After
Screenshot 2021-11-09 at 17 45 07

## Why?

1. I intent to use the same hook to solve the same problem on offer page:
Screenshot 2021-11-09 at 17 44 29

_Referenced ticket(s): [GRW-563](https://hedvig.atlassian.net/browse/GRW-563)_

### DIFF
--- a/src/client/components/ModalNew.tsx
+++ b/src/client/components/ModalNew.tsx
@@ -15,13 +15,19 @@ export interface ModalProps {
 
 const Wrapper = styled(motion.div)`
   position: fixed;
-  inset: 0;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
   z-index: 3000;
 `
 
 const Background = styled(motion.div)`
   position: fixed;
-  inset: 0;
+  top: 0;
+  right: 0;
+  bottom: 0;
+  left: 0;
   background-color: rgba(25, 25, 25, 0.4);
 `
 
--- a/src/client/components/ModalNew.tsx
+++ b/src/client/components/ModalNew.tsx
@@ -3,7 +3,7 @@ import { colorsV3 } from '@hedviginsurance/brand'
 import { motion } from 'framer-motion'
 import React from 'react'
 
-import useLockBodyScroll from '../utils/hooks/useLockBodyScroll'
+import { useLockBodyScroll } from '../utils/hooks/useLockBodyScroll'
 
 import { Cross } from './icons/Cross'
 export interface ModalProps {
--- a/src/client/components/ModalNew.tsx
+++ b/src/client/components/ModalNew.tsx
@@ -2,8 +2,10 @@ import styled from '@emotion/styled'
 import { colorsV3 } from '@hedviginsurance/brand'
 import { motion } from 'framer-motion'
 import React from 'react'
-import { Cross } from './icons/Cross'
 
+import useLockBodyScroll from '../utils/hooks/useLockBodyScroll'
+
+import { Cross } from './icons/Cross'
 export interface ModalProps {
   isVisible: boolean
   dynamicHeight?: boolean
@@ -13,23 +15,13 @@ export interface ModalProps {
 
 const Wrapper = styled(motion.div)`
   position: fixed;
-  width: 100%;
-  height: 100%;
-  left: 0;
-  top: 0;
-  bottom: 0;
-  right: 0;
+  inset: 0;
   z-index: 3000;
 `
 
 const Background = styled(motion.div)`
   position: fixed;
-  width: 100%;
-  height: 100%;
-  left: 0;
-  top: 0;
-  bottom: 0;
-  right: 0;
+  inset: 0;
   background-color: rgba(25, 25, 25, 0.4);
 `
 
@@ -39,24 +31,23 @@ interface ModalContainerProps {
 
 const ModalContainer = styled(motion.div)<ModalContainerProps>`
   position: relative;
+  left: 50%;
+  top: 50%;
   width: 100%;
   max-width: 56rem;
   max-height: 100vh;
+
   ${(props) =>
     !props.dynamicHeight &&
     `
-  height: 100%;
   min-height: 25rem;
   max-height: 56rem;`}
   background: ${colorsV3.white};
-  border-radius: 9px;
+  border-radius: 8px;
   position: absolute;
-  left: 50%;
-  top: 50%;
   transform: translateX(-50%) translateY(-50%);
   box-shadow: 0 0 14px rgba(0, 0, 0, 0.06);
-  box-sizing: border-box;
-  overflow-x: scroll;
+  overflow-y: auto;
 
   @media (max-height: 900px) {
     max-height: calc(100vh - 2rem);
@@ -115,14 +106,7 @@ export const Modal: React.FC<ModalProps> = ({
   onClose,
   children,
 }) => {
-  React.useEffect(() => {
-    if (isVisible) {
-      document.body.style.overflow = 'hidden'
-    }
-    return () => {
-      document.body.style.overflow = 'visible'
-    }
-  }, [isVisible])
+  useLockBodyScroll({ lock: isVisible })
 
   return (
     <Wrapper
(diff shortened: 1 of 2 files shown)


src/client/components/ModalNew.tsx
```diff
@@ -115,14 +106,7 @@ export const Modal: React.FC<ModalProps> = ({
...
 }) => {
-  React.useEffect(() => {
-    if (isVisible) {
-      document.body.style.overflow = 'hidden'
-    }
-    return () => {
-      document.body.style.overflow = 'visible'
-    }
-  }, [isVisible])
+  useLockBodyScroll({ lock: isVisible })
```
robinandeer, line 109: Like I mentioned to @gustaveen, I think it's a nice touch to name custom hooks that case side effects like:

`useLockBodyScrollEffect`

Just to make it clear without having to look at the implementation.

src/client/utils/hooks/useLockBodyScroll.ts
```diff
@@ -0,0 +1,20 @@
+import { useLayoutEffect } from 'react'
+
+type UseLockBodyScrollConfig = { lock: boolean }
+
+function useLockBodyScroll({ lock }: UseLockBodyScrollConfig = { lock: true }) {
```
robinandeer, line 5: Another thing which is hard to know since we don't seem to have a guide in Notion with our coding conversations 😅

However, we prefer **arrow function** definitions.

robinandeer, line 5: Also can't you write like this?

```suggestion
function useLockBodyScroll({ lock = true }: UseLockBodyScrollConfig) {
```

doesn't it make more sense that the default is "false" btw?

src/client/utils/hooks/useLockBodyScroll.ts
```diff
@@ -0,0 +1,20 @@
...
+  useLayoutEffect(() => {
+    const previousOverflowStyle = window.getComputedStyle(document.body)
```
robinandeer, line 7: This is a nice touch!

Just out of curiosity, why is it not enough doing?

```
    const previousOverflowStyle = document.body.style.overflow
```

robinandeer (commented):
Very nice tweak! I realize we should do more cross browser/OS testing to catch these visual bugs

src/client/components/ModalNew.tsx
```diff
@@ -13,23 +15,13 @@ export interface ModalProps {
 
 const Wrapper = styled(motion.div)`
   position: fixed;
-  width: 100%;
-  height: 100%;
-  left: 0;
-  top: 0;
-  bottom: 0;
-  right: 0;
+  inset: 0;
```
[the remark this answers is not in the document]
guilhermespopolin: 😢 I guess I'm trusting on CSS tutorials practices more than I should.

src/client/utils/hooks/useLockBodyScroll.ts
guilhermespopolin: I see. I actually prefer function declarations but if that's the common practice among the team I'm ok with it :)

src/client/components/ModalNew.tsx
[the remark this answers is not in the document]
robinandeer: haha that's why it's good to have different view points - I think I'm sometimes too conservative ;)

src/client/utils/hooks/useLockBodyScroll.ts
guilhermespopolin: 🤔 The idea of use `true` as the default value is because my initial idea for the api of this hook was:

- Create a hook that when used inside a component A would make A cause a block of scroll on the body element **while A is mounted**, which means everything would be back to normal as soon as **A is unmounted**

This makes sense if we start to think about the type of components that would probably use a hook like that: _Modal_s, for example. When a _Modal_ is opened it normally capture all of the user interactions so no scroll should (normally) be taken in place in the background. But as soon as the modal is resolved everything should be back to normal.

I like that approach very much but there's a corner case:

- Sometimes we want to keep a component like a _Modal_ mounted even though it's not being displayed..

In that case the initial API of this hook wouldn't work. That's why it also accepts a way to _control_ when scroll should be blocked or not. So if you want to use it blocking/unblocking based on mount/unmount just call it inside your component. However if you want to determine when it should block/unblock, just call it providing a proper value.

I'm sorry for the long post. Here, take a potato:
image

src/client/utils/hooks/useLockBodyScroll.ts
guilhermespopolin: Because if you try to read a style property value of an element that hasn't been implicitly defined you may get as the return something that is actually not real. For example:

Consider a viewport which `width` is `200px`. If you render a `<div></div>` without any explicitly styling on that viewport and try to read it's `width` you'll get:

- `""` if you do it like this: `.style.width`
- `"200px"` if you do it like this: `window.getComputedStyle(element).width`, because a `div` is a block element that will take the whole `width` of its container. So it's **computed style** when it comes about `width` will be `200px`.

src/client/components/ModalNew.tsx
guilhermespopolin: 🤔 Dunno about that. I mean, adding 'effect' in the name kind of suggest a pattern where we "import" how something was implemented (implementation details) to its name, which I think it's not usually a good thing.

Or maybe I misunderstood your point and your suggestion is based on a concern of yours that is generally good for a consumer component to know that a custom hook is causing a side effect?! Is that the case?

src/client/utils/hooks/useLockBodyScroll.ts
robinandeer: Right, I understand that part. What I don't understand 💯 is why we don't just reset to the previous value? If there wasn't any previous value we just reset to "nothing" and let the browser recalculate what the computed value should be, no?

If we calculate the style before and store it we will add a new style to the element when we run the clean-up function. Seems better to delegate back to the browser.

src/client/components/ModalNew.tsx
robinandeer: Yes, exactly! Just to give an indication to the one reading the code that when this hook is called some side effect will happen :)

guilhermespopolin: 🤔 I see. The only "problem" is it feels we would need to think when add or not the "effect" word in the name. Take `useQuoteIds` for example. It would feel odd if its name was `useQuoteIdsEffect`, right?! What do you think?

robinandeer: Aha good point - I agree that the name of that hook is not optimal.

I would name it something like `usePersistedQuoteIds`.

If it's a hook that _only_ does a side effect though I think it's still pretty straight forward. But we should write all these conventions down so totally fine to leave as is for this PR.

src/client/utils/hooks/useLockBodyScroll.ts
robinandeer: Haha okey, I'll try to be more clear:

```
const useLockBodyScroll = ({ lock = true }: UseLockBodyScrollConfig) => {
  useLayoutEffect(() => {
   const previousOverflowStyle = document.body.style.overflow

    if (lock) {
      document.body.style.overflow = 'hidden'
    }

    return () => {
      document.body.style.overflow = previousOverflowStyle
    }
  }, [lock])
}
```

This is a nit pick just to be clear. I don't really see a clear issue with your solution. Rather that this ⬆️ would be slightly simpler.

guilhermespopolin: Got it. 🤔 It seems it also works. This implementation is heavily based on [this](https://usehooks.com/useLockBodyScroll/) that's why the `getComputedStyle`. I couldn't see an edge case either just added the `getComputedStyle` because, you know, maybe the author did 😄

But, since it's working I gonna change to the proposed one.

robinandeer: Haha with that reference I would probably also just go with the "computed style" solution 😅 so whatever feels best for you -- but great that we point to a reference!

guilhermespopolin: 😆 😆 😆 😆

robinandeer (approved):
Code looks good to me! Haven't tested locally though :)

PR #676, opened 2021-11-10
fix(Tooltip): added min width

guilhermespopolin:
## What?

- Added `min-width` to `Tooltip`

## Why?

- So it not looks too narrow on mobile
Screenshot 2021-11-10 at 15 49 38
Screenshot 2021-11-10 at 15 47 56

_Referenced ticket(s): [GRW-565](https://hedvig.atlassian.net/browse/GRW-565)_

### DIFF
--- a/src/client/components/Tooltip/Tooltip.tsx
+++ b/src/client/components/Tooltip/Tooltip.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState, useEffect, useRef } from 'react'
 import styled from '@emotion/styled'
 import { colorsV3 } from '@hedviginsurance/brand'
 import { motion } from 'framer-motion'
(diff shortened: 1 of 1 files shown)


src/client/components/Tooltip/Tooltip.tsx
```diff
@@ -84,11 +84,11 @@ const TooltipText = styled.p`
...
 export const Tooltip: React.FC<TooltipProps> = ({ body }) => {
-  const ref = React.useRef<HTMLDivElement>(null)
-  const [isVisible, setVisible] = React.useState(false)
+  const ref = useRef<HTMLDivElement>(null)
+  const [isVisible, setVisible] = useState(false)
 
-  React.useEffect(() => {
-    if (isVisible === false) return
+  useEffect(() => {
+    if (!isVisible) return
```
robinandeer, line 91: nit: is this really more readable than what we had before? ;)

robinandeer (approved):
Nice addition!

src/client/components/Tooltip/Tooltip.tsx
guilhermespopolin: 🤔 I think so. I mean, it's almost like reading plain english: "if not visible". But I can revert it, no problem at all.

robinandeer: aha interesting 😅 I read it as "if not _is_ visible" which makes me confused -- while the other one is "if is visible is false". But don't change it unless other people feel the same as me hehe

gustaveen: I think when we use the `is` prefix for booleans, I have no trouble reading it like `!isVisible`. So I approve 🤗

gustaveen (approved):
Good job! 🙌

PR #677, opened 2021-11-10
Fix/Tabs overflow + some other small design tweaks

robinandeer:
## What?

- Fix tabs overflow by using "overflow-wrap"
    I can't claim to understanding how/why this works. However, according to [this article](https://css-tricks.com/almanac/properties/o/overflow-wrap/) that property is often used to solve text-related layout issues.
- Safari: fix underline floating on top of background
    I believe Safari was adding a default 2px margin to buttons 🙃🔫
- Correct padding between text and underline (4px)

## Why?

We want thing to look crisp 💎 
...and the layout was broken before.

## Demo

https://user-images.githubusercontent.com/1220232/141194813-98047944-b29b-46e0-a9cf-fa8bc45563b9.mov

### DIFF
--- a/src/client/pages/OfferNew/Tabs/Tab.tsx
+++ b/src/client/pages/OfferNew/Tabs/Tab.tsx
@@ -15,6 +15,7 @@ const TabContainer = styled.button<{ selected?: boolean }>`
   padding: 0 1rem 0.25rem 0;
   position: relative;
   flex-shrink: 0;
+  overflow-wrap: anywhere;
 
   &:hover {
     color: ${colorsV3.gray900};
--- a/src/client/pages/OfferNew/Tabs/Tab.tsx
+++ b/src/client/pages/OfferNew/Tabs/Tab.tsx
@@ -16,6 +16,7 @@ const TabContainer = styled.button<{ selected?: boolean }>`
   position: relative;
   flex-shrink: 0;
   overflow-wrap: anywhere;
+  margin: 0;
 
   &:hover {
     color: ${colorsV3.gray900};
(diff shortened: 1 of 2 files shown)


dendoesit (dismissed):
great stuff 👍 
I guess even with flex-shrink:0 if you had too many tabs they started to pile up and broke layout.

src/client/pages/OfferNew/Tabs/Tab.tsx
```diff
@@ -12,9 +12,11 @@ const TabContainer = styled.button<{ selected?: boolean }>`
   display: flex;
   font-size: 1.25rem;
   line-height: 1.4;
-  padding: 0 1rem 0.25rem 0;
+  padding: 0 1rem calc(0.25rem + 2px) 0;
```
styrelius, line 15: So is this due to the `button` margin thingy in Safari? Or why is `2px` added to the padding? If this is what we have to do I think a comment might be good to explain why we have this 😇

robinandeer: Oh that's just a coincidence but I can see the confusion.

These 2px is a reference to the height of the underline component. Perhaps we can break that out into a named constant.

styrelius:
Also, I can't see the tabs overflowing anywhere in any screen sizes, neither in Chrome nor Safari. 🤔  Maybe you could add a screenshot or recording of the issues that this PR fixes? Like a "before" in addition to your "after" recording 😀

robinandeer:
Good point @styrelius -- here is the before where you can see the issue:

Screenshot 2021-11-10 at 14 39 52

src/client/pages/OfferNew/Tabs/Tab.tsx
styrelius: That would be great! 👍

styrelius (approved):
Thanks for the screenshot 👌 In a way it seems weird to fix this _layout_ issue by doing something to the `Tab` component very deep down the tree but as a quick fix I guess this is a pragmatic approach.

PR #679, opened 2021-11-12
Fix - Update close icon checkout

amandaflod:
## What?

- Change from back arrow to CloseButton component to close the checkout drawer. 
- Update CloseButton component to make it even more reusable
- Remove commented out code left over from previous PR 😇

## Why?

Being consistent with our design and providing a better experience 💅

_Referenced ticket(s): [GRW-415]_

Screenshot 2021-11-12 at 16 36 08

[GRW-415]: https://hedvig.atlassian.net/browse/GRW-415?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/pages/OfferNew/Introduction/Sidebar/DiscountCodeModal.tsx
+++ b/src/client/pages/OfferNew/Introduction/Sidebar/DiscountCodeModal.tsx
@@ -110,7 +110,7 @@ export const DiscountCodeModal: React.FC<Props> = ({
       >
         <Title>{textKeys.SIDEBAR_ADD_DISCOUNT_HEADLINE()}</Title>
         <Paragraph>{textKeys.SIDEBAR_ADD_DISCOUNT_BODY()}</Paragraph>
-        <CloseButton onClick={close} />
+        <CloseButton onClick={close} position="right" />
         <Formik
           validateOnBlur
           validationSchema={discountSchema}
--- a/src/client/pages/OfferNew/Introduction/Sidebar/DiscountCodeModal.tsx
+++ b/src/client/pages/OfferNew/Introduction/Sidebar/DiscountCodeModal.tsx
@@ -44,6 +44,11 @@ const Container = styled(motion.div)`
   border: 1px solid ${colorsV3.gray300};
   position: relative;
 `
+const ButtonWrapper = styled.div`
+  position: absolute;
+  top: 1rem;
+  right: 1rem;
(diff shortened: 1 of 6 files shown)


src/client/pages/OfferNew/Introduction/Sidebar/DiscountCodeModal.tsx
```diff
@@ -44,6 +44,11 @@ const Container = styled(motion.div)`
   border: 1px solid ${colorsV3.gray300};
   position: relative;
 `
+const ButtonWrapper = styled.div`
```
guilhermespopolin, line 47: Gonna take this opportunity and try to clarify a question if mine:

I understand that you've removed _position_ related styles from a shared component which IMHO is very good since we're increasing reusability of `CloseButton`. My question tho is that if the team has a preference among the following options:

1) Your Solution: Apply specify styling to a component by using a wrapper
2) Use emotion to customize a component:
```javascript
import CloseButton from '<close-button-location>'

const Button = styled(CloseButton)`
   position: absolute;
   top: 1rem;
   right: 1rem;
`
```

@styrelius @robinandeer @gustaveen @dendoesit

guilhermespopolin (approved):

src/client/pages/OfferNew/Introduction/Sidebar/DiscountCodeModal.tsx
amandaflod: Yes! I thought about that as well but I'm not really sure how the rest of the team thinks about this. Maybe we can bring this up in regard to the style guides @styrelius is working on atm? :D

styrelius: @guilhermespopolin We don't have a preferred way (yet) but I think this is something we could discuss during Tech Forum 😀 

I think generally components shouldn't care about where they're positioned, so a wrapper for laying things out makes more sense to me. But I'm not _against_ extending styled components like in your example either. But yes, let's talk about it later (we should probably have a place, like a page in Notion, where we can add tech things like these that we want to discuss).

styrelius (approved):
So nice that we can reuse this lovely button in all these places now! 🤩

PR #682, opened 2021-11-12
GRW fix(Checkout): block background scroll when opened

guilhermespopolin:
## What?

- Prevent background scrolling when _Checkout_ is opened. To achieve that, I'm using `useLockBodyScroll` hook added on [this PR](https://github.com/HedvigInsurance/web-onboarding/pull/674)

## Why?

- Background content shouldn't be able to be scrolled while _Checkout_ is opened.

### Before
image

### After
Screenshot 2021-11-12 at 11 27 30

_Referenced ticket(s): [GRW-585](https://hedvig.atlassian.net/browse/GRW-585)_


robinandeer (approved):
Very nice 👌

PR #683, opened 2021-11-13
GRW-58 / Feature / Offer Page Debugger for new Onboarding API

robinandeer:
## What?

Co-authored with @dendoesit.

Implement a new Offer Page Debugger that uses the new QuoteCart API: `/se-en/new-member/offer-debugger`.

The goal is _not_ to improve the debugger but the setup is simplified quite a bit:
- Create quote bundles rather than multiple quotes per request
- Skip session handling (we worry about that only in the real app, if we need to)
- Implement `ErrorBoundary` to deal with errors in a React-y way (I can revert this if no one likes it 😜)
- Skip UUID generation of quote IDs since API can return them for us 🎉 

> Currently, the debugger passes the quote cart ID to the offer page as a query parameter. That will work okay for now but we might need to go back and revise this. Let's see.

## Why?

The new debugger is implemented as a copy of the old one. This will make it easy to remove the old one when we are ready to do so. It also makes it easy to skip complicated stuff like dealing with auth tokens and sessions that we don't need to do with the new API.

A new dependency is added `react-error-boundary` to handle errors in a more React-y way. There seems to be some conflict with Sentry that catches errors without letting them bubble up normally. Does anyone know how that works? @styrelius @gustaveen?

**Referenced ticket(s): [GRW-583]**

## Screenshots / recordings

Screenshot 2021-11-15 at 10 18 35

## Review app

Test it live: https://web-onboardi-grw-583-fe-hbvsqd.herokuapp.com/se-en/new-member/offer-debugger

[GRW-583]: https://hedvig.atlassian.net/browse/GRW-583?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/graphql/CreateQuoteBundle.graphql
+++ b/src/client/graphql/CreateQuoteBundle.graphql
@@ -1,5 +1,5 @@
-mutation CreateQuoteBundle($id: ID!, $quotes: [QuoteData!]!) {
-  quoteCart_createQuoteBundle(id: $id, input: { quoteData: $quotes }) {
+mutation CreateQuoteBundle($quoteCartId: ID!, $quotes: [QuoteData!]!) {
+  quoteCart_createQuoteBundle(id: $quoteCartId, input: { quoteData: $quotes }) {
     ... on QuoteCart {
       id
       bundle {
--- a/src/client/graphql/CreateQuoteBundle.graphql
+++ b/src/client/graphql/CreateQuoteBundle.graphql
@@ -1,0 +1,17 @@
+mutation CreateQuoteBundle($id: ID!, $quotes: [QuoteData!]!) {
+  quoteCart_createQuoteBundle(id: $id, input: { quoteData: $quotes }) {
+    ... on QuoteCart {
+      id
+      bundle {
+        quotes {
+          id
+        }
+      }
+    }
+    ... on CreateQuoteBundleError {
+      limits {
+        code
+      }
+    }
+  }
+}
--- a/src/client/graphql/QuoteCart.graphql
+++ b/src/client/graphql/QuoteCart.graphql
@@ -1,10 +1,10 @@
-query QuoteCart($id: ID!) {
+query QuoteCart($id: ID!, $locale: Locale!) {
   quoteCart(id: $id) {
     id
     bundle {
       possibleVariations {
         id
-        tag(locale: sv_SE)
+        tag(locale: $locale)
         bundle {
           bundleCost {
             monthlyGross {
--- a/src/client/graphql/QuoteCart.graphql
+++ b/src/client/graphql/QuoteCart.graphql
@@ -1,0 +1,96 @@
+query QuoteCart($id: ID!) {
+  quoteCart(id: $id) {
+    id
+    bundle {
+      possibleVariations {
+        id
+        tag(locale: sv_SE)
+        bundle {
+          bundleCost {
+            monthlyGross {
+              amount
+              currency
+            }
+          }
+        }
+      }
+      bundleCost {
+        monthlyGross {
+          amount
+          currency
+        }
+        monthlyNet {
+          amount
+          currency
+        }
+        monthlyDiscount {
+          amount
+          currency
+        }
+        freeUntil
+      }
+      quotes {
+        id
+        currentInsurer {
+          id
+          displayName
+          switchable
+        }
+        price {
+          amount
+          currency
+        }
+        firstName
+        lastName
+        ssn
+        birthDate
+        quoteDetails {
+          __typename
+          ... on SwedishApartmentQuoteDetails {
+            street
+            zipCode
+            householdSize
+            livingSpace
+            type
+          }
+          ... on SwedishHouseQuoteDetails {
+            street
+          }
+          ... on SwedishAccidentDetails {
+            street
+          }
+        }
+        startDate
+        expiresAt
+        email
+        dataCollectionId
+        typeOfContract
+        initiatedFrom
+      }
+    }
+    campaign {
+      incentive {
+        ... on MonthlyCostDeduction {
+          amount {
+            amount
+            currency
+          }
+        }
+        ... on PercentageDiscountMonths {
+          percentageDiscount
+          quantity
+        }
+      }
+      code
+      owner {
+        displayName
+        id
+      }
+      expiresAt
+    }
+    checkoutMethods
+    checkout {
+      status
+    }
+  }
+}
--- a/src/client/pages/OfferDebugger/Debugger.tsx
+++ b/src/client/pages/OfferDebugger/Debugger.tsx
@@ -1,6 +1,7 @@
-import React from 'react'
+import React, { useEffect, useCallback } from 'react'
 import styled from '@emotion/styled'
 import { colorsV3 } from '@hedviginsurance/brand'
+import { ErrorBoundary, FallbackProps } from 'react-error-boundary'
 import { Button } from 'components/buttons'
 import { useCreateOnboardingSessionMutation } from 'data/graphql'
 import { useMarket } from 'components/utils/CurrentLocale'
@@ -18,6 +19,26 @@ const Row = styled.div`
   padding-bottom: 1rem;
 `
 
+const ErrorFallback: React.FC<FallbackProps> = ({
+  error,
+  resetErrorBoundary,
+}) => (
+  <div>
+    <h2>Something went wrong 😔</h2>
+    <div>
+      <p>Error message:</p>
+      <pre style={{ whiteSpace: 'pre-wrap' }}>{error}</pre>
+    </div>
+    <Button
+      background={colorsV3.red500}
+      foreground={colorsV3.gray100}
+      onClick={resetErrorBoundary}
+    >
+      👉 Try starting over by creating a new session!
+    </Button>
+  </div>
+)
+
 export const Debugger: React.FC = () => {
   const market = useMarket()
   const locale = useCurrentLocale().isoLocale
@@ -27,27 +48,39 @@ export const Debugger: React.FC = () => {
   ] = useCreateOnboardingSessionMutation()
   const sessionId = data?.onboardingSession_create
 
-  const handleClickNewSession = async () => {
+  const createNewSession = useCallback(async () => {
     await createOnboardingSession({ variables: { country: market, locale } })
-  }
+  }, [createOnboardingSession, market, locale])
+
+  useEffect(() => {
+    // create initial onboarding session
+    createNewSession()
+  }, [createNewSession])
 
   return (
     <Wrapper>
       <h1>Offer Page Debugger</h1>
 
       <Row>
-        <h3>Session</h3>
-        <Button background={colorsV3.purple900} onClick={handleClickNewSession}>
+        <h3>Session: {sessionId}</h3>
+        <Button
+          background={colorsV3.purple500}
+          foreground={colorsV3.gray900}
+          onClick={createNewSession}
+        >
           Create new session
         </Button>
-
-        {sessionId && <p>Active session: {sessionId}</p>}
       </Row>
 
       {sessionId && (
         <Row>
-          <h3>Offer</h3>
-          <QuoteData sessionId={sessionId} />
+          <ErrorBoundary
+            FallbackComponent={ErrorFallback}
+            onReset={createNewSession}
+          >
+            <h3>Offer</h3>
+            <QuoteData sessionId={sessionId} />
+          </ErrorBoundary>
         </Row>
       )}
     </Wrapper>
--- a/src/client/pages/OfferDebugger/Debugger.tsx
+++ b/src/client/pages/OfferDebugger/Debugger.tsx
@@ -1,6 +1,6 @@
 import React from 'react'
 import styled from '@emotion/styled'
-import { colorsV2, colorsV3 } from '@hedviginsurance/brand'
+import { colorsV3 } from '@hedviginsurance/brand'
 import { FetchResult } from '@apollo/client'
 import { Button } from 'components/buttons'
 import {
@@ -13,6 +13,7 @@ import { QuoteData } from './components/QuoteData'
 
 const Wrapper = styled.div`
   max-width: 900px;
+  padding: 1rem;
   margin: 0 auto;
   color: ${colorsV3.gray100};
 `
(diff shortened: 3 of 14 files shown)


src/client/pages/OfferDebugger/Debugger.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { ErrorBoundary, FallbackProps } from 'react-error-boundary'
+import { Button } from 'components/buttons'
+import { useCreateOnboardingSessionMutation } from 'data/graphql'
+import { useMarket } from 'components/utils/CurrentLocale'
+import { useCurrentLocale } from 'l10n/useCurrentLocale'
+import { QuoteData } from './components/QuoteData'
+
+const Wrapper = styled.div`
+  max-width: 900px;
+  padding: 1rem;
```
robinandeer, line 13: I've been wanting to fix this forever 😅

src/client/pages/OfferDebugger/components/QuoteData.tsx
```diff
@@ -0,0 +1,358 @@
...
+
+      await refetch()
+    } catch (error) {
+      setSubmitError(error)
+    }
+  }
+
+  if (submitError) {
+    throw submitError
+  }
```
robinandeer, line 240: It seems like Sentry catches all errors and just logs them to the console which isn't super helpful when you want to use error boundaries. That's why I had to come up with this workaround.

src/client/pages/OfferDebugger/components/QuoteFormSweden.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export const initialSeApartmentValues = {
+  firstName: 'Sven',
+  lastName: 'Svensson',
+  currentInsurer: '',
+  birthDate: '1995-09-29',
+  ssn: '199509291234',
+  startDate: '',
+  email: 'sven.svensson@hedvig.com',
+  swedishApartmentData: {
```
robinandeer, line 14: Here's one of the few changes in the new API: `swedishApartment` => `swedishApartmentData`

src/client/pages/OfferDebugger/components/QuoteFormSweden.tsx
```diff
@@ -0,0 +1,65 @@
...
+    street: 'Storgatan 1',
+    zipCode: '12345',
+    livingSpace: 23,
+    householdSize: 1,
+    subType: ApartmentType.Rent,
```
robinandeer, line 19: ...and the other change in the new API: `type` => `subType`

guilhermespopolin (commented):
LGTM. Great approach used to insert such change. One thing tho is that I was trying to test it locally but it keeps returning the following error on `QueryBundleVariants` query:

```
{
  "errors": [
    {
      "message": "invalid json response body at https://gateway.test.hedvig.com/underwriter/public/quotes? reason: Unexpected end of JSON input",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "quoteBundle"
      ],
      "extensions": {
        "code": "INTERNAL_SERVER_ERROR",
        "exception": {
          "errors": [
            {
              "message": "invalid json response body at https://gateway.test.hedvig.com/underwriter/public/quotes? reason: Unexpected end of JSON input",
              "locations": [],
              "path": [
                "quoteBundle"
              ]
            }
          ]
        }
      }
    }
  ],
  "data": null
}
``` 
Maybe something change on the API 🤷🏻‍♂️

robinandeer:
@guilhermespopolin hmmm which quote are you trying to create? I would expect the API to be a little shaky but SE quotes have worked for me at least.

guilhermespopolin:
> @guilhermespopolin hmmm which quote are you trying to create? I would expect the API to be a little shaky but SE quotes have worked for me at least.

All right, jokes on me I guess. It happen when I try to go to offer page but that's because this part is not in the scope of this PR, right?! 🤦🏻
Tested again creating quotes of all types (Swedish Apartment, Swedish House and Swedish Apartment + Accident) and it works :)

guilhermespopolin (dismissed):

src/client/pages/OfferDebugger/components/QuoteFormSweden.tsx
robinandeer: This will use the generic api in the future but it's a little unclear how this will work at the moment so let's go with this for now.

src/client/graphql/CreateQuoteBundle.graphql
```diff
@@ -0,0 +1,17 @@
+mutation CreateQuoteBundle($id: ID!, $quotes: [QuoteData!]!) {
```
guilhermespopolin, line 1: Do you think would be worth it this renaming?

```suggestion
mutation CreateQuoteBundle($quoteCardId: ID!, $quotes: [QuoteData!]!) {
```

src/client/graphql/QuoteCart.graphql
```diff
@@ -0,0 +1,96 @@
+query QuoteCart($id: ID!) {
+  quoteCart(id: $id) {
+    id
+    bundle {
+      possibleVariations {
+        id
+        tag(locale: sv_SE)
```
guilhermespopolin, line 7: Wouldn't be the case of getting `locale` as an argument of `QuoteCart` query?

src/client/pages/OfferDebugger/Debugger.tsx
```diff
@@ -0,0 +1,88 @@
...
+    >
+      👉 Try starting over by creating a new session!
+    </Button>
+  </div>
+)
+
+export const Debugger: React.FC = () => {
+  const locale = useCurrentLocale()
+  const [
+    createOnboardingSession,
```
guilhermespopolin, line 44: Just a question: Do we wanna stick with this name or what we understood as a _onboarding session_ is now a _quote card_?

guilhermespopolin (dismissed):
LGTM :)
Just left a couple of comments

src/client/graphql/CreateQuoteBundle.graphql
robinandeer: Right, yeah I can see that this is clearer.

src/client/graphql/QuoteCart.graphql
robinandeer: haha yeah sorry, this is confusing -- I actually update this query in another PR! But maybe I should fix it here already

src/client/pages/OfferDebugger/Debugger.tsx
robinandeer: Oh what -- good catch! I tried to update it everywhere but must've missed this one

src/client/pages/OfferDebugger/Debugger.tsx
```diff
@@ -0,0 +1,88 @@
...
+      foreground={colorsV3.gray100}
+      onClick={resetErrorBoundary}
+    >
+      👉 Try starting over by creating a new session!
+    </Button>
+  </div>
+)
+
+export const Debugger: React.FC = () => {
+  const locale = useCurrentLocale()
```
styrelius, line 42: Maybe some destructuring? 😀 
```suggestion
  const locale = useCurrentLocale()
  const { apiMarket, isoLocale } = locale
```

src/client/pages/OfferDebugger/Debugger.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const [
+    createOnboardingQuoteCart,
+    { data },
+  ] = useCreateOnboardingQuoteCartMutation()
+  const quoteCartId = data?.onboardingQuoteCart_create
```
styrelius, line 47: I know this is not anything that can be done in this PR of course but it feels kind of weird that the `onboardingQuoteCart_create` just returns an ID as a string and not a data structure like {id: [the quote cart ID]}. Because when using this mutation it's not clear what the data you get back is, unless you check the schema 🤔 But it's nice to not always have to do that, and just get help by VS Code instead.

src/client/pages/OfferDebugger/Debugger.tsx
```diff
@@ -0,0 +1,88 @@
...
+    // create initial onboarding session
+    createNewQuoteCart()
+  }, [createNewQuoteCart])
+
+  return (
+    <Wrapper>
+      <h1>Offer Page Debugger</h1>
+
+      <Row>
+        <h3>Session: {quoteCartId}</h3>
```
styrelius, line 65: Maybe be clearer on what the "Session" is to someone only using the debugger and not looking at the code? 🤔 
```suggestion
        <h3>Quote cart session ID: {quoteCartId}</h3>
```

gustaveen (dismissed):
Nicely done! Worked great for me, let's get this out the door 🙌

src/client/pages/OfferDebugger/components/QuoteData.tsx
```diff
@@ -0,0 +1,393 @@
...
+}
+
+const ButtonLoadingIndicator = styled(LoadingDots)`
+  display: inline-flex;
+  margin-left: 0.5rem;
+`
+
+export const QuoteData: React.FC<OfferProps> = ({ quoteCartId }) => {
+  const currentLocale = useCurrentLocale()
+  const currentMarket = useMarket()
```
styrelius, line 171: Can we instead use `useCurrentLocale` from `/l10n/useCurrentLocale.ts` and then pick the `market` property from  the returned `locale` object instead? 🙏 I'm trying to get rid of all these old locale/market utils functions (and yes, I should have marked them as deprecated but I didn't think about that 🙈)

guilhermespopolin (dismissed):
Forgot to approve hahah

src/client/pages/OfferDebugger/components/QuoteFormSweden.tsx
styrelius: OK 🙃

src/client/pages/OfferDebugger/index.tsx
```diff
@@ -0,0 +1,21 @@
...
+  const module = await import(
+    /* webpackChunkName: "offer-debugger" */ './Debugger'
+  )
+  return { default: module.Debugger }
+})
+
+export const OfferDebugger: React.FC = () => {
+  if (window.hedvigClientConfig.appEnvironment === 'production') {
+    return null
+  }
```
styrelius, line 14: I know this is what we do currently with the debugger but it's kind of weird that if you go to this route in prod you just get a black page. Could we redirect to the landing page instead? As we do with all other non-existent routes?

styrelius (changes requested):
Great job 💐 ✨  

Maybe this is a golden opportunity to rename this whole feature something like **Quote Generator** or **Quote Creator**  which we've talked about previously ("debugger" is kind of a confusing name IMO)?

styrelius (commented):
OK I know we want to merge this so I'm changing this to an *Approve*. That said, I still want my comments answered and I'd still love for the `useMarket` hook to be replaced 🙏

styrelius:
Aha LOL you can't apparently just change your review from "Requesting changes" to "Approving". I'm so sorry 😔

src/client/pages/OfferDebugger/Debugger.tsx
```diff
@@ -0,0 +1,88 @@
...
+    <div>
+      <p>Error message: {error.message}</p>
+      <pre style={{ whiteSpace: 'pre-wrap' }}>{error.stack}</pre>
+    </div>
+    <Button
+      background={colorsV3.red500}
+      foreground={colorsV3.gray100}
+      onClick={resetErrorBoundary}
+    >
+      👉 Try starting over by creating a new session!
```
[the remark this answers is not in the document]
robinandeer: We are not storing any state outside React in this version of the debugger. Do you still think something is missing that needs to be reset?

robinandeer:
> Aha LOL you can't apparently just change your review from "Requesting changes" to "Approving". I'm so sorry 😔

No worries, I'll review the changes and re-request review.

> Maybe this is a golden opportunity to rename this whole feature something like Quote Generator or * Quote Creator* which we've talked about previously ("debugger" is kind of a confusing name IMO)?

Sure, probably a good idea. However, like I wrote in the "What?" section I did not set out to make general improvements in this PR -- just make sure the debugger works with the new API!

src/client/pages/OfferDebugger/index.tsx
robinandeer: Sure, I don't see why not 👍

src/client/pages/OfferDebugger/Debugger.tsx
robinandeer: Yes indeed, let's stick to one concept as much as possible.

src/client/pages/OfferDebugger/Debugger.tsx
robinandeer: Sure, please comment that to Tobias or directly in https://github.com/HedvigInsurance/rfc/pull/7!

src/client/graphql/CreateQuoteBundle.graphql
```diff
@@ -0,0 +1,17 @@
+mutation CreateQuoteBundle($quoteCartId: ID!, $quotes: [QuoteData!]!) {
+  quoteCart_createQuoteBundle(id: $quoteCartId, input: { quoteData: $quotes }) {
+    ... on QuoteCart {
```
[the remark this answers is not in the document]
robinandeer: Yep, we can make that optimisation but don't think it will give us much in this debugger tool, do you agree?

I think we can assume everything is fine unless we get back `CreateQuoteBundleError`.

src/client/pages/OfferDebugger/Debugger.tsx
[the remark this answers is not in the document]
robinandeer: You are referring to the API, right? I think it would be great to add that to https://github.com/HedvigInsurance/rfc/pull/7!

guilhermespopolin (approved):

src/client/pages/OfferDebugger/Debugger.tsx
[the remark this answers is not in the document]
styrelius: No it was a joke, sorry 🙃

src/client/graphql/CreateQuoteBundle.graphql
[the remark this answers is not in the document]
styrelius: Yeah I agree maybe the "debugger" is not the place where those kinds of optimizations but it just feels strange and it's something I've been thinking about every time I've seen the debugger code 🤷‍♀️ But let's skip doing anything about that then.

src/client/pages/OfferDebugger/Debugger.tsx
styrelius: So should we change the `<h3>`?

styrelius (approved):
👏

PR #685, opened 2021-11-16
GRW-591 / Feature / New API on Offer Page

robinandeer:
## What?

Implement the new Onboarding API on the Offer Page.
It works by picking up a Session ID from the "session" query parameter in the URL.

> I've only made sure that things work superficially so it's easy to review the code changes. Checkout is not working. Campaign codes are not working etc.

## Why?

I'm surprised that we need so few changes to support the basic-basic functionality on the offer page! I feel like this is the best approach - both making very few changes and making it clear what we can remove post-migration.

I don't use any feature flag since the session query is enough already. But please provide any feedback if you feel differently.

**Referenced ticket(s): [GRW-591]**

## Screenshots / recordings

No visual changes.

## Review app

[Debugger](https://web-onboardi-grw-591-fe-uh4sw7.herokuapp.com/se/new-member/offer-debugger)

[GRW-591]: https://hedvig.atlassian.net/browse/GRW-591?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robinandeer:
close in favour of #701

PR #686, opened 2021-11-16
Phone number offer page

amandaflod:
## What?

- Created a PhoneNumber component for the Swedish web-onboarding page.
- Added the component to ConnectPayment, Embark, Landing and Offer page

## Why?

To make our IEX-Team even more accessible! 

**Referenced ticket(s): [GRW-584]**

Screenshot 2021-11-16 at 15 47 53

[GRW-584]: https://hedvig.atlassian.net/browse/GRW-584?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/components/PhoneNumber/PhoneNumber.tsx
```diff
@@ -0,0 +1,89 @@
...
+`
+
+const PhoneLink = styled.a`
+  text-decoration: none;
+  color: inherit;
+  font-size: 0.875rem;
+
+  :hover {
+    color: ${colorsV3.gray600};
+  }
```
robinandeer, line 38:
```suggestion
  &:hover {
    color: ${colorsV3.gray600};
  }
```

src/client/l10n/phoneNumbers.tsx
```diff
@@ -0,0 +1,19 @@
+import { MarketLabel } from './locales'
+
+type PhoneNumbers = Partial<Record<MarketLabel, PhoneNumberData>>
+
+export type PhoneNumberData = {
+  displayNumber: string
+  hrefNumber: string
+  opens: string
+  closes: string
```
robinandeer, line 9: Should we name these as `opensAt` and `closesAt` to hint that they refer to a time?

src/client/l10n/phoneNumbers.tsx
```diff
@@ -0,0 +1,19 @@
...
+}
+
+export const phoneNumbers: PhoneNumbers = {
```
robinandeer, line 12: Long term I think this should come from a CMS but okey for now :)

src/client/pages/ConnectPayment/index.tsx
```diff
@@ -20,7 +23,11 @@ export const ConnectPayment: React.FC = () => {
           <title>{textKeys.ONBOARDING_CONNECT_DD_PAGE_TITLE()}</title>
         </Helmet>
         <TopBar>
-          <LanguagePicker path="/new-member/connect-payment" />
+          {market === Market.Se ? (
```
robinandeer, line 26: Let's use a feature flag here - just let me know and I can introduce you to them! 

We want to avoid "market" checks in the code -- everything should be a global feature :)

src/client/pages/Landing/Landing.tsx
```diff
@@ -203,7 +204,11 @@ export const Landing: React.FC = () => {
             {![Variation.IOS, Variation.ANDROID].includes(variation!) && (
               <>
                 <TopBar isTransparent>
-                  <LanguagePicker />
+                  {market === Market.Se ? (
+                    <PhoneNumber color="white" />
+                  ) : (
+                    <LanguagePicker />
+                  )}
```
robinandeer, line 211: Are we sure that this should be implemented on landing page as well? How can a user switch language then?

robinandeer (commented):
Cool stuff!

The hover state doesn't seem to work 100% though:

Screenshot 2021-11-16 at 16 38 29

src/client/components/PhoneNumber/PhoneNumber.tsx
```diff
@@ -0,0 +1,89 @@
...
+  const isPhoneOpen = (): boolean => {
+    if (
+      (currentLocale.phoneNumber &&
+        currentTime < currentLocale.phoneNumber?.closes) ||
+      (currentLocale.phoneNumber &&
+        currentTime > currentLocale.phoneNumber?.opens)
+    ) {
+      return true
+    } else return false
+  }
```
guilhermespopolin, line 65: [Suggestion] In case you think it's worth it, maybe we could simplify the logic a little bit like the following: 

```suggestion
  const isCallServiceAvailable = () => {
    if (!currentLocale.phoneNumber) {
      return false
    }

    const { opens, closes } = currentLocale.phoneNumber
    return currentTime >= opens && currentTime >= closes
  }
```

guilhermespopolin (commented):
We probably would need to enhance the contrast on Offer page:

Screenshot 2021-11-17 at 09 21 13

Despite of that, I've also added a couple of comments. 👍🏻

src/client/l10n/phoneNumbers.tsx
amandaflod: Sounds good! :))

src/client/l10n/phoneNumbers.tsx
amandaflod: Yes, Ellen and I had that discussion but came to the conclusion that this is good enough for now and then we'll see if it should be expanded to all markets etc.

src/client/pages/ConnectPayment/index.tsx
amandaflod: Ok! I would love some insights into how we use feature flags!

src/client/pages/Landing/Landing.tsx
amandaflod: That makes sense. I wasn't sure before because I thought that everyone came to the landing page either through the website or the forever page and that the language was picked there but maybe they should be able to switch here too! :) I will remove it here either way, as people probably won't have any questions at this point anyway :P

src/client/components/PhoneNumber/PhoneNumber.tsx
```diff
@@ -0,0 +1,89 @@
...
+const Text = styled.p<Props>`
+  color: ${(props) =>
+    props.color === 'black' ? colorsV3.gray800 : colorsV3.gray500};
+  font-size: 0.75rem;
+  && {
+    margin: 0.2rem 0 0 1.4rem;
+  }
+`
+
+export const PhoneNumber: React.FC<Props> = ({ color }: Props) => {
```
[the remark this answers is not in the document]
amandaflod: Thank you! I love removing excess code :)

src/client/l10n/phoneNumbers.tsx
robinandeer: For sure! Totally agree

src/client/pages/ConnectPayment/index.tsx
robinandeer: Cool, just let me know and I can introduce you

src/client/pages/Landing/Landing.tsx
robinandeer: haha hopefully not - we can double-check with @zakkjellin as well :)

src/client/components/PhoneNumber/PhoneNumber.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+const InnerWrapper = styled.div`
+  display: flex;
+  align-items: center;
+`
+
+const IconWrapper = styled.div`
+  display: flex;
+  align-items: center;
+  && {
```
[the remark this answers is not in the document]
robinandeer: `&` is a reference to the class name of the styled component itself. By adding additional `&&` you can increase the CSS specificity of the rule to override e.g. global styles.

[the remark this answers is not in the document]
guilhermespopolin: 🤔 I see. But why do we need this here? Are we concerned with some global style rule that sets `margin` for regular  `div`s?

src/client/pages/ConnectPayment/index.tsx
styrelius: Oh, or just check if `currentLocale` (which you import from `useCurrentLocale` from `/client/l10n/useCurrentLocale.ts`) has the phone number stuff?

styrelius: That's one of the points of adding it to the `locales` object in my mind 😊 But then if we want to merge this we also need a feature flag to hide this until we know the correct opening hours etc.

@robinandeer @amandaflod

src/client/components/PhoneNumber/PhoneNumber.tsx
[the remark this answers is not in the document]
styrelius: I think in general it's good to avoid `&&` if it's not for some strange reason strictly necessary. It mostly just raises a lot of questions and makes it harder to understand what style will actually be applied to what. And I truly hope that we don't have any global rules setting `margin` for `div`s since that would be crazy 🤪

src/client/components/PhoneNumber/PhoneNumber.tsx
```diff
@@ -0,0 +1,89 @@
...
+  }
+`
+
+export const PhoneNumber: React.FC<Props> = ({ color }: Props) => {
+  const textKeys = useTextKeys()
+  const currentLocale = useCurrentLocale()
+
+  const currentTime = format(new Date(), 'HH')
+
+  const isPhoneOpen = (): boolean => {
```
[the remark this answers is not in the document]
styrelius: Explicit return types can be great, and really bring clarity to what the idea of a function is. So I'm not sure if one can say that "in general" it's good or bad :) However when only returning a `boolean` there's probably no need to be explicit about it, especially not when following the convention of naming booleans prefixed with `is` or `has` etc.

src/client/components/PhoneNumber/PhoneNumber.tsx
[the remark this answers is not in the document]
guilhermespopolin: > And I truly hope that we don't have any global rules setting margin for divs since that would be crazy

That's my point 😆 
I've got my questions answered so gonna resolve this. Thanks guys 🙇🏻

src/client/pages/Embark/index.tsx
```diff
@@ -151,7 +159,13 @@ const Embark: React.FunctionComponent<EmbarkProps> = (props) => {
...
                 storyData={state.data}
                 startPageLink={props.startPageLink}
-                customTrailingContent={<LanguagePicker />}
+                customTrailingContent={
+                  market === Market.Se ? (
+                    <PhoneNumber color="white" />
+                  ) : (
+                    <LanguagePicker />
+                  )
+                }
```
styrelius, line 168: Related to the comment @robinandeer made that I also answered:
So instead of using the `useMarket` that's actually deprecated and shouldn't be used (I've just failed to communicate this properly) I suggest importing `useCurrentLocale` from `l10n/useCurrentLocale.ts` and then do
```suggestion
                customTrailingContent={
                  currentLocale.phoneNumber ? (
                    <PhoneNumber color="white" />
                  ) : (
                    <LanguagePicker />
                  )
                }
```
Same goes for the other files where markets are checked to see if the phone number should show 😊

src/client/components/PhoneNumber/PhoneNumber.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { Telephone } from '../icons/Telephone'
+
+type Props = {
+  color?: 'black' | 'white'
+}
+
+const Wrapper = styled.div<Props>`
+  color: ${(props) =>
+    props.color === 'black' ? colorsV3.black : colorsV3.white};
+`
```
styrelius, line 16: I'd destructure this to remove words and in that way make it a bit more readable:
```suggestion
const Wrapper = styled.div<Props>`
  color: ${({ color }) =>
    color === 'black' ? colorsV3.black : colorsV3.white};
`
```

src/client/components/PhoneNumber/PhoneNumber.tsx
```diff
@@ -0,0 +1,89 @@
...
+  color: inherit;
+  font-size: 0.875rem;
+
+  :hover {
+    color: ${colorsV3.gray600};
+  }
+`
+
+const Text = styled.p<Props>`
+  color: ${(props) =>
```
styrelius, line 42: Same here - destructuing props makes the code easier to read, at least for me. But it anyone wants to challenge that, of course go ahead 😇

src/client/components/PhoneNumber/PhoneNumber.tsx
```diff
@@ -0,0 +1,89 @@
...
+    } else return false
+  }
+
+  return (
+    <Wrapper color={color}>
+      <InnerWrapper>
+        <IconWrapper>
+          <Telephone size="1rem" />
+        </IconWrapper>
+        <PhoneLink href={currentLocale.phoneNumber?.hrefNumber}>
```
styrelius, line 73: [Nit] Since you're referencing `currentLocale.phoneNumber` a lot it might be nice to save it to a variable somewhere (outside of the return statement though)

styrelius (dismissed):
Great job Amanda! ☎️  Just a few comments 😊

src/client/components/PhoneNumber/PhoneNumber.tsx
styrelius: I think that should be a general pattern we follow actually - to always destructure props. For readability and also you can directly see what the props are, they're not hidden in an obscure object 🤠

src/client/components/PhoneNumber/PhoneNumber.tsx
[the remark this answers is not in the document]
amandaflod: Ok, I am not quite sure how to act on this since after looking through the code base we have been a bit inconsistent about this. 
There is a lint rule that we can use if we want to enforce this consistently across the code base:
```
        "@typescript-eslint/explicit-function-return-type": [
          "error",
          {
            "allowExpressions": true
          }
        ]
```

I'm not sure what the general thinking is for the codebase but maybe it would be good to keep it consistent and enforced through a linter rather than through convention? Feel free to challenge me on this! :D @robinandeer @styrelius @gustaveen

src/client/components/PhoneNumber/PhoneNumber.tsx
amandaflod: Great to know! :D

src/client/components/PhoneNumber/PhoneNumber.tsx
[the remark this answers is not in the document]
robinandeer: Yes, automate everything!

src/translations/en.json
```diff
@@ -309,10 +309,6 @@
   "CHECKOUT_LASTNAME_LABEL": "Last name",
   "HERO_FIRSTNAME_PLACEHOLDER": "You",
   "PHONE_OPEN_TODAY": "Open today",
-<<<<<<< HEAD
```
robinandeer: Ohh... seems like you've committed some merge conflict comments :/

src/client/components/PhoneNumber/PhoneNumber.tsx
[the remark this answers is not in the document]
gustaveen: For sure, as you said it can be nice with return types and be explicit, but in other cases it feels bloated 😄 So a linter rule might good to have instead 🙌

src/translations/en.json
amandaflod: Thank you!

src/client/pages/ConnectPayment/index.tsx
amandaflod: Yes! Now Sonny wants some A/B testing as well so we will not be able to merge this without a feature flag before things are more specified! :)

amandaflod: Now we're not doing A/B testing anymore! :P

robinandeer (dismissed):
I'm happy with the code now! Very nice work :D

(haven't tried changes locally)

src/client/components/PhoneNumber/PhoneNumber.tsx
```diff
@@ -74,16 +115,7 @@ export const PhoneNumber: React.FC<Props> = ({ color }) => {
...
-        <Text color={color}>
-          {textKeys.PHONE_OPEN_TODAY()} {phoneNumber?.opensAt}-
-          {phoneNumber?.closesAt}
-        </Text>
-      ) : (
-        <Text color={color}>
-          {textKeys.PHONE_CLOSED_UNTIL()} {phoneNumber?.opensAt}
-        </Text>
-      )}
+      <PhoneOpeningHours phoneNumber={phoneNumber} color={color} />
```
robinandeer, line 118: I would prefer to handle the case where `phoneNumber` is not defined right here:

```suggestion
      {phoneNumber ? <PhoneOpeningHours phoneNumber={phoneNumber} color={color} /> : null}
```

Then we don't need to worry about that case inside the component :)

robinandeer (dismissed):
I think this looks good! Just one comment on how to handle missing data.

Would be nice to add this component to storybook to visualise the different states but we can do that in a separate PR :)

styrelius (commented):
Should this also be "feature flagged" so we can merge this even if the case is that the call center / sales partner is ready to start taking calls?

src/client/components/PhoneNumber/PhoneNumber.tsx
```diff
@@ -0,0 +1,121 @@
+import styled from '@emotion/styled'
+import { colorsV3 } from '@hedviginsurance/brand'
+import React from 'react'
+import { format as formatDate, getDay } from 'date-fns'
+import { useCurrentLocale } from 'l10n/useCurrentLocale'
+import { useTextKeys } from 'utils/textKeys'
+import { PhoneNumberData } from 'l10n/phoneNumbers'
+import { Telephone } from '../icons/Telephone'
+
```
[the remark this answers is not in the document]
amandaflod: Good suggestion! Will do :))

guilhermespopolin (dismissed):

src/client/components/PhoneNumber/PhoneNumber.tsx
```diff
@@ -0,0 +1,121 @@
...
+  if (!phoneNumber) {
+    return null
+  }
+
+  const { opensAt, closesAt, lunchStartsAt, lunchEndsAt } = phoneNumber
+  const isPhoneOpen = currentTime >= opensAt && currentTime < closesAt
+  const isLunchHour = currentTime >= lunchStartsAt && currentTime < lunchEndsAt
+  const isFridayAfterHours = currentDay === 5 && currentTime >= closesAt
+  const isWeekend = currentDay === 6 || currentDay === 0
```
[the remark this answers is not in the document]
amandaflod: Ok, but then I would need an intro to you test setup! 😇 Should we prioritize that in this PR?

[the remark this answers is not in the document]
styrelius: (Sorry about the late response) Hm, maybe not - although for unit tests we just use the basic Jest setup so shouldn't be super complicated.

src/client/utils/tracking/gtm.ts
```diff
@@ -100,13 +106,19 @@ export enum EventName {
   OfferCreated = 'offer_created',
   SignedCustomer = 'signed_customer',
   InsuranceSelectionToggle = 'insurance_selection_toggle',
+  ClickCallNumber = 'click_call_number',
+}
+
+type OptionalParameters = {
```
gustaveen, line 112: Nice! 👍

gustaveen (dismissed):
Super good job Amanda, this is good to go IMO! 🌟

src/client/pages/ConnectPayment/index.tsx
```diff
@@ -2,16 +2,33 @@ import { SemanticEvents } from 'quepasa'
...
 
 export const ConnectPayment: React.FC = () => {
   const textKeys = useTextKeys()
+  const currentLocale = useCurrentLocale()
+
+  const history = useHistory()
+
+  const onClickPhoneTracking = (path: string, status: 'opened' | 'closed') => {
+    pushToGTMDataLayer({
+      event: 'click_call_number',
```
robinandeer, line 25: Should/could we use the event enum here?

src/client/pages/ConnectPayment/index.tsx
```diff
@@ -2,16 +2,33 @@ import { SemanticEvents } from 'quepasa'
...
+import { pushToGTMDataLayer } from 'utils/tracking/gtm'
 import { ConnectPaymentPage } from './sections/ConnectPayment'
 
 export const ConnectPayment: React.FC = () => {
   const textKeys = useTextKeys()
+  const currentLocale = useCurrentLocale()
+
+  const history = useHistory()
+
+  const onClickPhoneTracking = (path: string, status: 'opened' | 'closed') => {
```
robinandeer, line 23: I would consider naming this:

```suggestion
  const handleClickPhoneNumber = (path: string, status: 'opened' | 'closed') => {
```

src/client/pages/ConnectPayment/index.tsx
```diff
@@ -20,7 +37,16 @@ export const ConnectPayment: React.FC = () => {
           <title>{textKeys.ONBOARDING_CONNECT_DD_PAGE_TITLE()}</title>
         </Helmet>
         <TopBar>
-          <LanguagePicker path="/new-member/connect-payment" />
+          {currentLocale.phoneNumber ? (
+            <PhoneNumber
+              color="white"
+              onClick={(status) =>
+                onClickPhoneTracking(history.location.pathname, status)
```
robinandeer, line 44: nit: we don't need to pass `history.location.pathname` here. We could just pick that up in the `onClickPhoneTracking` function.

src/client/utils/tracking/gtm.ts
```diff
@@ -23,7 +23,7 @@ type GTMUserProperties = {
   environment: AppEnvironment
 }
 
-type GTMOfferData = {
+export type GTMOfferData = {
```
robinandeer, line 26: I don't this being imported anywhere else so we should probably not export it here :)

robinandeer (dismissed):
Nice stuff :)

I have a few comments that would be nice to address but nothing major if you wanna do that in a follow-up PR

src/client/components/PhoneNumber/PhoneNumber.tsx
```diff
@@ -0,0 +1,138 @@
...
+import { useTextKeys } from 'utils/textKeys'
+import { PhoneNumberData } from 'l10n/phoneNumbers'
+import { Telephone } from '../icons/Telephone'
+
+const { black, white, gray700, gray500 } = colorsV3
+
+type Color = 'black' | 'white'
+
+const Wrapper = styled.div<{ color: Color }>`
+  color: ${({ color }) => (color === 'black' ? black : white)};
```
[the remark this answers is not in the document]
amandaflod: Cool! Will add that :D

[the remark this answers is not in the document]
amandaflod: Already doing that, see line 10 :)
```
const { black, white, gray700, gray500 } = colorsV3
```

src/client/pages/ConnectPayment/index.tsx
amandaflod: Great, I'll fix!

src/client/pages/ConnectPayment/index.tsx
amandaflod: 👍

src/client/utils/tracking/gtm.ts
amandaflod: Good catch!

src/client/pages/ConnectPayment/index.tsx
amandaflod: Sounds good, we can look at this later

robinandeer:
Sorry @amandaflod! Must've merged something that caused another conflict 🤦

src/client/pages/ConnectPayment/index.tsx
```diff
@@ -20,7 +37,14 @@ export const ConnectPayment: React.FC = () => {
           <title>{textKeys.ONBOARDING_CONNECT_DD_PAGE_TITLE()}</title>
         </Helmet>
         <TopBar>
-          <LanguagePicker path="/new-member/connect-payment" />
+          {currentLocale.phoneNumber ? (
+            <PhoneNumber
+              color="white"
+              onClick={(status) => handleClickPhoneNumber(status)}
```
robinandeer, line 43: nit:

```suggestion
              onClick={handleClickPhoneNumber}
```

robinandeer (dismissed):
Looks good :D

robinandeer (approved):

PR #687, opened 2021-11-16
GRW-587 / Feature / Pass QuoteCart ID to initial Embark store

robinandeer:
## What?

Pass quote cart ID to initial Embark store.
Create onboarding quote cart on landing page.
Pick up quote cart ID when rendering Embark provider.

## Why?

The new APIs no longer rely on an implicit token but rather an explicit ID to identify which onboarding session is active.

No visual changes.

**Referenced ticket(s): []**

> This is mostly to get other peoples input on the approach. We still need some error handling if the "create quote cart" mutation fails for example.


src/client/pages/Embark/index.tsx
```diff
@@ -208,9 +209,9 @@ interface AngelVariables {
 export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
   const history = useHistory()
   const [data, setData] = React.useState<[string, any] | null>(null)
-  const [initialStore, setInitialStore] = React.useState<null | {
-    [key: string]: any
-  }>()
+  const [initialStore, setInitialStore] = React.useState<Record<string, any>>({
+    quoteCartId: props.quoteCartId,
+  })
```
robinandeer, line 214: @sampettersson @hugokallstrom @hedvigoscar I guess this is sort of what you had in mind as well to communicate the quote cart ID by setting it in the initial Embark store, right?

robinandeer:
Updates needed:

- [ ] Create quote cart on landing page (show black screen while loading)
- [ ] Ask Zak about possible error state

robinandeer:
Close in favour of #694

PR #688, opened 2021-11-18
Chore: update translations

guilhermespopolin:
## What?

- Update translations on web onboarding

## Why?

- So we have access to newer translations


robinandeer (approved):
Looks good!

PR #691, opened 2021-11-18
GRW-579 fix: close Checkout DatePicker when sidebar is closed

guilhermespopolin:
## What?

- In case its open, DatePicker now is being closed when Checkout sidebar is closed

## Why?

- Because being kept opened when Checkout Sidebar is closed might be confusing
- Also took the opportunity to use the same _click outside detection_ in the Tooltip

### How?
- The idea was change the "click outside" implementation since the current one make usage of a _Backdrop_ that were only considering the sidebar content.

**Referenced ticket(s): [GRW-579](https://hedvig.atlassian.net/browse/GRW-579)**


src/client/components/Tooltip/Tooltip.tsx
```diff
@@ -87,18 +88,12 @@ export const Tooltip: React.FC<TooltipProps> = ({ body }) => {
...
-  useEffect(() => {
-    if (!isVisible) return
-
-    const handleClickOutside = (event: TouchEvent) => {
-      if (ref.current && !ref.current.contains(event.target as Node)) {
-        setVisible(false)
-      }
+  const handleClickOutside = useCallback(() => {
+    if (isVisible) {
+      setVisible(false)
```
robinandeer, line 93: Can't this be simplified to:

```
const handleClickOutside = useCallback(() => setVisible(false), [])
```

- `setVisible` doesn't need to be included in the dependency array
- after clicking outside, `isVisible` should always be `false`

src/client/components/DateInput/index.tsx
```diff
@@ -72,10 +65,17 @@ export const DateInput: React.FC<DateInputProps> = ({
   hasCurrentInsurer,
 }) => {
   const textKeys = useTextKeys()
+  const containerRef = useRef<HTMLDivElement>(null)
+
+  const handleClickOutside = useCallback(() => {
+    if (open) {
+      setOpen(false)
+    }
+  }, [open, setOpen])
```
robinandeer, line 74: See other comment.

src/client/components/DateInput/index.tsx
```diff
@@ -72,10 +65,17 @@ export const DateInput: React.FC<DateInputProps> = ({
...
+  useClickOutside(containerRef, handleClickOutside)
 
   return (
     <>
```
robinandeer, line 78: This fragment can be removed, no?

robinandeer (commented):
Very nice! Just have a look at the comments if you think they make sense

src/client/components/DateInput/index.tsx
guilhermespopolin: Thanks

src/client/components/Tooltip/Tooltip.tsx
guilhermespopolin: Yeah, makes sense. Thanks

src/client/components/DateInput/index.tsx
guilhermespopolin: Here's a little bit different because we receive `setOpen` as `prop`. If I remove it from dependency array the "missing dependency" lint warning will be shown. Which makes sense if we think on the scenario where consumer passes a different `setOpen` callback, right?!

src/client/components/Tooltip/Tooltip.tsx
guilhermespopolin: 🤔 I believe I can remove `setVisible` from dependency array but `isVisible` I can't. If we look back to the [previous](https://github.com/HedvigInsurance/web-onboarding/blob/master/src/client/components/Tooltip/Tooltip.tsx#L90) implementation, we're returning early in case it's not visible. I believe that's happening due animation. We're not removing `TooltipContainer` from the dom so we can animate it's `visibility`. However since it's being unmounted `handleClickOutside` would still be attached and if we, while using mobile, click on tooltip icon `handleClickOutside` will be executed and, as we would be setting `isVisible` always to false, the tooltip content would never be shown.

guilhermespopolin: Actually, it turn out `useClickOutside` does not apply to `Tooltip` so I'm reverting it to previous implemeation.

src/client/components/DateInput/index.tsx
robinandeer: yeah, if it's props they should be in the dependency array, you are right :)

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -133,12 +133,12 @@ export type CheckoutProps = {
   refetch: () => Promise<void>
 }
 
-export const Checkout: React.FC<CheckoutProps> = ({
+export const Checkout = ({
```
robinandeer, line 136: nit: don't really understand why this change is preferable

robinandeer (approved):
Very nice :)

src/client/pages/OfferNew/Checkout/index.tsx
guilhermespopolin: [This Github issue](https://github.com/facebook/create-react-app/pull/8177) was mentioned on a article that I've read about [writing react components with Typescript](https://kentcdodds.com/blog/how-to-write-a-react-component-in-typescript). It's not a big deal. In fact that change shouldn't have been committed. 😬

PR #671, opened 2021-11-08
GRW-499 Support loading offer page when first & last names are missing in quote + collect this info at checkout

mehdielket:
A review app was created so you can test it by clicking [here](https://web-onboardi-grw-499-su-6vpwgl.herokuapp.com/se-en/new-member)

## What?

Add support for loading the offer page while quotes do not have a set first & last name
* Displays a placeholder for the name in the introduction
* Collects the two fields at checkout
* Is show the fields if the first and last names are missing on load. After adding them and reloading, they will disappear (user can still change that info in the personal details popup)

What is still missing

- [x] Update types to reflect that the names are nullable from the GQL API (awaiting [PR](https://github.com/HedvigInsurance/giraffe/pull/1417))
- [x] Fix copy - translate the keys needed

## Why?

External partners would like to be able to generate quotes through the partner API without setting the names and be able to redirect to our offer page. More details in the ticket

## Important Notes

This PR makes conscious compromises by not refactoring the checkout form functionality.
* The external partners are expecting this in the very need future.
* This checkout form needs large refactor as it is suffering for a few reported issues. 
* Refactoring in a holistic way can be considered as part of the upcoming checkout re-design.

_Referenced ticket(s): [GRW-499]_

Screenshot 2021-11-09 at 11 53 30

Screenshot 2021-11-09 at 11 53 43

https://user-images.githubusercontent.com/89857278/140912250-c63eb157-e3a5-4c3e-b2cb-282febde53a3.mov

[GRW-499]: https://hedvig.atlassian.net/browse/GRW-499?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #693, opened 2021-11-23
GRW-400 - Feature: Added UpsellCard for Accident Sweden

guilhermespopolin:
## Disclaimer
It's opened as a Draft because we need to add _text keys_

## What?

- This adds `UpsellCard` as an experiment to check it's impact while trying to selling _Accident Insurance_ on Swedish market.

## Why?

- We would like to know if we could increase selling of _Accident Insurance_ with it.

**Referenced ticket(s): [GRW-400](https://hedvig.atlassian.net/browse/GRW-400)**

## Screenshots / recordings

desktop-se
desktop-en
Screenshot 2021-11-26 at 09 44 13
Screenshot 2021-11-26 at 09 43 36

## Review app

It can be test using this [review-app](https://web-onboardi-feature-gr-hww2oz.herokuapp.com/se-en/new-member)


src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -126,19 +164,84 @@ const Backdrop = styled('div')<Openable>`
...
+      description={textKeys.UPSELL_DESCRIPTION()}
+      price={price}
+      onAcceptDeal={() => onAcceptDeal(betterDealQuoteIds)}
+    />
+  )
+}
+
 export type CheckoutProps = {
-  offerData: OfferData
+  variants: QuoteBundleVariant[]
```
robinandeer, line 224: We should probably agree a name for what we call these. Maybe `quoteBundleVariants`?

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -126,19 +164,84 @@ const Backdrop = styled('div')<Openable>`
...
+  variants,
+  selectedQuoteBundle,
+  onAddQuotes,
   isOpen,
   onClose,
   refetch,
 }: CheckoutProps) => {
+  const textKeys = useTextKeys()
+  const locale = useCurrentLocale()
+  const market = useMarket()
```
robinandeer, line 242: Feel free to switch these out to the new ones from the `l10n` module!

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -158,18 +261,25 @@ export const Checkout = ({
   const [emailUpdateLoading, setEmailUpdateLoading] = useState(false)
   const [ssnUpdateLoading, setSsnUpdateLoading] = useState(false)
   const [isShowingFailModal, setIsShowingFailModal] = useState(false)
+  const [fakeLoading, setFakeLoading] = React.useState(false)
+  const [reallyLoading, setReallyLoading] = React.useState(false)
```
robinandeer, line 265: Oh these are the ones @mehdielket removed in #671, no? I hope you're not gonna run into to a nasty merge conflict 🤭

src/client/pages/OfferNew/Checkout/index.tsx
guilhermespopolin: You're right. I've taken the lazy path on this one based on another component used in that context, like [InsuranceSelector](https://github.com/HedvigInsurance/web-onboarding/blob/030c8620c3dd3b77bbbf90bae967334cb0eade5d/src/client/pages/OfferNew/index.tsx#L177). But, no excuses.

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -96,6 +105,29 @@ const InnerWrapper = styled('div')`
...
+const StartDateLabel = styled.p`
+  margin: 0 0 0.5rem 0.5rem;
+  font-size: 0.875rem;
+  line-height: 1;
+`
+
+const Heading = styled.h3`
+  margin: 0;
+  font-size: 2rem;
+  line-height: 2.5rem;
```
dendoesit, line 127:
```suggestion
  line-height: 1.25;
```

we try to have line-height set as a multiplier of the value. Referenced here https://www.notion.so/hedviginsurance/Styling-Documentation-4d829669ede84b1ba741609336231e74

src/client/pages/OfferNew/Checkout/UpsellCard.tsx
```diff
@@ -0,0 +1,72 @@
...
+  margin-bottom: 1rem;
+
+  & > *:not(:last-child) {
+    margin-bottom: 1rem;
+  }
+`
+
+const Title = styled.h1`
+  margin: 0;
+  font-size: 1.25rem;
```
dendoesit, line 22: I see this is the size in design. not sure if I have the latest version or not. 
```suggestion
  font-size: 1.5rem;
  line-height: 1.33
```

guilhermespopolin:
I saw that I've broke `Checkout` stories. Gonna fix that later.

gustaveen:
Seems like the typecheck fails @guilhermespopolin :)

src/client/utils/testData/possibleVariationsMock.ts
```diff
@@ -6,10 +6,297 @@ import {
...
+  SwedishAccidentDetails,
+  NorwegianHomeContentsDetails,
+  NorwegianHomeContentsType,
+  DanishHomeContentsDetails,
+  DanishHomeContentsType,
 } from 'data/graphql'
 
 import { perilsMock } from './offerDataMock'
 
+export const possibleVariationsApartmentSE: QuoteBundleVariant[] = [
```
robinandeer, line 19: Feels like this is gonna get out of hand quickly 😅 but I don't really know a better solution

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -130,19 +162,84 @@ const Backdrop = styled('div')<Openable>`
   }};
 `
 
+const renderUpsellCard = (
```
robinandeer, line 165: Why not write a normal react component `UppsellCard` => `<UppsellCard {...props}>`?

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -130,19 +162,84 @@ const Backdrop = styled('div')<Openable>`
...
+  selectedQuoteBundleVariant: QuoteBundleVariant,
+  onAcceptDeal: (quoteIds: string[]) => void,
+  textKeys: Record<string, any>,
+) => {
+  const isSwedishAccidentAdded = selectedQuoteBundleVariant.bundle.quotes.some(
+    ({ quoteDetails }) => isSwedishAccident(quoteDetails),
+  )
+  if (isSwedishAccidentAdded) {
+    return null
+  }
```
robinandeer, line 176: I would prefer to hide this market-specific information under a feature flag. We should consider all features as global -- but possibly only enabled in a few markets at the moment.

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -372,14 +474,28 @@ export const Checkout = ({
...
+                    </StartDateLabel>
+                    <StartDate offerData={offerData} refetch={refetch} />
+                  </StartDateWrapper>
+                  {locale.marketLabel == 'SE' &&
+                    renderUpsellCard(
+                      quoteBundleVariants,
+                      selectedQuoteBundleVariant,
+                      onAddQuotes,
+                      textKeys,
+                    )}
```
robinandeer, line 496: See above -- I don't see why this shouldn't be a normal component => `<UpsellCard />`

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -162,25 +259,28 @@ export const Checkout = ({
   const [emailUpdateLoading, setEmailUpdateLoading] = useState(false)
   const [ssnUpdateLoading, setSsnUpdateLoading] = useState(false)
   const [isShowingFailModal, setIsShowingFailModal] = useState(false)
+
+  const offerData = getOfferData(selectedQuoteBundleVariant.bundle)
+  const quoteIds = getQuoteIds(offerData)
```
robinandeer, line 264: I would try to avoid this -- now we are doing the transformation in two places (OfferNew/index + here). I think we gain a lot from keeping that limited to a single place in the code. For example, the migration to the new API will be more complicated if we do this transformation in multiple places.

robinandeer (commented):
Great work! Have a look at the comments and let me know what you think

src/client/utils/testData/possibleVariationsMock.ts
guilhermespopolin: Right?! I fell the same. That just have cost me a lot of time. Previously, we're passing `offerData` to `Checkout` which is a smaller portion of data when compared with the solution now. Although, I have the feeling that even `offerData` has more information than what `Checkout` cares about. With the `UpsellCard`, `Checkout` needs to now about all the `possibleVariants` and `selectedBundleVariant` so we can choose between display `UpsellCard` or not. Without any further investigation, I think the ideal world would be something like providing a list of quote ids to `Checkout` and them, `Checkout` would make a query to get just the information it needs to populate it properly. I would like to implement that but I think it's a major refactor and since this is an experimental feature it kind of feels its not the right moment.

src/client/pages/OfferNew/Checkout/index.tsx
guilhermespopolin: Right. So the idea was to try to keep things easy to remove as this is being defined as an experimental thing. But to honest I guess things got out of hand and there's no way to easily remove this anymore. So I guess I can just do what you've suggested.

src/client/pages/OfferNew/Checkout/index.tsx
guilhermespopolin: I see that. It's kind of a complicated problem. Do you think this would be weird:

```
<Checkout
   offerData={offerData}
   bundleVariants={bundleVariants}
   selectedBundleVariant={selectedBundleVariant}
 />
```

Another alternativa would be renders `Checkout` content as children inside `OfferNew`. That might make `OfferNew` a huge thing.

src/client/pages/OfferNew/Checkout/index.tsx
robinandeer: haha sure sure, I just mean this is a functional component already -- we just don't use a variable name that suggests it is 😅

src/client/pages/OfferNew/Checkout/index.tsx
robinandeer: Yeah, complexity can usually only be moved around to different places. I don't have a clear idea what will be the most logical setup in this case.

I guess in general we should only pas the information necessary to `Checkout`. But I'm not sure what that would be.

src/client/pages/OfferNew/Checkout/UpsellCard/Card.tsx
```diff
@@ -4,9 +4,9 @@ import { colorsV3 } from '@hedviginsurance/brand'
 
 import { MEDIUM_SCREEN_MEDIA_QUERY } from 'utils/mediaQueries'
 
-import { Button } from '../../../components/buttons'
+import { Button } from '../../../../components/buttons'
```
robinandeer, line 7: Don't we normally import like this:

```suggestion
import { Button } from 'components/buttons'
```

src/client/pages/OfferNew/Checkout/UpsellCard/index.tsx
```diff
@@ -0,0 +1,68 @@
...
+    ({ quoteDetails }) => isSwedishAccident(quoteDetails),
+  )
+  if (isSwedishAccidentAdded) {
+    return null
+  }
+
+  // Look for a bundle variant that includes Accident Insurance.
+  // In case it founds it, store the bundle variant and the accident
+  // insurance quote so we can use them to accept UpsellCard deal and
+  // get the price of Accident Insurance.
```
robinandeer, line 30: Is this the kind of "what" comment that we should try to avoid? Can we extract it to a named function instead?

Otherwise we need to remember to update this text when making changes to the function...

We should probably memoize this anyways!

robinandeer (commented):
We've figured out tracking for this feature, right? Just so we can evaluate quickly and refactor the code if it's a feature we intend to keep/maintain.

src/client/pages/OfferNew/Checkout/UpsellCard/index.tsx
mehdielket: I made some updates to avoid the comment and improve readability

src/client/pages/OfferNew/Checkout/index.tsx
mehdielket: As we've discussed, this is a larger issue around the how we handle the data that is pretty much used everywhere is the page.

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -170,7 +170,7 @@ export type CheckoutProps = {
 export const Checkout = ({
   quoteBundleVariants,
   selectedQuoteBundleVariant,
-  onAddQuotes,
+  onUpsellAccepted,
```
robinandeer, line 173: Now we are calling the same thing "deal" and "upsell" -- not crazy about that but not a deal breaker

src/client/pages/OfferNew/index.tsx
```diff
@@ -106,6 +106,26 @@ export const OfferNew: React.FC = () => {
     }
   }
 
+  const onCheckoutUpsellCardAccepted = (
```
robinandeer, line 109: nit: I would name this `handleCheckoutUpsellCardAccepted` ;)

robinandeer (dismissed):
I've looked at this too many times now 😅 but I think I'm happy with this for now until we decide it's a maintained feature that should scale globally.

Let's hold each other accountable to refactor this then.

robinandeer (approved):
re-approve!

PR #694, opened 2021-11-23
GRW-587 / Feature / Create QuoteCart ID in initial Embark store

robinandeer:
## What?

Create Quote Cart ID before starting Embark.
Pass in ID to initial store.
Add feature flag for Quote Cart API.
Update to new locale utils.

- [ ] Ask Zak about possible error state

## Why?

The new APIs no longer relies on an implicit token but rather an explicit ID to identify the active quote cart.

No visual changes.

I chose to handle it inside the Embark root component since it already has a loading state while we fetch the Angel story.

**Referenced ticket(s): []**

### DIFF
--- a/package.json
+++ b/package.json
@@ -158,6 +158,7 @@
     "react": "^16.13.0",
     "react-animate-height": "^2.0.6",
     "react-dom": "^16.13.0",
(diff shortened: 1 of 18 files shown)


robinandeer:
close in favour of #695

PR #695, opened 2021-11-24
GRW-587 / Feature / Create `QuoteCart` ID in initial Embark store

robinandeer:
## What?

Create Quote Cart ID before starting Embark.
Pass in ID to initial store.
Add feature flag for Quote Cart API.
Update to new locale utils.

- [x] Ask Zak about possible error state => use same modal as for NO credit check, mange will provide copy

## Why?

The new APIs no longer relies on an implicit token but rather an explicit ID to identify the active quote cart.

No visual changes.

I chose to handle it inside the Embark root component since it already has a loading state while we fetch the Angel story.

**Referenced ticket(s): []**

## Demo

Screenshot 2021-11-24 at 11 06 59


src/client/utils/hooks/useFeature.ts
```diff
@@ -12,24 +14,29 @@ type Env = 'staging' | 'production'
...
   },
   {
     name: Features.TEST_FEATURE,
     envs: ['staging'],
-    markets: [Market.Se],
+    markets: ['SE'],
+  },
+  {
+    name: Features.QUOTE_CART_API,
+    envs: process.env.ENABLE_QUOTE_CART_API === 'true' ? ['staging'] : [],
```
robinandeer, line 38: I'm trying this out to be able to toggle features based on an environment variable instead of requiring code changes -- let me know what you think!

src/client/pages/Embark/index.tsx
```diff
@@ -254,37 +277,41 @@ export const EmbarkRoot: React.FunctionComponent<EmbarkRootProps> = (props) => {
...
     if (!prevStore) {
-      setInitialStore({})
+      setInitialStore(defaultStore)
       return
     }
 
     try {
       const parsedPrevStore = JSON.parse(prevStore)
-      setInitialStore(parsedPrevStore as { [key: string]: any })
+      setInitialStore({ ...parsedPrevStore, ...defaultStore })
```
robinandeer, line 310: Not sure if we wanna override the `quoteCartId` here (if you e.g. reload the page after starting Embark)

guilhermespopolin (dismissed):
LGTM

src/client/utils/hooks/useFeature.ts
mehdielket: As we've discussed, maybe it would be good to move all config to the .env vars
Here is a ticket I created https://hedvig.atlassian.net/browse/GRW-599

mehdielket (dismissed):
Looks good! 🚀

I feel like this part could use some love as it seems quite complex. Let's try to keep that in mind when we touch this another time 🙂

mehdielket (approved):

PR #698, opened 2021-11-26
Feature / l10n/i18n solution, part two

styrelius:
## What?

- Old `useCurrentLocale` hook in `CurrentLocale.tsx` is marked as `@deprecated` with JSDoc comment.
- Old i18n utility functions/hooks/components, from `CurrentLocale.tsx` are replaced throughout the codebase with the new `useCurrentLocales` hook.

## Why?

This is a step on the journey to making how we work with i18n and locale specific data a bit smarter. 

**Ticket(s): [GRW-519]**

[GRW-519]: https://hedvig.atlassian.net/browse/GRW-519?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robinandeer:
@styrelius should we review and merge this already?

I'm only concerned you will run in a lot of merge conflicts otherwise since this PR is touching a lot of files.

styrelius:
@robinandeer Yep, was thinking the same :) Let's do it!

src/client/components/DateInput/Calendar.tsx
```diff
@@ -136,10 +136,11 @@ export const Calendar: React.FC<DayzedCalendarProps> = ({
 }) => {
   const [weekdayNamesShort, setWeekdayNamesShort] = React.useState<string[]>([])
   const [monthNamesShort, setMonthNamesShort] = React.useState<string[]>([])
-  const locale = useCurrentLocale()
+  const currentLocale = useCurrentLocale()
+  const currentLocalePath = currentLocale.path
```
robinandeer, line 140: I've been destructuring the results on the same line. I'm not sure if it's concern if we don't agree on one way of doing tho :)

src/client/containers/SessionContainer.tsx
```diff
@@ -75,7 +79,9 @@ export const setupSession = async (
 export const SessionContainer: React.SFC<SessionContainerProps> = ({
   children,
 }) => {
-  const pickedLocale = getIsoLocale(useCurrentLocale())
+  const currentLocale = useCurrentLocale()
+  const { isoLocale } = currentLocale
```
robinandeer, line 83: nit: Why not?

```suggestion
  const { isoLocale } = useCurrentLocale()
```

robinandeer (dismissed):
Very nice update! 🕺

src/client/components/DateInput/Calendar.tsx
gustaveen: I would vote on a one liner :)

gustaveen (dismissed):
Nicely done! 👌

src/client/containers/SessionContainer.tsx
styrelius: No idea :) Was tired maybe?

src/client/components/DateInput/Calendar.tsx
styrelius: Yep, I just think in the case of the `path` it's not a very clear name outside of its context you know? But maybe the solution is rather to change name of that property

styrelius: But of course it's also possible to do `const { path: currentLocalePath } = useCurrentLocale()`

robinandeer: Exactly, that's what I tend to do. But I would be up for changing/aliasing the `path` field so something less context dependent!

styrelius: Let's do that in the next PR. Changing name of the `path` property I mean

robinandeer (approved):
Looks good to me!

src/client/containers/SessionContainer.tsx
```diff
@@ -5,14 +5,15 @@ import { SegmentAnalyticsJs } from 'quepasa'
...
-import { getIsoLocale, useCurrentLocale } from 'components/utils/CurrentLocale'
+import { UpdatePickedLocaleDocument } from 'data/graphql'
+import { useCurrentLocale } from 'l10n/useCurrentLocale'
 import { captureSentryError } from 'utils/sentry-client'
 import { Storage, StorageContainer } from 'utils/StorageContainer'
 import {
   apolloClient as realApolloClient,
   ApolloClientUtils,
 } from '../apolloClient'
+import { LocaleData } from '../l10n/locales'
```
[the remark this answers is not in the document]
styrelius: Yeah, that's strange. I assume it's the automatic imports messing things up again 🙃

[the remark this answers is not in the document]
styrelius: But I'll fix that in the next PR

PR #699, opened 2021-11-29
Fix / Update response type of quote cart mutation

robinandeer:
## What?

Update `onboardingQuoteCart_create` mutation with new response type object.

## Why?

GraphQL schema has been updated. 

**Ticket(s): []**


gustaveen (approved):
🙌

PR #703, opened 2021-11-30
Remove deprecated colors

gustaveen:
## What?

- Replace colorsV2 with colorsV3
- Delete unused components

Colors are replaced with the equivalent v3 brand color and are visually proofed.

## Why?
Align with brand guidelines


src/client/components/inputs/index.tsx
```diff
@@ -325,7 +325,8 @@ export const InputGroupDeleteButton = styled.button`
   }
 
   :hover {
-    background: ${colorsV2.coral100};
+    /* HEX value with opcaity (1a = 10% transparency) */
+    background-color: ${colorsV3.red500}1a;
```
gustaveen, line 329: Since we don't have a translucent red in brand and this seems to be the only occurrences  
https://user-images.githubusercontent.com/6661511/144036700-b91377bd-7a68-47c2-8387-c770258eaa9b.mov

styrelius (approved):
Nice! 🎨 When you say these changes "are visually proofed" - does that mean that someone from design has looked through this and approved? Or what does it mean? 😳

gustaveen:
> Nice! 🎨 When you say these changes "are visually proofed" - does that mean that someone from design has looked through this and approved? Or what does it mean? 😳

Ah sorry, was a bit vague of me, just meant that I didn't just change the values but made sure to look through all components :)

src/client/components/inputs/index.tsx
robinandeer: No IE11 compatibility: https://caniuse.com/css-rrggbbaa 😅 

Did we decide on that? Let's write what the browser requirements are in Notion!

robinandeer (approved):
👍 As long as we write down in Notion that we don't support IE11 ;)

I have a PR in web-onboarding that uses `URLSearchParams` but not sure if we need to polyfill that, that's why I'm asking

src/client/components/inputs/index.tsx
gustaveen: Sorry not sorry IE11 🤪 We've said that we shouldn't support IE11, looking at the stats we have 0,09% of visitors on IE11 since 1/6. But let me know if you disagree :) This would be more of a graceful degradation since it's only for a hover state and not a breaking change.

gustaveen:
> 👍 As long as we write down in Notion that we don't support IE11 ;)
> 
> I have a PR in web-onboarding that uses `URLSearchParams` but not sure if we need to polyfill that, that's why I'm asking

Maybe we can have it as a point for the tech forum tomorrow what we want to support just to make sure we are aligned. We could for example have a functional site for older browsers but no need for them to look the same as up-to-date browsers :)

PR #704, opened 2021-12-03
Fix / Broken redirect from /offer to landing page

styrelius:
## What?

Fix the redirect from `/offer` to landing page which happens if you try to go to `/offer` without a valid session.

## Why?

The reason this was broken is that a switch had been made from the old `useCurrentLocale` hook to the new one in `/OfferNew/index.tsx`, but the `Redirect` path wasn't changed, which means it tries to use an object as the locale part of the path.

## Screenshots / recordings 
**Broken redirect:**

https://user-images.githubusercontent.com/42962286/144586748-460f98c1-0e40-4f8a-a1a0-3221f684a5f7.mov

**Fixed redirect:**

https://user-images.githubusercontent.com/42962286/144586766-cbed76ec-5e7f-444b-8a0a-cc094def9bd4.mov


robinandeer (approved):
Nice catch! 

sports-football

PR #706, opened 2021-12-06
Fix / Offer page can't be scrolled

styrelius:
## What?

Attempt to fix bug with `/offer` page being initially un-scrollable when it's first loaded after going through the onboarding flow.

**Bonus:** Minor refactoring to follow the general code style/patterns in the codebase.

## Why?

After introducing the use of a `Modal` in `EmbarkRoot` the `useLockBodyScroll` cleanup function will apply style `overflow: hidden;` to the body on `/offer` page. This is caused by the `EmbarkRoot` component (for some reason) setting this style on the body. Since we don't know what problem was solved when this styling was introduced (two years ago) we'll try to just remove it and do some thorough checking of a review app in different viewports/browsers/devices.

Another solution would be to change the logic of the `useScrollLock` `useLayoutEffect` cleanup function - instead of setting `document.body.style.overflow` to the previous (as in before the hook was called) body `overflow` style we could just set it to the default `overflow` style of the body`. 

But even if we change the hook we should probably not just set new styles of the body the way we do in `EmbarkRoot` so let's just remove that first :)

**Ticket(s): [GRW-610]**

### 👉  [Review app](https://web-onboardi-grw-610-fi-bxhcro.herokuapp.com/se-en/new-member)

[GRW-610]: https://hedvig.atlassian.net/browse/GRW-610?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/components/ModalNew.tsx
+++ b/src/client/components/ModalNew.tsx
@@ -86,8 +86,7 @@ export const Modal: React.FC<ModalProps> = ({
   onClose,
   children,
 }) => {
-  useLockBodyScroll({ lock: isVisible })
-
+  useLockBodyScroll({ isLocked: isVisible })
   return (
     <Wrapper
       initial={'hidden'}
--- a/src/client/components/ModalNew.tsx
+++ b/src/client/components/ModalNew.tsx
@@ -5,7 +5,7 @@ import React from 'react'
 import { useLockBodyScroll } from '../utils/hooks/useLockBodyScroll'
 import { CloseButton } from './CloseButton/CloseButton'
 
-export interface ModalProps {
+export type ModalProps = {
   isVisible: boolean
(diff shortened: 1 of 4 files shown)


dendoesit (approved):
Looks good. Tested on my machine as well and there are no extra scrollbars. 
Nice codescouting. 🥳

guilhermespopolin (approved):
Tested provided Review App on Chrome and Safari 🚀

styrelius:
Tested it on my phone (iPhone) in both Safari and Chrome, and in Chrome and Firefox on desktop and it seems to look the same as in staging 😀

PR #707, opened 2021-12-08
Fix / PHONE_OPENS_AT textkey

amandaflod:
## What?

Created new textkey in lokalise and downloaded it. With that, some more textkeys followed along. 

## Why?

The textkey for PHONE_OPENS_AT was missing. 

**Ticket(s): []**


dendoesit (approved):
Great. 👍

PR #710, opened 2021-12-13
Added locale to quoteBundle query

FredrikAreschoug:
@mightyost found a bug:
image (1)

This should fix the localization on extra buildings :slightly_smiling_face:


robinandeer:
Hmm nice catch -- we are moving away from these endpoints of course but could be nice to fix still!

PR #720, opened 2021-12-16
forever page light theme

dendoesit:
## What?

Forever page light theme

## Why?

Align to brand
https://web-onboardi-grw-667-fo-maw2fk.herokuapp.com/se-en/forever

Zak confirmed.

**Ticket(s): []**
(https://hedvig.atlassian.net/browse/GRW-667)

image


robinandeer:
There's a few elements that are invisible now on the page 😅 -- the link in the footer text as well as the language toggle.

Screenshot 2021-12-17 at 09 39 05

The footer text also fails accessibility -- not enough contrast between foreground/background.

Screenshot 2021-12-17 at 09 39 20

robinandeer:
nit: something that isn't anything new is to also add a placeholder to the input, I think that's missing today

gustaveen (approved):
👍

robinandeer (changes requested):
Like I mentioned in my comments -- currently, links and languages switcher are not visible on page so need to fix that before merge :)

dendoesit:
The review link is updated. Added the color change to language picker. 
The input field contains a placeholder but it's currently empty. Do you know what a code example looks like so I can put it inside @robinandeer  ?

gustaveen:
Ah good catch, missed the language picker 🙌 We should use the same styling as the footer component you added in Embark for the paragraph and link. @dendoesit 

Screenshot 2021-12-17 at 13 23 08

robinandeer:
> The review link is updated. Added the color change to language picker.
> The input field contains a placeholder but it's currently empty. Do you know what a code example looks like so I can put it inside @robinandeer ?

I believe this could be a good placeholder example: `7VEKCAG`

robinandeer (dismissed):
Cool! I think this works 🦾 

I still think the overall look of the page is a bit dull with the new theme -- but that's maybe something for @zakkjellin to look into in the future!

dendoesit:
It is my understanding that Zak plans to do a redesign on this but it's just not a priority right now. 

added @gustaveen's input and the placeholder 👍

robinandeer (approved):
Nice updates!

PR #723, opened 2021-12-21
Hotfix / Reorder routes to not override existing endpoints

robinandeer:
## What?

Also update graphql client to match new schema changes.

## Why?

Current setup overrides existing endpoints:

- /new-member/sign/success
- /new-member/sign/fail

**Ticket(s): []**

### DIFF
--- a/src/routes.tsx
+++ b/src/routes.tsx
@@ -81,16 +81,16 @@ export const serverSideRoutes: ServerSideRoute[] = [
     titleTextKey: 'OFFER_PAGE_TITLE',
   },
   {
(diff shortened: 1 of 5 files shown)


src/routes.tsx
```diff
@@ -170,6 +165,11 @@ export const reactPageRoutes: ReactPageRoute[] = [
     Component: SignLoading,
     exact: true,
   },
+  {
+    path: localePathPattern + '/new-member/(offer|sign)/:id',
```
guilhermespopolin, line 169: So this route was overriding `/new-member/sign/success`, right?!

guilhermespopolin (approved):

PR #726, opened 2021-12-22
Temporarily disable phone number

robinandeer:
## What?

Implement feature flag for customer service phone  …
Add color prop to correctly style language picker.
Update snapshots.

## Why?

Key solutions not available during x-mas.

**Ticket(s): []**

### DIFF
--- a/src/client/pages/Embark/LanguagePicker.tsx
+++ b/src/client/pages/Embark/LanguagePicker.tsx
@@ -5,39 +5,48 @@ import { Link, useLocation } from 'react-router-dom'
 import { useMarket } from 'components/utils/CurrentLocale'
 import { languagePickerData } from './languagePickerData'
 
+type ColorProps = {
+  color: 'black' | 'white'
+}
+
 const Wrapper = styled.div`
   display: flex;
   align-items: center;
   height: 1.5rem;
 `
 
-export const ActiveOption = styled.div`
-  color: ${colorsV3.gray100};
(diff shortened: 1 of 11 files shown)


guilhermespopolin (approved):
🤔 So in order to show/remove the phone number from the App we'd need to configure a env variable and rebuild the app? Would be possible to get customer service availability from an api?

robinandeer:
@guilhermespopolin maybe an API will be overkill but would be nice to have an interface where other people can toggle the default state of features like this. Much like they can add content in the CMS.

I talked to @styrelius about making that part of the project to setup A/B testing!

robinandeer:
@styrelius here's the review app: https://web-onboardi-hotfix-tem-yusvbv.herokuapp.com/se/new-member/

robinandeer:
Seems to work as expected

Screenshot 2021-12-23 at 10 11 49
Screenshot 2021-12-23 at 10 11 37
Screenshot 2021-12-23 at 10 10 49

src/client/pages/Embark/LanguagePicker.tsx
```diff
@@ -5,51 +5,63 @@ import { Link, useLocation } from 'react-router-dom'
 import { useMarket } from 'components/utils/CurrentLocale'
 import { languagePickerData } from './languagePickerData'
 
+type ColorProps = {
```
dendoesit, line 8: this is great. 👍

styrelius (approved):
👌

PR #705, opened 2021-12-03
Feature / Discount tag checkout view

amandaflod:
## What?

Add Discount tag to checkout drawer! It is the same tag that is also on the offer page in the "sidebar", so I created a new component to be used in both places. 

**Ticket(s): [GRW-416]**

Screenshot 2021-12-03 at 11 31 10

Here is a link to the review app!
https://web-onboardi-grw-416-di-zcjbet.herokuapp.com/se/new-member/debugger

[GRW-387]: https://hedvig.atlassian.net/browse/GRW-387?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[GRW-416]: https://hedvig.atlassian.net/browse/GRW-416?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #727, opened 2021-12-22
refactor(GRW-689): propagate OfferNew/Checkout changes to Offer/Checkout

guilhermespopolin:
## What?

We're in the process of integrate _QuoteCart API_ into _Offer_ page. Part of the strategy of this major refactor consists on doing the changes in an isolated mode. For example: If we need to do some changes on _OfferNew/Checkout/index.tsx_ we duplicate that file into _Offer/Checkout/index.tsx_ and make the changes there. That also means that any changes occurred in parallel on old offer files (_OfferNew/*_) needs to be propagated on the new files (_Offer/*_) as well. With that said, this PR is about to propagate recent changed occurred on _Checkout_ (#705 - add DiscountTag in Checkout view) into new _Checkout_ files. 

## Why?

We need to keep propagating these changes so we don't loose any added feature/fix during _QuoteCart API_ integration.

Screenshot 2021-12-27 at 18 27 42

**Ticket(s): [GRW-689](https://hedvig.atlassian.net/browse/GRW-689)**

### DIFF
--- a/src/client/components/CampaignBadge/CampaignBadge.tsx
+++ b/src/client/components/CampaignBadge/CampaignBadge.tsx
@@ -40,7 +40,6 @@ export const CampaignBadge: React.FC<CampaignBadgeProps> = ({
       id: quoteCartId,
       locale: isoLocale,
     },
-    onCompleted: () => console.log('called applied'),
   })
   const campaignText = campaignData?.quoteCart.campaign?.displayValue ?? ''
 
--- a/src/client/components/CampaignBadge/CampaignBadge.tsx
+++ b/src/client/components/CampaignBadge/CampaignBadge.tsx
@@ -4,8 +4,6 @@ import styled from '@emotion/styled'
 import { Badge, BadgeProps } from 'components/Badge/Badge'
 
 import { useAppliedCampaignNameQuery } from 'data/graphql'
-import { OfferData } from 'pages/OfferNew/types'
-import { isNorwegianBundle } from 'pages/OfferNew/utils'
 import { useTextKeys } from 'utils/textKeys'
 import { useCurrentLocale } from 'l10n/useCurrentLocale'
 
@@ -22,14 +20,14 @@ const CampaignBadgeStyled = styled.div`
 `
 
 export type CampaignBadgeProps = Omit<BadgeProps, 'children'> & {
-  offerData: OfferData
   quoteCartId: string
+  isNorwegianBundle: boolean
 }
 
 export const CampaignBadge: React.FC<CampaignBadgeProps> = ({
   className,
   quoteCartId,
-  offerData,
+  isNorwegianBundle,
   ...others
 }) => {
   const { isoLocale } = useCurrentLocale()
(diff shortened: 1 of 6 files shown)


src/client/components/CampaignBadge/CampaignBadge.tsx
```diff
@@ -0,0 +1,64 @@
...
+      id: quoteCartId,
+      locale: isoLocale,
+    },
+  })
+  const campaignText = campaignData?.quoteCart.campaign?.displayValue ?? ''
+
+  if (loading || !campaignText) return null
+
+  const discounts: Array<React.ReactNode> = [
+    ...(isNorwegianBundle(offerData)
```
guilhermespopolin, line 49: Can someone explain the use case for this check please? I'd love to be able to remove the dependency with `offerData` here.

src/client/graphql/AppliedCampaign.graphql
```diff
@@ -0,0 +1,7 @@
+query AppliedCampaignName($id: ID!, $locale: Locale!) {
```
robinandeer, line 1: nit: consider naming variable "quoteCartId" or query "AppliedQuoteCartCampaignName"

src/client/components/CampaignBadge/CampaignBadge.tsx
robinandeer: Hmmmm you could consider only passing the required data: `<CampaignBadge isNorwegianBundle={true/false}>` instead of `offerData`.

This is a static discount applied to all bundles in NO. However, this really should be something we apply automatically in Giraffe -- I think that's the solution going forward. So perhaps we keep this code until then?

robinandeer (approved):
Nice work!

src/client/pages/Offer/Introduction/Sidebar/index.tsx
```diff
@@ -190,31 +176,22 @@ export const Sidebar: React.FC<SidebarProps> = ({
...
-    ...(isNorwegianBundle ? [textKeys.SIDEBAR_NO_BUNDLE_CAMPAIGN_TEXT()] : []),
-    ...(campaignText ? [campaignText] : []),
-  ]
-
   const showRemoveCampaignButton = campaign != null
   const isDiscountPrice =
     campaign?.incentive?.__typename === 'MonthlyCostDeduction' ||
     campaign?.incentive?.__typename === 'PercentageDiscountMonths' ||
-    isNorwegianBundle
+    isNorwegianBundle(offerData)
```
guilhermespopolin, line 183: I'm open to suggestions for a variable name that would hold the value returned by `isNorwegianBundle` predicate, so I wouldn't need to calculate the value [twice](https://github.com/HedvigInsurance/web-onboarding/pull/727/files#diff-34ac2504068f1bf7af38914d4352369f38bad63521df594af26a944fb2425992R193).
It feels the correct name is the one we're already using for the function 🤣

robinandeer: haha yeah I agree that on second though the name is more like a variable. Function could be called e.g. `checkIsNorwegianBundle`. But let's leave it for now :P

robinandeer (approved):

PR #728, opened 2021-12-23
Feature / New routes

styrelius:
## What?

- New structure for our route declarations: one single `routes` array instead of two arrays of routes (on for client-side routes and one for server routes).
- New routes for the new Connect Payment At Sign checkout pages (_in progress_)
- New way to easily set a route as hidden in prod - by just adding `isHiddenInProd: true` to the `clientRouteData` of the route object.
- New components `CheckoutDetails` and `CheckoutPayment` (which don't have any content yet).

## Why?

- If we want a route to exist both on the client and on the server we currently have to add the exact same route path twice, which is annoying and also means we might mistype and make the routes not be the same.
- Seemed like a good opportunity to refactor `routes.tsx` in connection to adding the new checkout/signing routes.
- Since the new checkout routes, like some other routes, shouldn't be available in production (yet) it felt like a good idea to look into how we can prevent pages from being accessible already on the route level, instead of in the components rendered by those routes.
  But maybe it's better to use the feature flagging functionality in some way for this? 🤔 

**Ticket(s): [GRW-643]**

<hr />

### [Review app](https://web-onboardi-grw-643-fe-kq4xih.herokuapp.com/se-en/new-member)
This review app simulates production by having `APP_ENVIRONMENT` env variable set to `production` in Heroku. Hence the new routes (and the /debugger route) will just result in a redirect to `/new-member`.
Test the branch locally to see that the all routes that should work in development, do.

[GRW-643]: https://hedvig.atlassian.net/browse/GRW-643?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/App.tsx
```diff
@@ -34,15 +37,31 @@ export const App: React.ComponentType<StorageState> = ({ session }) => {
...
-                    path={path}
-                    exact={exact}
-                    component={Component}
-                    render={render}
-                  />
-                ))}
+                {routes.map(({ path, clientRouteData }) => {
+                  if (!clientRouteData || !(typeof path === 'string')) {
+                    return null
+                  }
```
robinandeer, line 43: Why not filter out these routes? `routes.filter().map()`?

src/client/App.tsx
```diff
@@ -34,15 +37,31 @@ export const App: React.ComponentType<StorageState> = ({ session }) => {
...
+                  const {
+                    exact,
+                    Component,
+                    render,
+                    isHiddenInProd,
+                  } = clientRouteData
+
+                  if (isProductionEnvironment && isHiddenInProd) {
+                    return null
+                  }
```
robinandeer, line 53: Could use `Array.filter` for excluding these as well, no?

src/server/serverEntry.tsx
```diff
@@ -121,8 +121,11 @@ router.post(
   handleAdyen3dsPostRedirect,
 )
 
-serverSideRoutes.forEach((route) => {
-  router.get(route.path, getPage(route))
+routes.forEach((route) => {
```
robinandeer, line 124: Why don't we filter out the non-production routes here as well?

src/routes.tsx
```diff
@@ -36,269 +38,331 @@ enum EmbarkStory {
...
+}
+
+export type Route = {
+  path: string | RegExp
+  serverRouteData?: ServerSideRoute
+  clientRouteData?: ClientSideRoute
 }
 
-export const reactPageRoutes: ReactPageRoute[] = [
+export const landingRoute = '/new-member'
```
robinandeer, line 61: Does this need to be exported?

src/routes.tsx
```diff
@@ -36,269 +38,331 @@ enum EmbarkStory {
...
+export const routes: Route[] = [
+  {
+    path: onboardingLocaleBaseRoute,
+    clientRouteData: {
+      Component: Landing,
+      exact: true,
+    },
+  },
   {
     path: localePathPattern + '/forever/:code?',
```
robinandeer, line 73: nit: compose with template literal string like the other routes

src/routes.tsx
```diff
@@ -36,269 +38,331 @@ enum EmbarkStory {
...
+    clientRouteData: {
+      Component: OfferPage,
+      exact: true,
+    },
   },
   {
-    path: localePathPattern + '/new-member/debugger',
-    Component: Debugger,
-    exact: true,
+    path: `${onboardingLocaleBaseRoute}/checkout/details`,
```
robinandeer, line 186: These URLs will need to include quote cart ID :)

```suggestion
    path: `${onboardingLocaleBaseRoute}/offer/:cartId/checkout/details`,
```

src/routes.tsx
```diff
@@ -36,269 +38,331 @@ enum EmbarkStory {
...
+  },
+  {
+    path: /^(?!\/?new-member-assets).*$/,
+    serverRouteData: {
+      titleTextKey: 'FOUR_OH_FOUR_PAGE_TITLE',
+      status: 404,
+    },
   },
   {
     path: '/*',
```
robinandeer, line 362: Shouldn't we use this `serverRouteData` here?

```
serverRouteData: {
       titleTextKey: 'FOUR_OH_FOUR_PAGE_TITLE',
       status: 404,
     },
```

robinandeer (commented):
Good work!

mehdielket (approved):
Nice initiative and cleanup 👏

src/client/App.tsx
styrelius: Good point 👍

src/client/App.tsx
```diff
@@ -34,15 +37,31 @@ export const App: React.ComponentType<StorageState> = ({ session }) => {
...
+
+                  return (
+                    <Route
+                      key={path}
+                      path={path}
+                      exact={exact}
+                      component={Component}
+                      render={render}
+                    />
```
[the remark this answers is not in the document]
styrelius: Good point, I'll look into that 👍

src/server/serverEntry.tsx
styrelius: I'm not sure I understand what you mean? The non-production routes are still needed, or? Do you mean the client-side routes?

src/routes.tsx
styrelius: Doesn't seem like it. If I remember it correctly I had some idea of using that in the million places where we hard code `/new-member` throughout the code base, but probably then thought that would mean an unreasonably big PR 🙃

src/routes.tsx
styrelius: Ah, true! Because we don't want to store it in e.g. local storage?

guilhermespopolin (dismissed):
Nice refactoring 🚀

src/routes.tsx
styrelius: For the `/*` paths? The way it is currently is that this one is only part of the `reactPageRoutes` array. And then the server route instead uses the regex `/^(?!\/?new-member-assets).*$/` to set the status to `404` for unknown paths. So I don't think we need to add it again. But maybe I'm misunderstanding. 

Also I'm not entirely sure why this is even done this way in the first place haha 🙈 Apart from the fact that we don't want a `404` status when requesting images from **https://www.hedvig.com/new-member-assets**, like [https://www.hedvig.com/new-member-assets/landing/hedvig_table_landscape_medium.jpg](https://www.hedvig.com/new-member-assets/landing/hedvig_table_landscape_medium.jpg)

src/server/serverEntry.tsx
robinandeer: Hmm... I guess I was thinking of mimicking the same setup as for client side routes, i.e. filter out routes that match: `isProductionEnvironment && isHiddenInProd`

src/routes.tsx
robinandeer: I think it makes for a more standard and cleaner solution :)

- We can build a link to the offer page in a super simple way
- We can server render the page if we want to
- No need to clean up stale state in local storage

src/routes.tsx
robinandeer: Hmm okey, yeah I guess it's fine then :)

src/server/serverEntry.tsx
styrelius: Ah, OK, I see. Maybe we can skip that for now though?

src/routes.tsx
styrelius: Made the routes like `${onboardingLocaleBaseRoute}/checkout/details/:id` since `${onboardingLocaleBaseRoute}/offer/:cartId/checkout/details` looked a bit strange to me when actually having a quote cart ID. Named the URL parameter `id` since that's what we call it for the `(offer|sign)` route, but I think your `cartId` is probably better.

src/routes.tsx
```diff
@@ -36,269 +38,332 @@ enum EmbarkStory {
...
+    clientRouteData: {
+      Component: OfferPage,
+      exact: true,
+    },
   },
   {
-    path: localePathPattern + '/new-member/debugger',
-    Component: Debugger,
-    exact: true,
+    path: `${onboardingLocaleBaseRoute}/checkout/details/:id`,
```
robinandeer, line 187: Should we keep "offer" in the URL to make more sense which "id" we are referring to?

```suggestion
    path: `${onboardingLocaleBaseRoute}/offer/:id/checkout/details`,
```

styrelius: As I wrote [above](https://github.com/HedvigInsurance/web-onboarding/pull/728#discussion_r778842864) i think the URL didn't look great with `/offer` in it. But as I also wrote, I think the URL parameter should probably be called `cartId` to make it clearer what we're referring to. I went with `id` though, since that's what you call it for the `/(offer|sign)` route.

src/client/App.tsx
styrelius: OK so TypeScript wasn't very happy about that solution which made the code very ugly with a lot of type assertions etc. so I skipped this.

src/server/serverEntry.tsx
styrelius: So I actually tried to do this but didn't really get anywhere since Koa isn't behaving at all like I expect it to, and since we're moving away from this setup anyway I don't think it's worth spending more time on that :)

src/routes.tsx
mehdielket: Could you please elaborate on why you think it "looked a bit strange to me when actually having a quote cart ID"?

src/client/App.tsx
mehdielket: For reference, I faced this issue before and I think type predicates are the recommended way to approach filtering. https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates

src/routes.tsx
styrelius: @mehdielket I went with this option since I think the first one looks cluttered with that long ID in the middle. Given that we’re reading from left to right it seems reasonable to me to have the information more relevant to the user as far left as possible, for the few users who would have a glance on the URL. Those users probably are more interested in seeing that this page is part of the **checkout** than reading some string of characters that makes no sense to them.

You could also argue that this follows the pattern of `/offer/:quoteCartId`, which I think is as consistent and make as much sense as having the checkout routes as sub-routes to the  `/offer/:quoteCartId` route.

See [this Notion page](https://www.notion.so/hedviginsurance/OTHER-NOTES-e4d18b73f46f4f188d1daad52cab12a9) for a comparison between the URL's (and the same text with my thoughts as the one posted above).

mehdielket (approved):
🚀

guilhermespopolin (approved):

styrelius:
@guilhermespopolin Exactly! And that would get pretty messy when I just wanted to make thins less messy haha 😅 Maybe worth it, but probably not since we're moving away from this setup altogether

PR #729, opened 2021-12-27
GRW-698 / Feature / Clear session/token when creating new quote cart

robinandeer:
## What?

Add new clear session method on storage.
Add new hook to create quote cart.

## Why?

We want to avoid situations where we mix up tokens for different sessions.

**Ticket(s): [GRW-698]**

[GRW-698]: https://hedvig.atlassian.net/browse/GRW-698?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/utils/hooks/useCreateQuoteCart.ts
+++ b/src/client/utils/hooks/useCreateQuoteCart.ts
@@ -1,0 +1,19 @@
+import { useCallback } from 'react'
+import { useCurrentLocale } from 'l10n/useCurrentLocale'
+import { useCreateOnboardingQuoteCartMutation } from 'data/graphql'
+import { useStorage } from './../StorageContainer'
+
+export const useCreateQuoteCart = () => {
+  const storage = useStorage()
+  const { apiMarket, isoLocale } = useCurrentLocale()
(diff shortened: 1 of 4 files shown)


src/client/utils/hooks/useCreateQuoteCart.ts
```diff
@@ -0,0 +1,19 @@
+import { useCallback } from 'react'
+import { useCurrentLocale } from 'l10n/useCurrentLocale'
+import { useCreateOnboardingQuoteCartMutation } from 'data/graphql'
+import { useStorage } from './../StorageContainer'
+
+export const useCreateQuoteCart = () => {
```
mehdielket, line 6: This is a great way to guarantee we do this all the time! Would have been great to be able to "ban" the use of the mutation directly

mehdielket (approved):
Clean 🚀

src/client/utils/hooks/useCreateQuoteCart.ts
robinandeer: Hmm yeah indeed, we could write our own eslint rule for that 😅

PR #730, opened 2021-12-28
GRW-599 / Feature / Refactor feature toggle to use env variables

robinandeer:
## What?

- Create checkFeature utility.
- Maintain same hook as before.
- Update tests.
- Use feature flag to enable new embark flow.
- Add feature flags to hedvig window config.

This is how it works:

1. Set environment variables like: `FEATURE_QUOTE_CART_API=SE,DK`
2. The server will pick this up and pass it along to `window.hedvigClientConfig.features`
3. This status of a feature can then be checked with: `const isEnabled = checkFeature(Feature.QUOTE_CART_API)`

Question:
Do we need a way to set default values in code?

## Why?

We want more flexibility to enable/disable feature per deployment.
We want to access feature toggles outside React.

**Ticket(s): []**


src/client/l10n/locales.ts
```diff
@@ -17,6 +18,8 @@ export const LOCALE_URL_PARAMS = [
 export type LocaleUrlParams = typeof LOCALE_URL_PARAMS
 export type LocaleLabel = LocaleUrlParams[number]
 
+export type MarketLabel = SharedMarketLabel
```
robinandeer, line 21: Re-export to maintain the same module imports as before.

src/client/pages/Landing/Landing.tsx
```diff
@@ -156,6 +156,7 @@ export const Landing: React.FC = () => {
   const textKeys = useTextKeys()
   const { marketLabel, path: localePath } = useCurrentLocale()
   const variation = useVariation()
+  const productsData = useMemo(getProductsData, [])
```
robinandeer, line 159: Memoize for performance reasons.

src/client/pages/Landing/landingPageData.ts
```diff
@@ -25,11 +27,13 @@ type ProductData = {
 
 type ProductsData = Record<Market, ProductData>
 
-export const productsData: ProductsData = {
+export const getProductsData = (): ProductsData => ({
```
robinandeer, line 30: Convert to a function since `checkFeature` is accessing window but the module is loaded on the server (for some reason...)

webpack/webpack.config.client.development.js
```diff
@@ -35,9 +35,6 @@ module.exports = webpackConfig({
         NODE_ENV: JSON.stringify('development'),
         SENTRY_DSN: JSON.stringify(process.env.SENTRY_DSN),
         SENTRY_ENVIRONMENT: JSON.stringify('local'),
-        ENABLE_QUOTE_CART_API: JSON.stringify(
-          process.env.ENABLE_QUOTE_CART_API,
-        ),
```
robinandeer, line 40: This is no longer needed!

src/client/utils/hooks/useFeature.ts
```diff
@@ -1,88 +1,18 @@
...
-    envs: ['staging', 'production'],
-    markets: ['SE'],
-  },
-  {
-    name: Features.CUSTOMER_SERVICE_PHONE_NUMBER,
-    envs: ['staging', 'production'],
-    markets: [],
-  },
-]
+export { Feature as Features } from 'shared/clientConfig'
```
robinandeer, line 4: Re-export to maintain same module imports

src/client/pages/Landing/landingPageData.ts
```diff
@@ -25,11 +27,13 @@ type ProductData = {
...
   SE: [
     {
       id: 'swedishNew',
-      linkSlug: '/new',
+      linkSlug: checkFeature(Feature.QUOTE_CART_API)
+        ? '/home-accident-needer'
```
gustaveen, line 35: Is this the route for the QuouteCart Embark flow? :)

.env.example
```diff
@@ -18,4 +18,9 @@ HEROKU_SLUG_COMMIT=<commit hash>
 SENTRY_AUTH_TOKEN=<sentry auth token>
 
 # Optional - feature toggles
-ENABLE_QUOTE_CART_API=false
+FEATURE_QUOTE_CART_API=<SE,DK,NO>
```
gustaveen, line 21: Could it be an idea to have a value for enabling all markets? Or maybe that's when you should remove the feature flag instead :)

gustaveen (commented):
Great stuff Robin, looks good overall 🙌 This would mean you need to have your local `.env` up to date right? So might be good to have default values as you said :)

src/client/pages/Landing/landingPageData.ts
robinandeer: Exactly, I followed the same standard as other markets to align SE

.env.example
robinandeer: haha yeah I guess so -- we can revisit that when we support say 10 markets

src/server/config.ts
```diff
@@ -19,3 +20,24 @@ export const ADYEN_ORIGIN_KEY = process.env.ADYEN_ORIGIN_KEY!
...
 export const APP_ENVIRONMENT = process.env.APP_ENVIRONMENT ?? 'development'
+
+const DEFAULT_FEATURES: FeatureMap = {
+  [Feature.CHECKOUT_CREDIT_CHECK]: ['NO'],
+  [Feature.CHECKOUT_UPSELL_CARD]: ['SE'],
+  [Feature.CUSTOMER_SERVICE_PHONE_NUMBER]: ['SE'],
+  [Feature.OFFER_PAGE_INSURANCE_TOGGLE]: ['SE'],
+  [Feature.QUOTE_CART_API]: [],
+  [Feature.TEST_FEATURE]: [],
+}
```
robinandeer, line 31: @gustaveen here's some default values we can use if you haven't configured your local `.env`

.env.example
mehdielket: I personally think it's preferable to be intentional when enabling features. Having empty mean all markets is risky and can lead to developer error.

I think we should rather remember to cleanup feature flags for things we know we will keep long term. For instance the quoteCartAPI flag should go away as soon as we are live on all markets.

src/client/utils/hooks/useFeature.ts
```diff
@@ -1,88 +1,18 @@
...
-        env,
-        market: marketLabel,
-      })
-      return [...acc, isEnabled]
-    } else {
+  return features.map((feature) => {
+    const enabledMarkets = window.hedvigClientConfig.features[feature]
+
+    if (!enabledMarkets) {
       throw new Error(`Unknown feature: ${feature}`)
```
mehdielket, line 13: Would this crash the app if a given feature flag is missing in the env file of an environment?

src/server/config.ts
mehdielket: I feel like this defeats the point of "not having to touch the code to change feature flags. This section may age badly if we all at some point have local env files.

.env.example
robinandeer: I agree, I was actually thinking that empty means disable the feature in all markets 😅 at least we need some way to disable a feature that is turned on by default.

src/server/config.ts
robinandeer: Hmm... any env variable will override these defaults so in my mind we don't need to change code.

But perhaps it's like you say that it's mostly now in the beginning. I'm just thinking if we miss one feature flag and that feature accidentally gets turned off in production let's say. This would make sure that we have a default for stable features.

src/client/utils/hooks/useFeature.ts
robinandeer: nope, since we have the default values. this will only fail if you pass a feature that isn't defined in the enum (TS protects us there as well).

I just kept the same behaviour as in your original code :)

mehdielket (approved):
Ship it 🚀

.env.example
gustaveen: Sounds good! 🙌

src/server/config.ts
gustaveen: That works IMO! 👌

gustaveen (dismissed):
⛳️

robinandeer:
@gustaveen after talking to @mehdielket we agreed that defining defaults in code might not be the way to go. It's also recommended not to inherit env vars according to the package we are using: https://github.com/motdotla/dotenv#should-i-have-multiple-env-files

Instead I updated the `.env.example` files so you can just copy paste all those feature flags over to your local `.env` and get it setup just like in staging env!

I've already added all the environment variable to the different environments in Heroku.

Please re-review!

gustaveen:
> @gustaveen after talking to @mehdielket we agreed that defining defaults in code might not be the way to go. It's also recommended not to inherit env vars according to the package we are using: https://github.com/motdotla/dotenv#should-i-have-multiple-env-files
> 
> Instead I updated the `.env.example` files so you can just copy paste all those feature flags over to your local `.env` and get it setup just like in staging env!
> 
> I've already added all the environment variable to the different environments in Heroku.
> 
> Please re-review!

Sounds reasonable! It's not that many so should be fine to keep the example file up to date 👍

gustaveen (approved):

PR #731, opened 2021-12-28
Avoid automatic quote cart re-creation

robinandeer:
## What?

Don't automatically re-create a quote cart unless specifically triggered.

## Why?

If the component is mounted a second time, we re-run use effect hook.

**Ticket(s): []**


dendoesit (approved):
nice ⭐

PR #732, opened 2021-12-28
GRW-561/input fields

dendoesit:
https://hedvig.atlassian.net/browse/GRW-651

## What?

Update to input fields.
Add stories in storybook for input fields.
Remove variants of input fields as we no longer support dark version. 
Update WarningTriangle icon
Add Helper Text option
Remove autoComplete and pattern as it was not currently being used . ( correct me if i'm wrong)
Remove using Field from Formik and istead use normal input with props from Formik

- [x] revert changes made on checkout for demo purposes.
- [x]  remove RawInputField from files
- [x] fix Edit Modal.
- [x] add Validation to Details Modal.
- [x] dropdown styling
- [x] fix the details modal in the new offer-debugger

## Why?

New redesign according to Figma : https://www.figma.com/file/zlX4FcM5dMRz9stsCNQJXB/Web-UI-Kit?node-id=1438%3A2616

image

Changed the ssn field in the checkout form in order to push a review app with a disabled button for testing purposes.

Review App: 

https://web-onboardi-grw-651-in-iwbres.herokuapp.com/se-en/new-member/debugger


src/client/components/inputs/index.tsx
```diff
@@ -222,29 +223,32 @@ export const RawInputField: React.FC<React.InputHTMLAttributes<
...
   showErrorIcon?: boolean
+  disabled?: boolean
 }> = ({
   errors,
+  disabled,
   showErrorIcon = false,
+  helperText,
   label,
   className,
   variant = 'light',
```
robinandeer, line 237: do we still need this prop?

src/client/components/inputs/index.tsx
```diff
@@ -222,29 +223,32 @@ export const RawInputField: React.FC<React.InputHTMLAttributes<
...
+      <StyledRawInput disabled={disabled} {...props} />
     </Wrapper>
-    {errors && <ErrorText variant={variant}>{errors}</ErrorText>}
+    {helperText && <HelperText>{helperText}</HelperText>}
+    {errors && (
+      <ErrorText>
+        <WarningTriangle size={13} />
+        {errors}
+      </ErrorText>
+    )}
```
robinandeer, line 251: From figma it looks like the error replaces the helper text - or have you discussed something else with Zak?

robinandeer (commented):
Nice update!

Don't we use the `WarningTriangle` in some error modals => should they change as well?

src/client/components/inputs/index.tsx
dendoesit: yeah not sure about this one yet. Asked Zak in Figma but he is in vacation so still waiting on a reply for this.

src/client/components/inputs/index.tsx
dendoesit: this was just added. we didn't have it before and we need it according to design. 

I added it separate from ErrorText just to make it more clear.

src/client/components/inputs/index.tsx
gustaveen: true, we should probably not display the helper text and error message

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
```diff
@@ -206,6 +206,8 @@ export const UserDetailsForm: React.FC<Props> = ({
         name="ssn"
         id="ssn"
         type="text"
+        helperText={'info message'}
```
gustaveen, line 209: Is this just to demo the helperText?

gustaveen (commented):
Nice Dan! 🙌 Did Zak conclude that we should remove the dark variant?

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
dendoesit: yes. just for demo . will revert back after we validate the PR.

mehdielket:
I just tested this and I can notice the following:

- This seems to break the look of the `<InputField>` component. It is used in the Edit Details Modal as well as the debugger. Apart from the look, one issue is that Formik errors are displayed by default and in English since we have no translations. We used to only show a warning icon on the right of fields.

Screenshot 2021-12-29 at 10 27 31

Screenshot 2021-12-29 at 10 30 49

- On the checkout form, there is a fair amount of jumping due to the errors showing under the inputs. Also the warning sign does not seem to be aligned to the text.

https://user-images.githubusercontent.com/89857278/147647771-5e523c77-7fc1-4d4a-ab45-cca04b41c24f.mov

- Lastly, it would be great to offer an extensive Storyblock story that allows to play we these inputs and show their different states. There is an exisiting one for `<InputField>` but none for `<RawInputField>`

robinandeer:
> there is a fair amount of jumping due to the errors showing under the inputs

aha, yeah I think we should always take up the same space to avoid layout shifts like that. This is how I worked it out in the Next.js POC:

https://github.com/HedvigInsurance/racoon/blob/main/apps/onboarding/src/components/input-field.tsx#L39

src/client/components/inputs/index.tsx
```diff
@@ -41,65 +41,47 @@ export const masks: Record<string, Mask> = {
 
 type variantType = 'light' | 'dark'
```
guilhermespopolin, line 42: If we still end up needing this, maybe convert it to an Enum and/or name it like `VariantType` to easily distinguish it between types and variables.

guilhermespopolin (commented):
Nice work. Just added a couple of comments.

src/client/components/inputs/index.tsx
dendoesit: oh.. that makes sense. i'll remove that.

guilhermespopolin (commented):
Just added a couple of comments :)

Additionally, I'm not sure if we need to confirm some details with Zack when it comes about error inputs like:

1. What should happen with the border of an error input when we hover/focus it? Should it be black or continue red?
2. Should the label of an error input also be in red?

Bellow I just record those scenarios comparing our component against Material UI's just to illustrate the questions:

**Ours**
ours


**Material's**
materials

src/client/components/inputs/index.tsx
```diff
@@ -188,120 +191,80 @@ export interface CoreInputFieldProps {
...
   onBlur?: FieldInputProps<string>['onBlur']
+  onFocus?: ((e: FocusEvent<HTMLInputElement>) => void) | undefined
 }
 
 export interface TextInputProps extends CoreInputFieldProps {
   showErrorMessage?: boolean
   touched?: boolean
-  errors?: string
-  variant?: variantType
+  errors?: string | boolean
```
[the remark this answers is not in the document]
dendoesit: The reason behind this is that sometimes we have errors that are explicit -> strings . And sometimes we just want to show the error triangle so a boolean would do just that. 
Helper text should be separate as it is independent of error.

dendoesit:
@guilhermespopolin thanks for the review.
Regarding error state, according to the design the input title should remain black on error. Not sure about what the color on focus should be, will confirm with @zakkjellin . My opinion is that it should be red as well. Will fix comments shortly.

dendoesit:
Update: fixed the disable state for input fields as well. Got approval from Zak. 
Removed the last 2 items in the description to new tasks so we can have focused tickets. 
Ready for review.

guilhermespopolin (dismissed):

mehdielket:
I think you should take a look at the house extra-buildings section. You can get to that from the debugger by choosing house
Screenshot 2022-01-12 at 09 43 17

Here is how it used to look
Screenshot 2022-01-12 at 09 47 39

robinandeer:
@dendoesit has the review app stopped working?

dendoesit:
@robinandeer it looks like it. 
Updated with new link : https://web-onboardi-grw-651-in-iwbres.herokuapp.com/se-en/new-member/debugger

robinandeer (commented):
Should we update the debugger to either have a light background or light labels?

Screenshot 2022-01-17 at 16 48 06

The spacing in the details modal is still a little tight when there's an error shown. I feel bad for pointing it out but we should have a liiiiiitle more patience to get it right :D

Screenshot 2022-01-17 at 16 49 36

dendoesit:
@robinandeer I have added a separate ticket for redoing the errors on the details modal as I didn't want to make this a bigger PR than it already is. the debugger is a quick fix 👍 . I would go for light labels.

src/client/pages/Debugger/ActualDebugger.tsx
```diff
@@ -4,13 +4,18 @@ import { colorsV3 } from '@hedviginsurance/brand'
...
+import { Label } from 'components/inputs/index'
 import { QuoteData } from './components/QuoteData'
 import { Session } from './components/Session'
 
 const Wrapper = styled.div`
   max-width: 900px;
   margin: 0 auto;
   color: ${colorsV3.gray100};
+
+  ${Label} {
```
guilhermespopolin, line 16: I guess declaring a dependency with a component just for styling purposes like that can be a little bit hard to follow since `Label` is not explicitly present in `ActualDebugger` scope (`ActualDebugger` > `OfferData` > `InputField` > `Label`). But I don't know how to reach the same result in a different way, so...

I guess the problem is that `body` `background` is set to black in globalStyles so everything inherits that color. So maybe in this case the more suitable solution would be change Debuggers background. Any thoughts?!

Either way, I'm ok with this solution :)

guilhermespopolin (approved):
What a precise set of changes 😆 

Screenshot 2022-01-18 at 15 42 37

src/client/pages/Debugger/ActualDebugger.tsx
dendoesit: When implementing this change I initially thought it would be easier to just change the Label. 
I agree it is not the best approach, however I feel like this will be the only exception that will be made to change the Label color. 
Also, worth mentioning that the Debugger is used internally so we can change the background/ label colors how we feel like it whenever we need to.

guilhermespopolin (approved):

dendoesit:
Closing this one as it is better if we release it together with the new errors. 
Related task here : https://github.com/HedvigInsurance/web-onboarding/pull/781

PR #734, opened 2021-12-29
Store campaign code in cookie

robinandeer:
## What?

Update `CampaignCode` interface to match existing cookie-based implementation.

## Why?

Makes sense to keep using the existing solution unless we have a reason to move away from it.

**Ticket(s): []**


dendoesit (approved):
nice ⭐

guilhermespopolin (approved):

PR #735, opened 2021-12-30
refactor(GRW-604): integrate QuoteCart API in change offer start date operation 

guilhermespopolin:
## What?

Integrate QuoteCart related API on change start date operation

## Why?

- It's part of the QuoteCart API project.

**Ticket(s): [GRW-604](https://hedvig.atlassian.net/browse/GRW-604)**


src/client/pages/Offer/Checkout/index.tsx
```diff
@@ -236,13 +236,6 @@ export const Checkout = ({
   const isDiscountMonthlyCostDeduction =
     campaign?.incentive?.__typename === 'MonthlyCostDeduction'
 
-  const offerData = useMemo(
```
guilhermespopolin, line 239: I needed to remove this as we're calculating `offerData` info twice (at `Offer` and `Checkout` levels) and while changing star date were triggering an update of `offerData` on `Offer` level the same wasn't happening at `Checkout` level due the memoization, which were resulting in a inconsistent sate between the date pickers present on `Sidebar` and `Checkout` components.

src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
```diff
@@ -0,0 +1,146 @@
...
+  const textKeys = useTextKeys()
+  const { isoLocale } = useCurrentLocale()
+
+  const [isLoading, setIsLoading] = React.useState(false)
+
+  const [setStartDate] = useSetStartDateMutation()
+
+  const checked = !quote.startDate
+
+  const toggle = async () => {
```
guilhermespopolin, line 96: This is pretty much a copy of the component placed inside `OfferNew`. The major is this method which uses the new api.

src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
```diff
@@ -0,0 +1,380 @@
...
+        return textKeys.SIDEBAR_STARTDATE_CELL_VALUE_NEW()
+      }
+
+      return format(dateValue, getDateFormat(marketLabel)!, {
+        locale: dateLocale!,
+      })
+    }
+  }
+
+  const handleSetDate = async (newDateValue: Date | null) => {
```
guilhermespopolin, line 215: This is a copy of `StartDate` component placed inside `Offer`. The major difference is this method which uses the new API.

src/client/pages/Offer/Introduction/Sidebar/StartDateLabelSwitcher.tsx
```diff
@@ -0,0 +1,112 @@
+import React from 'react'
```
guilhermespopolin, line 1: This is just a copy of the same component placed inside `Offer`. Nothing was changed.

src/client/pages/Offer/Introduction/Sidebar/utils.ts
```diff
@@ -0,0 +1,75 @@
+import React from 'react'
```
guilhermespopolin, line 1: This is just a copy of the same file placed inside Offer. Nothing was changed.

src/client/pages/Offer/Introduction/Sidebar/StartDateLabelSwitcher.tsx
mehdielket: If nothing has changed, can we keep the one in OfferNew instead to avoid any changes that need to happen in 2 places?

src/client/pages/Offer/Introduction/Sidebar/utils.ts
mehdielket: Same, can we point to the old file instead to avoid fixing bugs or making changes in two places?

src/client/components/DateInput/index.tsx
```diff
@@ -106,7 +106,7 @@ export const DateInput: React.FC<DateInputProps> = ({
       >
         <CalendarWrapper>
           <Dayzed
-            date={new Date()}
+            date={date}
```
mehdielket, line 109: I believe that in some markets, the startDate is mandatory, the original code sets automatically that value to today. What's the idea behind removing this?

mehdielket (commented):
Nice work. Added a couple of comments.

Also, I understand how conservative you have been with this PR trying to only fix the API and I think that's probably the right thing to do. But I think if you feel that there are some "opportunistic refactorings" possible, I think it would be nice too (eg: moving away from offerData, passing more specific props to components...)

src/client/components/DateInput/index.tsx
guilhermespopolin: 🤔 What happened is that I've checked `dayzed` library docs and [date](https://www.npmjs.com/package/dayzed#date) prop defaults to `new Date()` and it's used to _calculate what month to display on initial render_. So by keeping it as `new Date()` was resulting on the scenario where the month displayed when I open the date picker doesn't correspond with the selected date, which I think it's a not so good UX.

src/client/pages/Offer/Checkout/index.tsx
robinandeer: Hmmm I guess our assumptions were wrong :P

src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
```diff
@@ -0,0 +1,146 @@
...
+  setShowError,
+}) => {
+  const textKeys = useTextKeys()
+  const { isoLocale } = useCurrentLocale()
+
+  const [isLoading, setIsLoading] = React.useState(false)
+
+  const [setStartDate] = useSetStartDateMutation()
+
+  const checked = !quote.startDate
```
robinandeer, line 94: nit: consider re-naming `isChecked`

robinandeer (approved):
Great work! Every time I review this code I wonder how we made things so difficult for ourselves -- at least you got rid of the "refetchOfferData" props! :D

(haven't tested review app)

robinandeer:
Super nice work!

robinandeer (approved):
Super nice work!

PR #737, opened 2021-12-30
Add OneTrust domain to CSP to be able to load the cookie banner

gustaveen:
## What?
Add OneTrust domain to CSP to be able to load the cookie banner

## Why?
To be able to test the Cookie banner


dendoesit (approved):
👍

PR #738, opened 2022-01-03
Set features as disabled if no env var is found

robinandeer:
## What?

Set feature flags as disabled if no environment variable is found.

## Why?

Production env didn't have some feature flags set and that caused a runtime error.

**Ticket(s): []**


dendoesit (approved):

gustaveen (approved):
👍

guilhermespopolin (approved):

PR #739, opened 2022-01-03
Make sure we set correct html lang

gustaveen:
## What?
Make sure we set correct html lang

## Why?
To identify the language of the site to give correct info to screen readers, search engines and such. The cookie banner also follows the set html lang, that's how I discovered it :) 

https://user-images.githubusercontent.com/6661511/147944389-b953c19d-7247-4ea8-ad9f-e0e26c12b0d2.mov


robinandeer (approved):
Nice catch!

PR #741, opened 2022-01-04
fix(Checkout): updated toggle method

guilhermespopolin:
## What?

- Changed `Checkout` toggle function so url gets updated properly; `redirect` is being used instead of `push` because IMHO as a user, I don't expect a UI like a _Sidebar_ incrementing browser's history list.

## Why?

- Given a link that triggers Checkout opening (/sign), when user clicks on 'close' button
app goes back in the browser history which results on an empty browser tab; By clicking on `Checkout` close button, user should be able to see/interact with _Offer_ page.

### Current Behaviour

https://user-images.githubusercontent.com/19200662/148051562-452831a5-db9e-4114-99c1-6d8cfc9bdb74.mov

 
### New Behaviour

https://user-images.githubusercontent.com/19200662/148051576-b181a588-09f5-415c-a9e3-d34d74e72101.mov

### DIFF
--- a/src/client/pages/Offer/index.tsx
+++ b/src/client/pages/Offer/index.tsx
@@ -45,11 +45,13 @@ const createToggleCheckout = (
   locale?: LocaleLabel,
 ) => (isOpen: boolean) => {
   if (isOpen) {
-    history.push(
+    history.replace(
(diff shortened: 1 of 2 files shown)


robinandeer (approved):
Yeah I also think this makes sense while we move towards the new checkout flow

PR #743, opened 2022-01-10
refactor/GRW-640: redesign CancellationOptions

guilhermespopolin:
## What?

Update `<CancellationOptions />` to be compliance with its redesign proposed as part of _Connect Payment at Sign_ feature.

## Note for reviewers

You might notice that there are two 'versions' of `CancellationOptions` and `StartDate`  files updated on this PR:
- `pages/OfferNew/Introduction/Sidebar/CancellationOptions.tsx` and `pages/Offer/Introduction/Sidebar/CancellationOptions.tsx`
- `pages/OfferNew/Introduction/Sidebar/StartDate.tsx` and `pages/Offer/Introduction/Sidebar/StartDate.tsx`

That's because we're keeping duplicated files while transitioning to the new _QuoteCart API_ and, as a consequence, we need to propagate changes in the old files into the new ones. So even though the PR states _15 files changed_, only the following are worth reviewing:

- `src/client/components/Switch.tsx`
- `src/client/pages/OfferNew/Introduction/Sidebar/CancellationOptions.tsx`
- `src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx`

**Ticket(s): [GRW-640](https://hedvig.atlassian.net/browse/GRW-640)**

## How to test

Access the review app linked below and complete the switcher flow in _Embark_. In case you don't have an account with a different insurance provider, you can just skip the part of the flow which is about to collect info about your current insurance provider by selecting any insurance provider at the beginning of the flow and then clicking on [Skip] button:


src/client/pages/OfferNew/Introduction/Sidebar/StartDate.tsx
```diff
@@ -141,17 +141,6 @@ const DateInputModalWrapper = styled.div<{
   border-radius: 8px;
 `
 
-const StyledDateInput = styled(DateInputForm)<{
```
guilhermespopolin, line 144: I'm removing this part since these styles are never being applied since `DateInputForm` [doesn't accept a className prop](https://emotion.sh/docs/styled#:~:text=styled%20can%20style%20any%20component%20as%20long%20as%20it%20accepts%20a%20className%20prop.)

src/client/components/Switch.tsx
```diff
@@ -23,31 +23,57 @@ const SwitchNobble = styled(Tick)`
...
+/**
+ * The idea is to build a custom checkbox component that is accesibible and that
+ * provides a good UX when it comes about controlling it: e.g it should possible
+ * to bind it with a label e toggle it by clicking on the label.
+ * We could solve the accessible part of the problem by using aria-* attributes.
+ * However, the UX part would require a bunch of JS to emmit the events properly.
+ * With that said, one well adopted solution is to use a hidden checkbox to handle
+ * check/unchecked state that's visually hidden along with your custom implmentation
+ * of a checkbox (styles, animation, etc)
+ */
```
robinandeer, line 35: Can we explain this in a shorter way? Otherwise, I love the initiative!

src/client/components/Switch.tsx
```diff
@@ -23,31 +23,57 @@ const SwitchNobble = styled(Tick)`
...
-        borderColor: colorsV3.gray300,
-      },
-    }}
-    transition={{ duration: 0.1 }}
-    role="checkbox"
-    aria-checked={value}
-  >
-    <SwitchNobble />
-  </SwitchContainer>
+export const Switch: React.FC<SwitchProps> = ({
```
robinandeer, line 53: nit: skip `React.FC`

src/client/components/Switch.tsx
```diff
@@ -23,31 +23,57 @@ const SwitchNobble = styled(Tick)`
...
+  className,
+  value,
+  onChange,
+}) => (
+  <>
+    <InclusiveHiddenCheckbox type="checkbox" onChange={onChange} />
+    <SwitchContainer
+      className={className}
+      aria-hidden
```
robinandeer, line 62: nit: I would just change to

```suggestion
      aria-hidden={true}
```

src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
```diff
@@ -68,14 +63,26 @@ export const CancellationOptions: React.FC<CancellationOptionsProps> = ({
...
+  })
+}
+
+type QuoteCancellationOptionProps = {
   quoteCartId: string
   quote: OfferQuote
   setShowError: (showError: boolean) => void
 }
+
 const QuoteCancellationOption: React.FC<QuoteCancellationOptionProps> = ({
```
robinandeer, line 85: nit: consider removing `React.FC`

src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
```diff
@@ -88,6 +95,7 @@ const QuoteCancellationOption: React.FC<QuoteCancellationOptionProps> = ({
   const [editQuote] = useEditBundledQuoteMutation()
 
   const isChecked = !quote.startDate
+  const labelContent = getLabelContent(textKeys, quote)
```
robinandeer, line 98: nit: I would call this variable like `checkboxLabel`

src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
```diff
@@ -114,25 +122,14 @@ const QuoteCancellationOption: React.FC<QuoteCancellationOptionProps> = ({
   return (
     // TODO: This logic needs some clarification
```
robinandeer, line 123: Is this comment still relevant?

robinandeer (commented):
Nice work! I can't see much to improve but added just a few comments

Will test in review app before approving :)

src/client/components/Switch.tsx
```diff
@@ -23,31 +23,44 @@ const SwitchNobble = styled(Tick)`
...
-    role="checkbox"
-    aria-checked={value}
-  >
-    <SwitchNobble />
-  </SwitchContainer>
+export const Switch = ({ className, ...checkboxProps }: SwitchProps) => (
+  <>
+    <InclusiveHiddenCheckbox type="checkbox" {...checkboxProps} />
+    <SwitchContainer
+      className={className}
```
styrelius, line 48: Why do we want to be able to pass `className`?

styrelius (commented):
Great job! 🌟 

One thing I noticed though is that the checkmark in the implementation looks a lot thinner than in Figma.

Code: 
image


Figma: 
image


Also one question below, out of curiosity.

src/client/components/Switch.tsx
guilhermespopolin: TL;DR: To be able to control positioning/spacing outside `Switch`

I'm doing this because I wanted to be able to change its style in case I need to. Particularly in this case, I wanted to add some [spacing between](https://github.com/HedvigInsurance/web-onboarding/pull/743/files#diff-e5ba53806ba548c23022cb64eabc499c9f06d6cc1651ab055a8fea78717b6ec9R33) the checkbox (`Switch`) and the label. One might argue that I could solve the problem by adding the margin directly into `Switch` component but I think that by doing it in this way you end up with a more flexible component, leaving positioning/spacing matters to consumers, that usually have specifics about this, to handle them.

guilhermespopolin:
> Great job! 🌟
> 
> One thing I noticed though is that the checkmark in the implementation looks a lot thinner than in Figma.
> 
> Code: image
> 
> Figma: image

You're right :)
Gonna address that.

guilhermespopolin:
Screenshot 2022-02-04 at 14 22 53

FYI @styrelius

src/client/components/Switch.tsx
styrelius: Hm, I see. I generally think allowing for a `className` prop makes component a little _too_ flexible. Especially in this case where it's only for one purpose. How about a `gap` prop instead, or something similar?

robinandeer: I guess this would be a great discussion for a design system 😉 

I personally really like the ability to `styled(Switch)` to adapt styles to the context. I worry more about too much flexibility when it comes to things like our Storyblok blocks.

src/client/pages/OfferNew/Introduction/Sidebar/CancellationOptions.tsx
```diff
@@ -1,47 +1,39 @@
 import styled from '@emotion/styled'
-import { colorsV3 } from '@hedviginsurance/brand'
 import { format } from 'date-fns'
```
robinandeer, line 2: nit: make more explicit

```suggestion
import { format as formatDate } from 'date-fns'
```

robinandeer (dismissed):
I meeeeean, as far as I can tell it looks good. Let's do some proper testing when we merge!

Maybe add another end2end test or something

src/client/components/icons/Tick.tsx
```diff
@@ -1,6 +1,11 @@
 import React from 'react'
+import { colorsV3 } from '@hedviginsurance/brand'
 
-export const Tick: React.FC<{ color?: string }> = ({ color = '#fff' }) => (
+export type Props = {
```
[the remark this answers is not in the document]
guilhermespopolin: Oh, I thought it was a convention. I vaguely remember pointing out the same thing in a PR someone else have opened and I got the understanding that it was a convention. But I agree with you: I personally prefer `<component-name>Props` for prop type naming.

src/client/components/Switch.tsx
guilhermespopolin: Yeah, a `gap` prop would work. But then I would need to make decisions about where the gap should be applied (top, right, bottom, left, or a mix) or if should I implemented using it `margin` or `padding`, etc. I guess it's just simpler to enable opt-in style customisation of all sort. But as @robinandeer mentioned, that's debatable.

styrelius: Realized that I maybe wasn't super clear. I was thinking that we'd probably never have _just_ a checkbox so a `gap` prop would rather be for the container - with both the checkbox and label - to set a space between them. There could also be a prop to decide whether or not the label comes before or after the checkbox. 

But yes, let's maybe not continue that discussion here 😄

guilhermespopolin: Oh, I see it now and I agree with you. I have came with the same conclusion while working on this. It's on my "personal backlog" open another PR with such refactoring and other stuff as well; e.g  rename `Switch` --> `Checkbox`.

src/client/components/icons/Tick.tsx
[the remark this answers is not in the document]
styrelius: Haha, yes we had that discussion and I pointed out that the pattern we follow in most files is naming the type for a component's props simply `Props`. But that's definitely not a hill I'm willing to die on so maybe we should change pattern 😄

src/client/pages/OfferNew/Introduction/Sidebar/CancellationOptions.tsx
```diff
@@ -70,78 +61,80 @@ export const CancellationOptions: React.FC<CancellationOptionsProps> = ({
...
-        .then(() => refetch())
-        .catch(handleFail)
-        .finally(() => setIsLoading(false))
+      if (!checked) {
+        await removeStartDate({
+          variables: {
+            quoteId: quote.id,
+          },
+        })
+        await refetch()
```
styrelius, line 111: Question: Is this `await` necessary when this is the last thing that happens in this block?

styrelius (commented):
One thought: Have you checked with Zak or Amy if they're happy with the solution of using the display name for our insurance as name of the user's current insurance? Our product names are pretty long ("Home Insurance Renter" etc.) and probably not exactly the name of their current insurance 🤔

src/client/pages/OfferNew/Introduction/Sidebar/CancellationOptions.tsx
guilhermespopolin: Yes. That's because checkbox state (checked/unchecked) is defined based API data. So here we're executing some mutations when user clicks on the checkbox and then we need to wait until we get new data changed by the mutations before removing the loading state and show the checkbox with the correct state. This is the result in case I remove it:

https://user-images.githubusercontent.com/19200662/152811783-85229ba6-4fe7-4b02-86a4-7d5189616385.mov

guilhermespopolin:
> One thought: Have you checked with Zak or Amy if they're happy with the solution of using the display name for our insurance as name of the user's current insurance? Our product names are pretty long ("Home Insurance Renter" etc.) and probably not exactly the name of their current insurance 🤔

I assumed Zak kind of know this already but I'm checking now.

guilhermespopolin:
> One thought: Have you checked with Zak or Amy if they're happy with the solution of using the display name for our insurance as name of the user's current insurance? Our product names are pretty long ("Home Insurance Renter" etc.) and probably not exactly the name of their current insurance 🤔

I've checked with them and they are ok with this solution

src/client/pages/OfferNew/Introduction/Sidebar/CancellationOptions.tsx
styrelius: Ah yes 👍

styrelius (commented):
For some reason the spinner looks a little crazy in the checkout slide-in thingy 🤪 

https://user-images.githubusercontent.com/42962286/152950987-34b5cb6d-ebdf-438b-a0cf-6a583960dd58.mov

robinandeer (approved):
Code-wise looks good to me :D

I wasn't able to test it - when I select "Skip" I don't see the cancellation option...

robinandeer:
@guilhermespopolin now it worked super well! :D Great work

PR #744, opened 2022-01-11
Grw-686/Feature/Phone Number in NO

dendoesit:
## What?

Added phone number input field for NO checkout flow. 

## Why?

We want it so we can call customers and ask them why they didn't finish the sign up process. 

**Ticket(s): [GRW-686]**

image

### [Review app]()

https://web-onboardi-grw-686-ph-gfuisv.herokuapp.com/no-en/new-member/debugger

[GRW-686]: https://hedvig.atlassian.net/browse/GRW-686?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/l10n/inputFormats.ts
```diff
@@ -30,3 +30,9 @@ export const phoneNumberFormats: Record<MarketLabel, RegExp> = {
   NO: /^(0047|\+47|47)?[2-9]\d{7}$/,
   DK: /^(0045|\+45|45)?[\d]{8}$/,
 }
+
+export const phoneNumberPlaceholder: Record<MarketLabel, string> = {
+  SE: '+46 00 000 00 00',
+  NO: '+47 000 0000',
+  DK: '+45 00000000',
+}
```
mehdielket, line 38: Are we sure users can enter the phone number with spaces in it?

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
```diff
@@ -35,6 +33,11 @@ export const emailValidation = yup
 
 export const nameValidation = yup.string().required()
 
+export const phoneValidation = yup
+  .string()
+  .required()
+  .min(10)
```
mehdielket, line 39: Would this work for other markets?

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -452,9 +464,7 @@ export const Checkout = ({
                 />
               </InnerWrapper>
               <Sign
-                canInitiateSign={
-                  canInitiateSign && !ssnUpdateLoading && !emailUpdateLoading
-                }
+                canInitiateSign={true}
```
mehdielket, line 467: Is this intentional?

src/client/pages/OfferNew/index.tsx
```diff
@@ -157,6 +157,9 @@ export const OfferNew: React.FC = () => {
 
   const handleCheckoutToggle = (open: boolean) => {
     toggleCheckout(open)
+    if (offerData) {
+      trackOfferGTM(EventName.CheckoutOpen, offerData, false)
+    }
```
mehdielket, line 162: This does not seem to belong to this PR. Was this mistakenly deleted before?

mehdielket (commented):
Added a few questions 🙂

src/client/l10n/inputFormats.ts
guilhermespopolin: I think they can't since the validation taken into place check the number of digits.
Maybe we should use a more robust validation, them insert it with or without spaces wouldn't matter.

guilhermespopolin (commented):
This error state feels weird:

- Helper text is there but it doesn't provide any help about the error;
- Helper text content became centralized;

If that's what is suggested by the desginer team, maybe we should talk with them about it.

Screenshot 2022-01-12 at 16 10 08

src/client/pages/OfferNew/Checkout/index.tsx
dendoesit: no. will fix.

src/client/pages/OfferNew/index.tsx
dendoesit: It also helps in this PR specifically because you can track the number of people that actually open the checkout  compared to the number of people who actually sign / who put in their phone number.

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
dendoesit: we currently don't have the phoneNumber in other markets and only want to put it as part of the checkout in NO. When we release the new checkout this will get removed entirely. 😄

src/client/l10n/inputFormats.ts
dendoesit: hmm. you are right. I will remove the character max length value so the user can add spaces. 
A more robust validation + error handling will come for all fields but we need the phone number in NO checkout quickly.

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
```diff
@@ -149,6 +158,26 @@ export const UserDetailsForm: React.FC<Props> = ({
     )
   }
 
+  const handlePhoneChange = (event: React.FormEvent<HTMLInputElement>) => {
+    const { value } = event.currentTarget
+    const phoneNumber = value
```
guilhermespopolin, line 163: @mehdielket if I remember correctly a discussion in one of our daily, we have a ticket to add Phone number into `Offer/Checkout` too right?!

guilhermespopolin (commented):
I just opened the review app > Checkout the sign button was already loading 🤔 
Not sure if it's something related with this PR tho.

Screenshot 2022-01-13 at 17 34 46

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
dendoesit: yes. @mehdielket already finished his part. this PR is for the one currently on prod.

dendoesit:
@guilhermespopolin thanks for reviewing. I also noticed the spinner just keeps on loading. still investigating. 😮‍💨

src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
mehdielket: 👍🏼

PR #746, opened 2022-01-11
Checkout details page sections & headlines

styrelius:
## What?

- Components specific for the **/checkout/details** page, added to Storybook
- A `Headline` component based on the headlines in the [Web UI Kit Figma](https://www.figma.com/file/zlX4FcM5dMRz9stsCNQJXB/Web-UI-Kit?node-id=1350%3A1554), also added to Storybook

**Question:** should these be split up into two separate PR's?

## Why?

- Since I needed a headline for the `SubSection` component for the **/checkout/details** page it seemed like a great opportunity to add a `Headline` component. This can then just be copy-pasted over to the UI Kit / Design System in the new monorepo, if that's the route we're taking :)

**Ticket(s): [GRW-646]**

## Screenshots / recordings 

**Checkout Details page sections:**
image

**Headline component:**

https://user-images.githubusercontent.com/42962286/148930143-82d0eca5-42c5-4f5d-a9f2-6d23f9e9cd6b.mp4

[GRW-646]: https://hedvig.atlassian.net/browse/GRW-646?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dendoesit (approved):
the variant approach is great 💯 . I think the same PR is ok.

gustaveen (commented):
Great initiative Ellen! 🙌 I'm very pro having a Headline component. We have one in market-web that I worked on a long time ago, might be nice to merge the functionality of the two :D https://github.com/HedvigInsurance/market-web/blob/master/src/components/Heading/Heading.tsx 
I think it's neat to have some props for `color`, `as` (to render different heading elements) and possibly text alignment. :)

gustaveen:
Come to think of if the design always aligns as the ui kit where a L heading in mobile is always a L heading in desktop? I'm trying to get hold of Zak to discuss some of the ui kit implementation and market-web updates so I'd be happy to iron that out :) I know we in market-web use different heading sizes in mobile and desktop quite often.

src/client/components/CheckoutDetails/PageSection.stories.tsx
```diff
@@ -0,0 +1,45 @@
...
+import styled from '@emotion/styled'
+import { Story } from '@storybook/react'
+import { colorsV3 } from '@hedviginsurance/brand'
+import { SubSection } from './SubSection'
+import { PageSection } from './PageSection'
+
+const { gray700 } = colorsV3
+
+const Text = styled.p`
+  font-size: 14px;
```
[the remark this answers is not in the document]
styrelius: No :) This was just out of habit, but maybe it doesn't matter that much since it's just example text in a Storybook story?

src/client/components/Headline/Headline.tsx
```diff
@@ -0,0 +1,118 @@
...
+import { MEDIUM_SCREEN_MEDIA_QUERY } from 'utils/mediaQueries'
+
+const { FAVORIT } = fonts
+const { gray900 } = colorsV3
+
+export type Props = {
+  variant: 'xl' | 'l' | 'm' | 's' | 'xs' | 'overline'
+  children: React.ReactNode
+}
+
```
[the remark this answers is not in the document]
styrelius: I was thinking about that, but then I also thought that we might want to differentiate the headlines more in the future and that's easier to do if we don't have this. But the `margin: 0` and `font-family` will probably be consistent so I'll fix that, thank you 🙏

src/client/components/Headline/Headline.tsx
```diff
@@ -0,0 +1,118 @@
+import React from 'react'
+import styled, { StyledComponent } from '@emotion/styled'
+import { fonts, colorsV3 } from '@hedviginsurance/brand'
+import { MEDIUM_SCREEN_MEDIA_QUERY } from 'utils/mediaQueries'
+
+const { FAVORIT } = fonts
+const { gray900 } = colorsV3
+
+export type Props = {
+  variant: 'xl' | 'l' | 'm' | 's' | 'xs' | 'overline'
```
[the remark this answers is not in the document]
styrelius: Isn't there a point of the headline levels being consistent? 🤔 Maybe we just want to be able to pass `as="h1"` since we don't have an h1 here?

styrelius:
@gustaveen 
> Come to think of if the design always aligns as the ui kit where a L heading in mobile is always a L heading in desktop? I'm trying to get hold of Zak to discuss some of the ui kit implementation and market-web updates so I'd be happy to iron that out :) I know we in market-web use different heading sizes in mobile and desktop quite often.

Ah, OK I just followed what's in the UI Kit Figma. Seems like it would be easiest to use the same headline in desktop and on mobile, but then maybe the sizes aren't right? 🤔 

> I think it's neat to have some props for color, as (to render different heading elements) and possibly text alignment. :)

Yes, about the color - I started adding that but then I was like "But we don't even use light text for the headlines anymore". That's of course incorrect though (the hero section on the /offer page is one example) so I'll fix that :)

src/client/components/Headline/Headline.tsx
```diff
@@ -0,0 +1,118 @@
+import React from 'react'
+import styled, { StyledComponent } from '@emotion/styled'
+import { fonts, colorsV3 } from '@hedviginsurance/brand'
+import { MEDIUM_SCREEN_MEDIA_QUERY } from 'utils/mediaQueries'
+
+const { FAVORIT } = fonts
+const { gray900 } = colorsV3
+
+export type Props = {
```
guilhermespopolin, line 9: nit:
```suggestion
export type HeadlineProps = {
```

Otherwise we'd need to rename it in the scenario we import this type inside a file that already define a `Props` type (which seems to be the case on most of our components files).

src/client/components/Headline/Headline.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+    },
+  },
+  args: {
+    variant: 'xl',
+    children: 'This is a headline',
+  },
+}
+
+const Template: Story<HeadlineProps> = ({ variant, children }) => (
+  <>
```
guilhermespopolin, line 20: nit: no need for a fragment here.

src/client/components/CheckoutDetails/PageSection.tsx
```diff
@@ -0,0 +1,19 @@
+import styled from '@emotion/styled'
+import { colorsV3 } from '@hedviginsurance/brand'
+import { MEDIUM_SCREEN_MEDIA_QUERY } from 'utils/mediaQueries'
+
+const { gray200 } = colorsV3
+
+export const PageSection = styled.div`
+  background-color: ${gray200};
+  width: 100%;
+  max-width: 628px;
```
guilhermespopolin, line 10: Maybe use relative units rather them `px`?

src/client/components/CheckoutDetails/SubSection.tsx
```diff
@@ -0,0 +1,32 @@
...
+}
+
+const Wrapper = styled.div`
+  display: grid;
+  gap: 8px;
+`
+const Card = styled.div`
+  background-color: ${white};
+  width: 100%;
+  padding: 8px 16px;
```
guilhermespopolin, line 20: Maybe use relative units rather them `px`?

src/client/components/Headline/Headline.tsx
styrelius: I think we generally name the main props object type in a file `Props` - but if it feels like it's very annoying to import it `as HeadlineProps` we can of course adopt a pattern where we always name the props object type in a more specific way..

src/client/components/CheckoutDetails/PageSection.tsx
styrelius: Hmm, should we use that for `width` and `height` values as well? 🤔 Maybe we should, I'm trying to think what happens with the layout if the browser font size changes. But yes, maybe it's good that everything scales in the same way.

src/client/components/Headline/Headline.tsx
guilhermespopolin: No, It's not very annoying do the renaming 😄 
Let's keep it as it its

src/client/components/Headline/Headline.tsx
```diff
@@ -0,0 +1,118 @@
...
+  variant: 'xl' | 'l' | 'm' | 's' | 'xs' | 'overline'
+  children: React.ReactNode
+}
+
+const HeadlineXL = styled.h2`
```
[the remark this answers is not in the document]
guilhermespopolin: 🤔 I was thinking in maybe limit `as` prop value to available headings (h1 - h6) and assign `h1` as the default value. I guess that to turn a `span` into a heading we'd need to provide aria attributes to it (if we care about acessibility).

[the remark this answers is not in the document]
gustaveen: Yup, I think we could set a default heading element, could be a `h2` or a `h3` since they are probably more often used. We could also make the as prop required if we want to enforce you to think on what heading level you're going to use.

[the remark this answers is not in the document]
styrelius: I think we should probably enforce it, and not have `h1` as default since that makes the risk high that we'll end up with several `h1` on the same page? But OK, I don't know enough about this `as` prop - I think that made the component render as that element? I.e. no need for aria attributes?

src/client/components/CheckoutDetails/PageSection.stories.tsx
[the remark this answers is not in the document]
styrelius: Never mind this comment that comes from my lazy side, I fixed it :)

guilhermespopolin (approved):
🚀

src/client/components/Headline/Headline.tsx
```diff
@@ -11,9 +12,13 @@ export type Props = {
   children: React.ReactNode
 }
 
-const HeadlineXL = styled.h2`
+const baseStyles = css`
```
gustaveen, line 20: Great! 🎉 Isn't the `font-weight` the same for all as well? :)

styrelius: Yep, but that's something that feels like there's a high chance we want to change? But maybe I'm thinking about it wrong haha 🙈

gustaveen: I don't think we use any bold text right, or at least we shouldn't looking at the brand guidelines :D
Screenshot 2022-01-13 at 16 44 57

src/client/components/Headline/Headline.tsx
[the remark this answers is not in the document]
gustaveen: Precisely, it changes what element it's rendered, so as you say no need for aria attributes :) Makes sense to not have it as an `h1` 👍

src/client/components/Headline/Headline.tsx
styrelius: Ah OK, cool! 👍  Where are these guidelines? Haven't seen them 😳

src/client/pages/CheckoutDetails/components/SubSection.tsx
```diff
@@ -0,0 +1,34 @@
...
+const { white } = colorsV3
+
+type Props = {
+  headlineText: string
+  children: React.ReactNode
+}
+
+const Wrapper = styled.div`
+  display: grid;
+  gap: 8px;
```
robinandeer, line 15: Should this be `rem`s?

src/client/components/Headline/Headline.tsx
```diff
@@ -0,0 +1,133 @@
...
+  children: React.ReactNode
+}
+
+type ColorProp = Pick<Props, 'colorVariant'>
+
+type StyleProps = {
+  as: Props['headingLevel']
+} & ColorProp
+
+type GetBaseStylesParams = ColorProp
```
guilhermespopolin, line 23: 🤔 Maybe this type can be removed?! Maybe it can be simpler without it:

What do you think of this?
```
type ColorVariant = 'dark' | 'light'

export type Props = {
  variant: 'xl' | 'l' | 'm' | 's' | 'xs' | 'overline'
  headingLevel: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6'
  colorVariant: ColorVariant
  children: React.ReactNode
}

type StyleProps = {
  as: Props['headingLevel']
  colorVariant: ColorVariant
}

const getBaseStyles = (colorVariant: ColorVariant) => {
  const color = colorVariant === 'light' ? gray100 : gray900

  return css`
    margin: 0;
    padding: 0;
    font-family: ${FAVORIT};
    font-weight: 400;
    color: ${color};
  `
}
```

guilhermespopolin (approved):
Well done 🚀

src/client/components/Headline/Headline.stories.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react'
```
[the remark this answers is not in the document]
styrelius: Ah, cool with the typing 👌 I just copied from some other Storybook files since I haven't really worked much with Storybook (for some reason I keep on forgetting that we use that).

Sure, I can make another story as well, thought it was fine to just change the background but maybe it's easier to have two stories :)

styrelius:
> I'm just wondering if you based it on some prior art we can refer to so we don't have to write too many custom guidelines/docs when we build this out. A few examples I found:
> 
> * MUI (https://mui.com/components/typography/#changing-the-semantic-element)
> * Shopify Polaris (https://polaris.shopify.com/components/titles-and-text/heading)
> * Braid (https://seek-oss.github.io/braid-design-system/components/Heading)

@robinandeer That's a great point, and no I didn't. I just took the `components` file we had in the `/OfferNew` folder and cleaned it up :) My thinking was just making something that's at least better than what we have now, and that can easily be moved, reused and changed. But maybe we should look at references more before doing things like this.

robinandeer:
> @robinandeer That's a great point, and no I didn't. I just took the `components` file we had in the `/OfferNew` folder and cleaned it up :) My thinking was just making something that's at least better than what we have now, and that can easily be moved, reused and changed. But maybe we should look at references more before doing things like this.

Sure, I think that's a good approach! We don't need to overthink things now but before we get to far I think it would be a good idea to make sure we don't "reinvent the wheel" :)

styrelius:
> Sure, I think that's a good approach! We don't need to overthink things now but before we get to far I think it would be a good idea to make sure we don't "reinvent the wheel" :)

No of course :) An addition to what I wrote earlier is that my approach was to base this on what we have in the [Figma Web UI Kit](https://www.figma.com/file/zlX4FcM5dMRz9stsCNQJXB/Web-UI-Kit?node-id=1350%3A1554). So the names of the possible `variant` prop is based on what the headlines are called there for example. But that's also good to discuss of course, where the source of truth should be (in code or in Figma or somewhere else).

src/client/components/Headline/Headline.tsx
styrelius: OK so I got rid of the `GetBaseStylesParams` type since it was just the same as `ColorProp` type. But I decided to not go with this suggestion since I generally like having "named" parameter for functions (i.e. passing an object as a single parameter). This is of course more beneficial when having multiple function parameters (which we don't in this case) but it's an easier principle to follow to just always do it like that than changing it when we want to add more parameters. And no, I'm not super consistent with this haha 🙈 But I try!

PR #749, opened 2022-01-13
GRW-717 / Fix / Routing bugs

styrelius:
## What?

- `serverRouteData` with `titleTextKey` added to landing page route in order to fix the page title being "undefined" for a moment when page loads (before the page title rendered with react-helmet on the client pops up)
  This weirdly enough also seems to fix the bug of the `/connect-payment` page described below 🤔 
- A check if we have both window (window as any).Sentry is added before trying to run `(window as any).Sentry.captureMessage(e)` in `adtraction.ts`

## Why?

The `/connect-payment` page i broken. When signing in Denmark and Norway the redirect is just a white page (seems to not be an issue for Sweden though). This is caused by `ADT` being `undefined` (line 116 in `adtraction.ts`) and when moving into the `catch` on line 130 everything crashes since `(window as any).Sentry` is also `undefined` 🤷‍♀️ 

**Ticket(s): [GRW-717]**

### [Review app](https://web-onboardi-fix-routin-lem6er.herokuapp.com/dk-en/new-member)

[GRW-717]: https://hedvig.atlassian.net/browse/GRW-717?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/utils/tracking/adtraction.ts
+++ b/src/client/utils/tracking/adtraction.ts
@@ -128,6 +128,8 @@ export const adtraction = (
     adt.Tag.tp = getBundleAdtractionProductValue(offerData)
     adt.Tag.doEvent()
   } catch (e) {
-    ;(window as any).Sentry.captureMessage(e)
+    if (typeof window !== 'undefined' && (window as any).Sentry) {
+      ;(window as any).Sentry.captureMessage(e)
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


guilhermespopolin (approved):
Looks good. Just didn't understand why adtraction in only defined for Swedish market.. 🤔

PR #750, opened 2022-01-13
refactor(GRW-669): quote details deprecated field

guilhermespopolin:
## What?

This intents to remove the dependency with `BundledQuote.quoteDetails` field in `pages/Offer/*` files.

## Why?

`BundledQuote.quoteDetails` is going to be deprecated in the _QuoteCart_ API.

**Ticket(s): [GRW-669](https://hedvig.atlassian.net/browse/GRW-669)**


mehdielket (commented):
One comment

src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
```diff
@@ -124,13 +120,9 @@ const QuoteCancellationOption: React.FC<QuoteCancellationOptionProps> = ({
             return textKeys.SIDEBAR_REQUEST_CANCELLATION_GENERIC_INSURANCE()
           }
 
-          if (isNorwegianHomeContents(quote.quoteDetails)) {
-            return textKeys.SIDEBAR_REQUEST_CANCELLATION_HOME_INSURANCE()
-          }
-
-          if (isNorwegianTravel(quote.quoteDetails)) {
-            return textKeys.SIDEBAR_REQUEST_CANCELLATION_TRAVEL_INSURANCE()
-          }
```
robinandeer, line 133: Do we not need these checks anymore?

src/client/pages/Offer/Introduction/HeroOfferDetails.tsx
```diff
@@ -132,7 +155,7 @@ export const HeroOfferDetails: React.FC<Props> = ({
           {numberCoInsured > 0 && ` +${numberCoInsured}`}
         </NameAndCoInsured>
 
-        {address && <Address>{address}</Address>}
+        {address && <QuoteAddress>{address}</QuoteAddress>}
```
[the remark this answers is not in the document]
guilhermespopolin: 🤔 Empty strings are one of the [Javascript Falsy values](https://developer.mozilla.org/en-US/docs/Glossary/Falsy) and In general, I think empty strings are a good way to indicate 'nothing to display' since types became simpler: `string` rather than `string | null`. The exception would be scenarios where we want to do something in case that variable equals to `null`, which not seems to be the case. With that said, I'm also totally fine to change it to `string | null`.

Maybe someone else can 'untie' this decision 🤣

src/client/pages/Offer/Introduction/HeroOfferDetails.tsx
```diff
@@ -103,23 +104,45 @@ const EditDetailsButton = styled(Button)`
...
+    return interestedInFields.every((field) => quote.data[field] != null)
+  }
+
+  const parseAddress = (address: Address) => {
+    const { street, apartment, floor } = address
+
+    return street
+      .concat(floor ? `, ${floor}.` : '')
+      .concat(apartment ? ` ${apartment}` : '')
+  }
```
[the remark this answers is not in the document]
guilhermespopolin: No, I can move it outside.

🤔 It's pretty much the same implementation we're using before with `InsuranceSummaryDetails` [getAddress](https://github.com/HedvigInsurance/web-onboarding/pull/750/files#diff-7c371857b44f7fecfd1b97f2c85ceea91357515d3c034c39faaba60335e4156eL13) function. But I understand your point as this could be an opportunity to reevaluate it. Gonna give a look.

src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
guilhermespopolin: I'm glad you've pointed this out

This is an attempt of moving away from code that rely on quote's type as a consequence of a more generic approach with `quote.data`. But ofc I would like to discuss it more as I don't fully understand the business logic here.

In a scenario of Swedish market and multiple quotes ([isGenericQuote](https://github.com/HedvigInsurance/web-onboarding/pull/750/files#diff-7c371857b44f7fecfd1b97f2c85ceea91357515d3c034c39faaba60335e4156eL13) is based on the amount of quotes) we'd end up rendering the Checkbox with no label. But it looks like [switchable](https://github.com/HedvigInsurance/web-onboarding/pull/750/files#diff-7c371857b44f7fecfd1b97f2c85ceea91357515d3c034c39faaba60335e4156eL13) is always false on such cases.

Couldn't we assume `switchable` flag will take also market into account and display cancellation options as following?

If only one quote (`isGenericQuote === true`), then display one cancellation option with generic wording (`textKeys.SIDEBAR_REQUEST_CANCELLATION_GENERIC_INSURANCE()`)
else display, for each quote, a cancellation option that mentions the quote name.

src/client/pages/Offer/Introduction/HeroOfferDetails.tsx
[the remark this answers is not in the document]
robinandeer: Cool, but it that's the way it worked before we can keep it :)

src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
robinandeer: Aha, so you are proposing essentially making this component market-agnostic? At least in terms of the labels, right?

guilhermespopolin: Yes

src/client/pages/Offer/Introduction/HeroOfferDetails.tsx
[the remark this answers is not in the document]
guilhermespopolin: Yeah, I've checked but it feels any solution that takes locale into account would be premature right now so I'm just leave as it is :)

src/client/pages/Offer/utils.ts
```diff
@@ -0,0 +1,31 @@
...
+
+const parseAddress = (address: Address) => {
+  const { street, apartment, floor } = address
+
+  return street
+    .concat(floor ? `, ${floor}.` : '')
+    .concat(apartment ? ` ${apartment}` : '')
+}
+
+export const getAddress = (quotes: BundledQuote[]) => {
```
robinandeer, line 12: I would rename this file to something less generic if it only has one export:

```
src/client/pages/Offer/utils.ts => src/client/pages/Offer/getAddress.ts
```

guilhermespopolin: I see. The idea is to use that file to gradually move `OfferNew/utils` content to it. I believe that we could use the same function in [InsuranceSummaryDetails](https://github.com/HedvigInsurance/web-onboarding/blob/master/src/client/pages/OfferNew/Checkout/InsuranceSummaryDetails.tsx#L207) when the migration is done.

But I change for something more specific as well, no problem :)

PR #751, opened 2022-01-14
GRW-686/Feature/phone checkout no

dendoesit:
## What?

Added phone number input field for NO checkout flow.

## Why?

We want it so we can call customers and ask them why they didn't finish the sign up process.

Ticket(s): [GRW-686](https://hedvig.atlassian.net/browse/GRW-686)

Image: 
image

Ended up starting with a new branch in order to make the spinner in the button dissapear. This is the same ticket with all the implemented comments. 

Review app : 

https://web-onboardi-feature-ph-3l97p6.herokuapp.com/no-en/new-member/debugger


src/client/pages/OfferNew/Checkout/UserDetailsForm.tsx
```diff
@@ -149,6 +159,26 @@ export const UserDetailsForm: React.FC<Props> = ({
...
+    }
+    reallySetPhone(phoneNumber)
+    setPhoneChangeTimeout(
+      window.setTimeout(() => {
+        if (phoneValidation.isValidSync(phoneNumber)) {
+          onPhoneChange(phoneNumber)
+          setHasPhoneError(false)
+        } else {
+          setHasPhoneError(true)
+        }
```
guilhermespopolin, line 177:
```suggestion
       // This could be lift up to avoid being created every time
       // You already has access to a regex to validate phone number inside `locale` object
       const phoneValidator = yup.string().matches(locale.phoneNumber.formatRegex)
       const trimmedNumber = phoneNumber.replace(/\s/g,'')
       const isPhoneValid = phoneValidator(trimmedNumber)
       
        if (isPhoneValid) {
          onPhoneChange(phoneNumber)
          setHasPhoneError(false)
        } else {
          setHasPhoneError(true)
        }
```

Additionally, if you end up using mentioned phone regex (`locale.phoneNumber.formatRegex`), you might want to update the placeholder since it doesn't seen to be compliance with it.

I understand that, as the result of QuoteCart API migration, this will be replaced but my concern is to get a simpler version out that's so simple that might not help us at all: i.e If we want to call users to understand why they didn't connect a payment but we don't try to make sure provided number is valid, we might end up with a bunch a numbers that will lead us to nowhere.

guilhermespopolin (commented):
👍🏻 Just added a couple of comments

dendoesit:
@styrelius it seems the BundleQuote varianta has changed and phoneNumber has dissappeared. The button has a check for offerData.person.phoneNumber. 😢

dendoesit:
@styrelius I forgot to update the graphql schema in this branch as well 😢 . should be fine now. 
@guilhermespopolin added the new placeholder and validation 👍

src/client/pages/OfferNew/types.ts
```diff
@@ -35,6 +35,7 @@ export type OfferPersonInfo = Pick<
 > & {
   householdSize: number
   address: Address | null
+  phoneNumber?: string | null
```
guilhermespopolin, line 38: We can pick `phoneNumber` from `BundledQuote`:

```
export type OfferPersonInfo = Pick<
  BundledQuote,
  'firstName' | 'lastName' | 'email' | 'ssn' | 'birthDate' | 'phoneNumber'
> & {
  householdSize: number
  address: Address | null
}
```

guilhermespopolin (approved):
It seems we're having issues with `editQuote` mutation which I don't think it's related with these code changes but I'd double check that. Also, just added small suggestion. Despite of that, code seems fine. Good job :)

styrelius (commented):
No something's wrong with the `editQuote` mutation in back-end so impossible to test this properly 🙃 But one thing I was thinking about is that this error message is not really right when the phone number just doesn't match the regex (i.e. being to long or having a "1" after "+47") 

Maybe it's good enough but still feels a bit strange. Also, all of the fields are required, but we only use that word for this field which is also slightly confusing. 

image

dendoesit:
@styrelius hmm, I can change the error message to 'That is not the right format' as it is coming from Lokalize and do a separate PR for that.

robinandeer (approved):
I got the edit mutation to work to update the phone number!

Didn't work to sign the quote but I'm not sure why...

PR #752, opened 2022-01-14
GRW-719 / Feature / Enable Quote Cart on Embark story level

robinandeer:
## What?

Define which Embark routes that use quote cart.
Use `EmbarkRoot` prop to determine if using quote cart.

## Why?

This way we can enable quote cart on a per story basis. Before we had to enable/disable it on a market level so this is more granular.

**Ticket(s): [GRW-719]**

[GRW-719]: https://hedvig.atlassian.net/browse/GRW-719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


guilhermespopolin (approved):

src/routes.tsx
```diff
@@ -304,6 +304,7 @@ export const routes: Route[] = [
                   return {
                     baseUrl: `/${locale}/new-member/home-accident-switcher`,
                     name: EmbarkStory.SwedenSwitcher,
+                    quoteCart: true,
```
mehdielket, line 307: Shouldn't it be `EmbarkStory.SwedenQuoteCartNeeder` a bit down this file that should have this prop?

robinandeer: Wow, good catch! You are correct!

guilhermespopolin (approved):
Well done

PR #753, opened 2022-01-14
GRW-719 / Feature / Always use existing forever page

robinandeer:
## What?

...independently if quote cart is enabled.
Add comment to switch to new component when migration is complete.

## Why?

We can use the existing setup for all users - when Embark starts up in a quote cart enabled story we will read the saved campaign code from the cookie storage and add to the quote cart.

When migration to the new API is complete we can remove the existing forever page component and use the new one.

**Ticket(s): [GRW-719]**

[GRW-719]: https://hedvig.atlassian.net/browse/GRW-719?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/pages/Forever/useRedeemCode.ts
```diff
@@ -55,6 +56,7 @@ export const useRedeemCode = () => {
 
       switch (result.data?.redeemCodeV2.__typename) {
         case 'SuccessfulRedeemResult':
+          CampaignCode.save(code)
```
guilhermespopolin, line 59: [Question]
How the old flow works? I mean, on the new one the campaign code is get from a cookie. But, as far as I understood, in the old one the code is present int the URL until [PreOnboardingScreen](https://github.com/HedvigInsurance/web-onboarding/blob/master/src/client/pages/Forever/components/Intro/PreOnboardingScreen.tsx#L56) is rendered. But it seems we do nothing with it when redirecting PreOnboardingScreen -> Oboarding page. 🤔

guilhermespopolin (approved):

src/client/pages/Forever/useRedeemCode.ts
robinandeer: Hmm let's see if I can explain it:

1. Add the campaign code to the session
2. Show `PreOnboardingScreen`
3. Show `Onboarding` page (code already added to session)

And I've just made sure to also save it to the cookie so we can pick it up in quote cart flows when Embark loads.

Makes sense?

guilhermespopolin: Absolutely. Thank you

PR #760, opened 2022-01-19
Remove redirect on offer page

robinandeer:
## What?

Remove redirect on (quote cart) offer page.

## Why?

I don't think this is strictly necessary since no one will end up on that page unless they go through a quote cart enabled Embark flow.

It also makes testing in staging easier since we don't need to turn on the quote cart API for that whole market.

**Ticket(s): []**


amandaflod (approved):
Looks good to me! 😄

PR #762, opened 2022-01-20
Remove phoneNumber from canInitiateSign

dendoesit:
## What?

Remove phoneNumber from canInitiateSign

## Why?

Checkout cannot be signed. 

image


robinandeer (approved):
Seems good :)

Just wondering if this is something we will need to re-add later? Or should it really be toggled by the feature flag?

guilhermespopolin:
> Seems good :)
> 
> Just wondering if this is something we will need to re-add later? Or should it really be toggled by the feature flag?

I guess we do need to add that again after Giraffe is released. So maybe we need to find a way to make sure we'd not forget?

dendoesit:
Created a ticket for this one in order not to forget. https://hedvig.atlassian.net/browse/GRW-727

PR #764, opened 2022-01-20
Icon tweaks and overview

styrelius:
## What?

- Icons we don't use are deleted
- An `IconsOverview` component which renders all components we have, and a Storybook story so this component actually is useful
- Updated `path` of `Arrow` icon , in order for it to correspond with the arrow icon in our [Figma Icon Library](https://www.figma.com/file/mNTQbhvXiNxUIuFBS7dNuW/Icon-Library?node-id=1%3A2) - the difference is that the lines are slightly thinner
- `'direction'` prop added to `Arrow` component (can be `'forward'` or `'backward'`)
- `ArrowBack` in `BackButton.tsx` is replaced with `Arrow`, using the `direction` prop to make it point backwards

## Why?

Icon component situation was pretty messy IMO. It was hard to understand which icons we actually have (and use) in this repo.

Replacing the `ArrowBack` with having a `direction` prop on the `Arrow` component was something that was discussed in a [previous PR](https://github.com/HedvigInsurance/web-onboarding/pull/754#discussion_r786285883).

## Screenshots / recordings 

**Icons overview in Storybook:**
image

**Landing page insurance product card with new `Arrow`:**

image

**Landing page insurance product card with current `Arrow`:**

image

<hr />


styrelius:
The arrow change is [approved by Zak](https://hedviginsurance.slack.com/archives/C01GAC9FG2E/p1642764348011300) 😇

amandaflod (approved):
Great idea to create the overview component! Checked out the BackButton as well and everything looks good to me 😎

robinandeer (approved):
Great work!

Do we need a React component for the icon summary or can we accomplish it all in a storybook story file? We could put it directly under: `/src/components/icons/icons.stories.tsx`.

styrelius:
> Do we need a React component for the icon summary or can we accomplish it all in a storybook story file? We could put it directly under: /src/components/icons/icons.stories.tsx.

@robinandeer Yes, we could probably just have the `IconOverview` component inside the `.stories` file 🙃

PR #765, opened 2022-01-20
Add check to avoid completing checkout twice

robinandeer:
## What?

See PR title.

## Why?

The request to generate access token was called twice which resulted in an error.
This caused the UI to not redirect to the connect payment page.

**Ticket(s): []**


dendoesit (approved):
👌

PR #767, opened 2022-01-21
GRW-727/PhoneNumber checkout fix

dendoesit:
## What?

Added logic to checking the phoneNumber in checkout

## Why?

Should only check for phoneNumber in Norway flow.

**Ticket(s): []**
https://hedvig.atlassian.net/browse/GRW-727

## Relates with
#762 

### [Review app]

https://web-onboardi-bugfix-grw-8m1bfg.herokuapp.com/no-en/new-member/debugger


src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -338,9 +343,15 @@ export const Checkout = ({
   const onPhoneChange = async (phoneNumber: string) => {
     const { phoneNumber: currentPhone } = offerData.person
     if (!phoneNumber || currentPhone === phoneNumber) return
+    setIsPhoneNumberUpdating(true)
     await editQuotes(quoteIds, { phoneNumber })
+    setIsPhoneNumberUpdating(false)
   }
 
+  const isPhoneNumberGoodToGo = isPhoneNumberRequired
+    ? !isPhoneNumberUpdating && typeof offerData.person.phoneNumber === 'string'
```
guilhermespopolin, line 352: 🤔 So, I guess we might have a problem here if we end up in the scenario where `offerData.person.phoneNumber` is an empty string: phone number is required but empty string is acceptable. Maybe we can fix it with:

```
  const isPhoneNumberGoodToGo = isPhoneNumberRequired
    ? !isPhoneNumberUpdating && offerData.person.phoneNumber
    : true
```

Phone number type is `BundledQuote.phoneNumber: string | null | undefined` and `null`, `undefined`, and `''` are falsy values, which would make it pass only in the scenario where `offerData.person.phoneNumber` has some content.

guilhermespopolin (commented):
Just added a comment.

src/client/pages/OfferNew/Checkout/index.tsx
dendoesit: hmm, we did that in order to avoid a typing error. if we do not check the type of offerData.person.phoneNumber then it can be undefined and the Sign component does not accept that. 
image

guilhermespopolin: I see. In order to `isPhoneNumberGootToGo` to have proper type (`boolean`) we'd need to convert `offerData.person.phoneNumber` to a boolean:

```
  const isPhoneNumberGoodToGo = isPhoneNumberRequired
    ? !isPhoneNumberUpdating && Boolean(offerData.person.phoneNumber)
    : true
```
or even
```
  const isPhoneNumberGoodToGo = isPhoneNumberRequired
    ? !isPhoneNumberUpdating && !!offerData.person.phoneNumber
    : true
```

Yeah, I was thinking on a weird scenario where, for some reason, we get '' from backend initially. Then it would be a problem. But maybe I'm expecting errors where it shouldn't have any.

guilhermespopolin (dismissed):

src/client/pages/OfferNew/Checkout/index.tsx
dendoesit: @robinandeer  what do you think ?

src/client/pages/OfferNew/Checkout/index.tsx
```diff
@@ -338,9 +343,15 @@ export const Checkout = ({
   const onPhoneChange = async (phoneNumber: string) => {
     const { phoneNumber: currentPhone } = offerData.person
     if (!phoneNumber || currentPhone === phoneNumber) return
+    setIsPhoneNumberUpdating(true)
     await editQuotes(quoteIds, { phoneNumber })
+    setIsPhoneNumberUpdating(false)
```
robinandeer, line 348: What happens if there's a network error here? Should we have a catch/finally block that always makes sure to `setIsPhoneNumberUpdating(false)`?

src/client/pages/OfferNew/Checkout/index.tsx
robinandeer: Hmm yeah I see your point @guilhermespopolin although we need to trust the API for some things. We can't be expected to perform all the checks in the client as well.

I believe it is enough to check if the field is a string. Otherwise, we should solve the root cause in the API (not allow an empty string to be submitted).

robinandeer (approved):
I was part of coming up with this approach but I think it looks good. The part about error handling I think we can add in a separate PR if we want to merge this already.

src/client/pages/OfferNew/Checkout/index.tsx
styrelius: I agree with you @guilhermespopolin. Isn't what we want to make sure that there is a phone number? I see your point @robinandeer that the root issue is in the API but since we're already performing checks here, as I think we should, it makes sense to me to check for all the possible falsy values (i.e. `undefined`, `null` and `''`). But I might be missing something :)

Personally I like the explicit type conversion `Boolean(offerData.person.phoneNumber)` more than the implicit type coercion `!!offerData.person.phoneNumber` since the latter isn't as immediately obvious IMO.

src/client/pages/OfferNew/Checkout/index.tsx
dendoesit: I guess if there is a network error you would not be able to add the ssn so you cannot sign at all. 😞 
Right ?

robinandeer: Yeah it's an edge case perhaps but it could be a temporary issue and then we want users to be able to retry and send the request again, right?

guilhermespopolin (approved):

PR #768, opened 2022-01-21
Fix trigger to complete checkout

robinandeer:
## What?

Use separate state to track if we started the complete checkout step.

Tested locally.

## Why?

It was causing issues to rely on the "sign UI state" since it's used for tracking other things.

**Ticket(s): []**


guilhermespopolin (approved):
I understood the changes but couldn't understand the issues caused by relying on `signUiState` 🤔 
Do you mind of explaining them?

PR #769, opened 2022-01-21
Landing page card Storybook story

styrelius:
## What?

- One dynamic `Card` Storybook story instead of several different ones in different states. This new story has controls for the text elements, whether or not the card is `disabled`, whether the badge should be there and for the width of the wrapper around the `Card`.

## Why?

- I started to tinker with this when working with the new `Arrow` icon, and felt that it made more sense to have these controls in order to easily get different versions of the cards.

## Screenshots / recordings 

**New dynamic `Card` story:**

https://user-images.githubusercontent.com/42962286/150564108-304b0500-be3f-4a1e-af50-774aac7fbcdf.mp4

**Current Storybook - three stories for three different states of the `Card`:*

2022-01-21-current-card-stories


dendoesit (approved):
👍

PR #771, opened 2022-01-24
fix(embark): remove form pre-filling feature

guilhermespopolin:
## What?

It removes _embark_ form pre-filling feature given the issues associated with it.

## Why?

To store embark flow answers into Session Storage and using them to pre-filling embark flow form fields has the tradeoff of enabling embark store having undesired values inside of it.

## Disclaimer

Be aware this solves the problem in only one out of the two scenarios where this issue occurs:

1. Restating a flow while using the same session: _embark_ flow enables having its store initialised with some data. We use this mechanism to implement _form pre-filling feature_ by keeping users' answers into session storage and using them to populate _embark_ form fields with previously provided data. With that, in case user restarts the flow by changing the URL with the first step of it, we would be able to use _Session Storage_ info on each passage. 
2. _embark_ changes its internal storage to keep answers provided by the user on each passage. However, by going back with _Go Back_ button, we don't undo _embark_ store data changes associated with passages that happened before going back. This might lead to inconsistencies inside _embark_ store.

**Ticket(s): [GRW-594](https://hedvig.atlassian.net/browse/GRW-594)**


robinandeer (commented):
Let's bring this up with Maggie and/or in daily to get people's input!

guilhermespopolin:
I'm closing this one as it was decided to not remove this feature since it fills an UX purpose of accidental page refresh/tab closing.

PR #772, opened 2022-01-24
GRW-744 / Fix / Avoid multiple calls to create access token

robinandeer:
## What?

See title.

## Why?

The check if we already called the function was accidentally reset in the `finally` block which is incorrect.

Also moved the check inside the complete checkout function to co-locate that logic.

Tested locally.

**Ticket(s): [GRW-744]**

[GRW-744]: https://hedvig.atlassian.net/browse/GRW-744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/pages/Offer/Checkout/index.tsx
+++ b/src/client/pages/Offer/Checkout/index.tsx
@@ -360,7 +360,7 @@ export const Checkout = ({
     if (checkoutStatus === CheckoutStatus.Signed) {
       completeCheckout()
     }
-  }, [checkoutStatus, completeCheckout, isCompletingCheckout])
+  }, [checkoutStatus, completeCheckout])
 
   const startSign = async () => {
     setSignUiState('STARTED')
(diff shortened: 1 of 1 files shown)


src/client/pages/Offer/Checkout/index.tsx
```diff
@@ -350,10 +353,11 @@ export const Checkout = ({
...
     variation,
+    isCompletingCheckout,
   ])
 
   useEffect(() => {
-    if (checkoutStatus === CheckoutStatus.Signed && !isCompletingCheckout) {
+    if (checkoutStatus === CheckoutStatus.Signed) {
       completeCheckout()
     }
   }, [checkoutStatus, completeCheckout, isCompletingCheckout])
```
guilhermespopolin, line 363: We probably can remove `isCompletingCheckout` from de dependency array, right?!

guilhermespopolin (approved):

mehdielket (dismissed):

src/client/pages/Offer/Checkout/index.tsx
robinandeer: Sloppy of me :P

guilhermespopolin (approved):

PR #776, opened 2022-01-25
Fix / Font-smoothing

amandaflod:
## What?

Add text smoothing to global css and remove duplicated GlobalCss in Page.tsx. 

## Why?

To make the font extra crisp and make it look like it does in Figma. 💅

**Ticket(s): []**

Before: 

Screenshot 2022-01-25 at 08 40 11

After: 

Screenshot 2022-01-25 at 08 43 00


robinandeer (commented):
Nice find on the duplicated global styles :)

After reading this article (https://usabilitypost.com/2012/11/05/stop-fixing-font-smoothing/) I'm not so sure about using the font-smoothing setting on a global level though. Seems like it's useful on white text + black background but reduces legibility when used for dark text on white background.

What do you others think?

amandaflod:
I see your point! I think this is maybe more of a design question and since we use smoothed (is that a word? 😅) fonts both in our Figma files and the market web we should try to be consistent? @robinandeer

robinandeer:
Agree that we should be consistent @amandaflod :)

I just don't wanna move in a direction that looks crisper but reduces readability.

robinandeer (approved):
The community doesn't seem to have a clear stance on this. Sub-pixel rendering is disabled on apple.com but not on google.com but disabled on material.io.

Here's an extended discussion: https://github.com/material-components/material-components-web/issues/248

I am fine moving in this direction!

PR #779, opened 2022-01-25
GRW-705 - Add proper server error support to in edit details modal

mehdielket:
## What?

Displays errors on specific fields in the Edit Details Modal when a server limit is hit.

* Shows an error badge next the field that is invalid

Note:
* The same treatment should be done to the checkout form (Invalid SSN for example)

## Why?

Currently we do not provide any feedback to users that their inputs do not much our business rules. Some examples:

* Living Space too large / small
* Too many co-insured
* Customer is too young or too old (depends on student some times)
* The year of constructions is too old

**Ticket(s): [GRW-705]**

https://user-images.githubusercontent.com/89857278/150981470-ee18d318-3779-4469-af0e-474be9863dca.mov

## Review App

https://web-onboardi-grw-705-di-ivyfqq.herokuapp.com/se/new-member/

[GRW-705]: https://hedvig.atlassian.net/browse/GRW-705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/pages/Offer/Introduction/DetailsModal/index.tsx
```diff
@@ -86,6 +92,57 @@ const LoadingDimmer = styled.div<{ visible: boolean }>`
...
+function getFormErrorsFromUnderwritterLimits(
+  limits: UnderwritingLimit[],
+  invalidFieldErrorMessage: string,
+) {
+  const invalidFields = limits.reduce((acc, limit) => {
+    const { code } = limit
+    switch (code as LimitCode) {
+      case LimitCode.INVALID_BIRTHDATE:
+      case LimitCode.UNDERAGE:
+      case LimitCode.STUDENT_OVERAGE:
```
robinandeer, line 108: Isn't student overage rather related to not being allowed to pick "isStudent"?

robinandeer (dismissed):
Great improvement!

(not yet tested)

robinandeer:
...and tested in review app!

src/client/pages/Offer/Introduction/DetailsModal/index.tsx
mehdielket: That's a good point. Maybe it could be both.
- Changing age when isStudent is true
- Turning on the Student Flag when the age is too high

One solution could be to show an error on both isStudent flag and birthdate. What do you think?

robinandeer: hmmm...the solution really is to display error messages 😅 

but that's next step. So I would say the most likely scenario is: "Changing age when isStudent is true". So we should show the error on "birth date" like now.

If you are already "too old" then I guess you can't edit "isStudent" anyways, no?

robinandeer (approved):

src/client/pages/Offer/Introduction/DetailsModal/index.tsx
mehdielket: I am not sure how this behaves. I will have to test. I can create tickets if needed

But I think I will merge this anyway as it is an improvement over what we have today.

PR #781, opened 2022-01-26
GRW-713/New-Inputs

dendoesit:
## What?

This will be new branch for input fields + errors .  We consider the old one with all the comments for inputs as approved. https://github.com/HedvigInsurance/web-onboarding/pull/732

Update to input fields.
Add generic errors to detailsModal and checkout 

## Why?

The new styling for input fields needs error messages instead of using the defaults from Formik. 

## **NIT**
 Worth noting that we discussed we will not focus on fixing all errors and issues on the current OfferNew or DetailsModal as we plan on moving to the new Offer soon. 
**Ticket(s): [https://hedvig.atlassian.net/browse/GRW-713]**

## Testing
Most changes can be noticed in the Checkout and inside the Details Modal both on the old and new debugger but every input on the website should be affected.

Before:
image

After: 
image

### [Review app]

https://web-onboardi-grw-713-fe-bs8s79.herokuapp.com/se-en/new-member/debugger


src/client/pages/OfferNew/Introduction/DetailsModal/utils.ts
```diff
@@ -221,7 +230,7 @@ const getSwedishSchema = (
                 label: getExtraBuilding(value),
                 value,
               })),
-              validation: Yup.string().required(),
+              validation: Yup.string(),
```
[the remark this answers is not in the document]
dendoesit: Required fields are not needed on select dropdowns as we currently always take the first option in a select.

mehdielket (dismissed):
Looks good to me. I tested a bit in the review app and I think it is ok as it is. 

I think we should be very careful when deploying this and test it thoroughly

robinandeer (commented):
Hmm there's a couple of empty files that we should maybe remove:

- src/client/components/inputs/types.ts
- src/client/components/inputs/utils.ts

src/client/pages/Forever/components/RedeemCode.tsx
```diff
@@ -185,7 +185,6 @@ export const RedeemCode: React.FC<RedeemCodeProps> = ({
                   }
                   name="code"
                   type="text"
-                  autoComplete="off"
```
robinandeer, line 188: Why do we change this?

src/client/pages/LoginApp/components/LoginForm.tsx
```diff
@@ -122,7 +122,6 @@ export const LoginForm: React.FC<LoginFormProps> = ({ onSubmit }) => {
                 id="ssn"
                 type="text"
                 inputMode="numeric"
-                pattern="[0-9]*"
```
robinandeer, line 125: Why change this?

src/client/pages/Offer/Introduction/DetailsModal/Details/components/DetailInput.tsx
```diff
@@ -47,7 +46,7 @@ export const DetailInput: React.FC<DetailInputProps &
       showErrorMessage={false}
       errors={getIn(formikProps.errors, formikName)}
       touched={getIn(formikProps.touched, formikName)}
-      autoComplete="off"
+      {...formikProps.getFieldProps(formikName)}
```
robinandeer, line 49: Does this mean we pass "autoComplete" to formik instead?

src/client/pages/LoginApp/components/LoginForm.tsx
dendoesit: We already have inputMode numeric on this which only allows characters between 0 and 9 so this is not needed.

src/client/pages/Forever/components/RedeemCode.tsx
dendoesit: It is hard for me to see a scenario where your browser already knows the redeemed code for this so it actually uses the autoComplete option.

src/client/pages/Offer/Introduction/DetailsModal/Details/components/DetailInput.tsx
dendoesit: I have removed autoComplete to simplify code as I couldn't see it bring any value. 
Furthermore Chrome ignores the autoComplete='off' from version 34. https://stackoverflow.com/questions/31612867/autocomplete-off-not-working-in-chrome

I am not sure if Formik handle autoComplete.

dendoesit:
@robinandeer to me it looks like the files have been removed. 
image

src/client/pages/Forever/components/RedeemCode.tsx
robinandeer: Hmm.. I think the browser might decide to suggest any number of irrelevant things so I would keep "autoComplete" turned off.

src/client/pages/LoginApp/components/LoginForm.tsx
robinandeer: Seems like Safari (desktop) doesn't support "inputMode" => so perhaps we need to keep it for that reason. But could be good to add that as a comment.

https://caniuse.com/input-inputmode

src/client/pages/Offer/Introduction/DetailsModal/Details/components/DetailInput.tsx
robinandeer: Hmm interesting. I guess it's not a standard. I'm okey not using it.

robinandeer:
> @robinandeer to me it looks like the files have been removed. image

Oh sorry about that. Must've read it wrong.

src/client/pages/LoginApp/components/LoginForm.tsx
dendoesit: oh Safari 😦 . great catch 👍 .

robinandeer (approved):
Amazing patience and work @dendoesit!

Things behave as expected now in the review app as far as I can tell. Sorry for delaying the code review - but with these huge PRs I just get overwhelmed 😪

Still feel like the spacing is a bit off (especially in the edit details modal) -- but we can discuss that will Amy/Zak later

PR #783, opened 2022-01-27
fix(GRW-732): update embark

guilhermespopolin:
## What?

The issue was solved in the _Embark_ [side](https://github.com/HedvigInsurance/embark/pull/759). So this just bumps `embark` version on `web-onboarding` to version `2.5.0`.

## Why?

Refer to the ticket for more info.

## Tickets
[GRW-732](https://hedvig.atlassian.net/browse/GRW-732)

### Before

before

### After

after


robinandeer (commented):
Nice!

But why does so much change in `yarn.lock`? Like why does "typescript" disappear??

guilhermespopolin:
> Nice!
> 
> But why does so much change in `yarn.lock`? Like why does "typescript" disappear??

😬 No idea. I've just updated _embark_ and ran `yarn install`. Let me double check!

robinandeer:
@guilhermespopolin hmmm I think what I usually do is just manually updating the version in `package.json` and then running `yarn install` :/

guilhermespopolin:
> @guilhermespopolin hmmm I think what I usually do is just manually updating the version in `package.json` and then running `yarn install` :/

Yep. That was my first attempt and resulting `yarn.lock` was also weird 🤷🏻

robinandeer:
> > @guilhermespopolin hmmm I think what I usually do is just manually updating the version in `package.json` and then running `yarn install` :/
> 
> Yep. That was my first attempt and resulting `yarn.lock` was also weird 🤷🏻

I guess if it works to delete `node_modules` and re-install then we can go ahead with this!

guilhermespopolin:
> > > @guilhermespopolin hmmm I think what I usually do is just manually updating the version in `package.json` and then running `yarn install` :/
> > 
> > 
> > Yep. That was my first attempt and resulting `yarn.lock` was also weird 🤷🏻
> 
> I guess if it works to delete `node_modules` and re-install then we can go ahead with this!
> 
> Anyone else have input @mehdielket @styrelius?

Also did that. I have no clue why is behaving like that 🤔

robinandeer:
@guilhermespopolin aaaahh! I will try locally and see if the same happens for me :)

robinandeer (approved):
@guilhermespopolin same happens for me -- let's go ahead and merge then

PR #784, opened 2022-01-27
feat(GRW-725): added cypress support

guilhermespopolin:
## What?

This is an attempt of adding _cypress_ support into `web-onboarding`.

## Why?

So we can run some e2e tests

## Further info

To run the tests, you'd need to go to `e2e` folder (`cd e2e`), change the node version to 16 with `nvm use` and then run `yarn cypress` command.
To be honest I don't like this approach: having `cypress` as another package inside `web-onboarding`. I've trying other ways but never with success. Using a more traditional `cypress` setup I've been getting the following when trying to run a test:

Screenshot 2022-01-27 at 15 15 59

It has been really hard to understand the source of the issue but it seems it's related with node version: we're using 14 while it seems only work with 16 (but that is also weird since they [state](https://docs.cypress.io/guides/getting-started/installing-cypress#Node-js) Node 14 is supported).  However, I've tried to only change the node version before running cypress but I end up getting the same error. A more traditional setup can be found on this [branch](https://github.com/HedvigInsurance/web-onboarding/tree/feature/GRW-725-e2e-traditional-approach).

By pulling it and running `yarn cypress` on can see in the _Settings_ tab that the node version being used is 14.17.6

Screenshot 2022-01-27 at 15 38 13

But even if one changes the node version (`nvm use 16`) before running `yarn cypress`, the same issue will occur 🤷🏻 

**Ticket(s): [GRW-725**](https://hedvig.atlassian.net/browse/GRW-725)


robinandeer (commented):
I understand that this is a compromise solution but since we don't intend to run these in CI (for now) I think it's a good compromise. So we can start using it already!

I just want to allow more node versions like:

```
"engines": {
    "node": "^16.x"
  },
```

🙏

robinandeer (approved):
Cool stuff -- let's start with this!

Tested locally :)

robinandeer (approved):
Woho!

PR #788, opened 2022-01-28
Add e2e tests for Quote Cart flows

robinandeer:
## What?

Add end-2-end tests for quote cart flows.
Add one end-2-end test for offer page => sign.

## Why?

So we don't have to click through all embark flows manually :P

**Ticket(s): []**

## On selecting elements

Cypress [officially recommends](https://docs.cypress.io/guides/references/best-practices#Selecting-Elements) using `[data-*]` selectors. However, I see a lot of value in trying to mimic the behavior of a real user when it comes to testing.

[Testing Library](https://testing-library.com/docs/queries/bytestid/#api) and [Kent](https://kentcdodds.com/blog/making-your-ui-tests-resilient-to-change#whats-with-the-data-testid-query) suggests only using data attributes when you can't reliably select elements any other way. I think this is a better approach than putting `data-testid` attributes all over the code base.

However, testing lib is mainly for integration tests. Kent also points out using `data-testid` is good for e2e tests specifically.

The ultimate goal should be to avoid brittle tests. Tests should survive refactors and only need to change when we update functionality on the site. If that means using data attribute selectors, we should not avoid it.

## Some caveats

1. Cypress doesn't play well with iFrames so it's hard to test the Adyen integration. I will leave it out for now and we can discuss the trade-offs later.
2. I'm not sure if we can unsign users programmatically which would be needed to run the offer page test in CI. Here is the setup needed for the test at least:

    ```javascript
    const UNSIGN_MEMBER = /* GraphQL */ `
      mutation UnsignMember($ssn: String!) {
        unsignMember(ssn: $ssn)
      }
    `
    before(() => {
      cy.request('POST', 'https://hedvig-hope-staging.herokuapp.com/api/graphql', {
        query: UNSIGN_MEMBER,
        variables: { ssn: SSN },
      })
    })
    ```


robinandeer:
Tested locally:

```
       Spec                                              Tests  Passing  Failing  Pending  Skipped
  ┌────────────────────────────────────────────────────────────────────────────────────────────────┐
  │ ✔  dk-flows.spec.ts                         01:40        3        3        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✔  dk-offer-page.spec.ts                    493ms        1        1        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✔  forever-page.spec.ts                     00:06        1        1        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✔  helpers.ts                                 5ms        -        -        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✔  no-flows.spec.ts                         00:38        1        1        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✔  se-needer-apartment-flow.spec.ts         00:37        1        1        -        -        - │
  ├────────────────────────────────────────────────────────────────────────────────────────────────┤
  │ ✔  se-needer-house-flow.ts                  00:01        1        1        -        -        - │
  └────────────────────────────────────────────────────────────────────────────────────────────────┘
    ✔  All specs passed!                        03:05        8        8        -        -        -

✨  Done in 265.24s.
```

mehdielket (approved):
After our conversation, I believe this is in good enough shape to be shipped as is. 
Amazing work and great leap forward for our testing setup 🚀

gustaveen:
Really amazing work with adding Cypress! 🎉

> On selecting elements

As you mentioned, I think it's good to use a `data-testid` attribute as selectors since the copy in buttons and such change over time. At Tink we used test attribute as selectors for Cypress tests and even for testing library. But I understand the point of on mimicking the user perspective in testing library.

Would be really nice if we could un-sign automatically in the future to have them running in CI and maybe even scheduled tests  🙌

PR #789, opened 2022-01-31
Checkout Pages footer (and enabling Storybook with Apollo)

styrelius:
## What?

**Footer component for our new checkout pages `/checkout/details/:id` and `/checkout/payment/:id`**

- `Footer` component to which you can pass text and a click handler function for the button there (since that's what's differs between the pages using the footer) 
  **Update:** The Footer now either takes a `buttonOnClick` prop or a `buttonLinkTo` prop. If there's a `buttonOnClick` then a regular `Button` will be rendered, if there's a `buttonLinkTo` a `LinkButton` will be used instead.
- Storybook Footer story.
- A `Price` query which uses the `quoteCart` but only gets the price data (price for bundle and individual quotes, and campaign info). This could with some small addition be reused for the **My plan** section on the `/checkout/details/:id` page if we want to.
- `PaymentInfo` component which uses the `Price` query to get its relevant data.
- `storybook-addon-apollo-client` package is added to enable using Storybook for components that handle their own data fetching instead of getting data as props. 
  This is implemented for the `Footer` story, since `PaymentInfo` which is rendered here does data fetching :)
- **Update:** `storyHelpers.ts` which exports a function to get the correct translations based on the `LocaleLabel` / locale `path` and `localeArgTypes` which can be used if you want a dropdown in your story with all available locales.
- ~Currency data is added to the `locales`. Maybe this should live in Giraffe but to me this seemed like the easiest, most straightforward way to be able to use the same currency symbol everywhere based on current locale~
  **Update:** `getFormattedPrice` function is added to `/utils` function - this implements the built-in internationalization API with the `Intl.NumberFormat` to return the price correctly formatted (i.e. a currency symbol correctly positioned and spaces/commas/periods in the right place). 
- Fallback locale is removed from `getLocaleParamFromPath` function in `useCurrentLocale.ts`, since we really shouldn't just use a locale that's not correct. Instead I'm now throwing an error. 
  Question though: Is this risky as long as we don't have like an Error Boundary for the entire application or something?

## Why?

- This is of course part of Connect Payment at Sign / redesign of the checkout :)
- Why it made sense to spend time at enabling mocking requests for Storybook stories is that we probably will have more components getting their own data going forward, making them truly encapsulated and easily moved around.

## How to test

Use the review app links below to have a look at the `Footer` component when it's added to the Checkout Details and Checkout Payment pages. The price and price information will change based on the `locale` and `id` (quote cart ID) URL params.

I also recommend checking out this branch and testing the Storybook implementation locally :)

**Ticket(s): [GRW-649]**

## Screenshots / recordings 

image

image

https://user-images.githubusercontent.com/42962286/152531934-468917d9-db61-42cd-9759-7fc2fda51ea0.mov

<hr />

### [Review app /checkout/payment](https://web-onboardi-grw-649-ch-6svaeh.herokuapp.com/no-en/new-member/checkout/payment/ecbe985f-e93d-45b5-afe9-5e5104eb7372)

_Please note that the quote cart ID in these URL's is Norwegian, so if you change the locale path to be for another market the currency will still be Norwegian krone._ 

[GRW-649]: https://hedvig.atlassian.net/browse/GRW-649?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/graphql/fragments/CampaignDataFragment.graphql
```diff
@@ -15,6 +15,7 @@ fragment CampaignData on Campaign {
     }
   }
   code
+  # TODO: Remove owner, since it's deprecated
```
mehdielket, line 18: [Code Scout] : Maybe it would be nice to go the extra mile here and actually remove it?

src/client/l10n/currency.ts
```diff
@@ -0,0 +1,23 @@
...
+  },
+  NO: {
+    currencySymbol: 'kr',
+    currencyCode: 'NOK',
+  },
+  DK: {
+    currencySymbol: 'kr.',
+    currencyCode: 'DKK',
+  },
+}
```
mehdielket, line 23: On displaying currency properly: I had looked into this before and there is a good browser API that achieves this so we don't have to make it ourselves. The browser API also handles the position of the currency in the price string (eg `3kr` vs `$3`).

https://hedvig.atlassian.net/browse/GRW-559

mehdielket (changes requested):
Added some comments. Also the design seems to need some padding on mobile

Screenshot 2022-02-01 at 14 20 13

src/client/graphql/fragments/CampaignDataFragment.graphql
styrelius: Maybe. Just thought that this change might require work in other places as well, since I assume we use this data somewhere. And didn't want to make this PR even bigger.

src/client/l10n/currency.ts
styrelius: Ah cool, I hadn't seen that Jira issue. But also saw that @robinandeer commented about having the logic in Giraffe. Did you discuss this further? 🤔 Also, did you do any further investigations into fallback solutions and whether or not we need one of those? Sounds super neat if it would just work with that browser API 👌

src/client/components/checkout/Footer.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+import { PriceDocument, Locale } from 'data/graphql'
+import { localePathPattern } from 'l10n/localePathPattern'
+import { Footer, Props as FooterProps } from './Footer'
+
+type StoryProps = {
+  localePath: LocaleLabel
+} & FooterProps
+
+const storyMeta: Meta<StoryProps> = {
+  title: 'Components/Checkout/Footer/Footer',
```
gustaveen, line 19: Looking at the structure of components, shouldn't we keep all components that are specific to Checkout under something like `pages/Checkout/..`? To me it makes more sense, since `components/` usually consists of generic components that are not tied to specific views or pages.

src/client/l10n/currency.ts
robinandeer: Hmm, yeah I think the giraffe-approach is too far out to consider now

The browser API usually only supports English locale 🙄  but we could investigate if adding other locales is worth it. But yeah, I would do that as a separate PR/project.

src/client/components/checkout/Footer.stories.tsx
```diff
@@ -0,0 +1,142 @@
...
+  component: Footer,
+  args: {
+    buttonText: 'Continue to payment',
+    buttonOnClick: () => {
+      console.log('👉 Click!')
+    },
+    localePath: 'no-en',
+  },
+  parameters: {
```
gustaveen, line 28: To make the controls a bit easier to use you could predefine a dropdown with locales :) 
```suggestion
  argTypes: {
    localePath: {
      name: 'locale',
      options: ['se', 'se-en', 'no', 'no-en', 'dk', 'dk-en'],
      control: { type: 'select' },
    },
  },
  parameters: {
```

gustaveen (commented):
Awesome work with adding `storybook-addon-apollo-client` Ellen! 🎉

src/client/graphql/fragments/CampaignDataFragment.graphql
guilhermespopolin: I took the opportunity and I've created a [ticket](https://hedvig.atlassian.net/browse/GRW-768) for it so we can work on it outside this PR.

guilhermespopolin (commented):
Good stuff 🎉

src/client/components/checkout/Footer.stories.tsx
styrelius: Hmm, yeah maybe. These components that I put under "Components" are there since that's where they live in the folder structure. And the reason for that is that we have two checkout pages that share some components (like this one for example). So it can't really be placed in one page folder 🤔 

But maybe our Storybook folder structure doesn't have to mirror the actual folder structure? Or we could have a `Checkout` folder under `pages` with a `Details`, `Payment` and `shared` folder? What do you think?

src/client/components/checkout/Footer.stories.tsx
styrelius: OMG thank you, that's what I really wanted but didn't understand how to do that (and just gave up without googling lol) 👌 🙏

src/client/l10n/currency.ts
styrelius: But doesn't it support all (or most locales)? Or seems like it [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat) at least? Makes a lot of sense to me to use this, since as @mehdielket pointed out that also solves position of currency symbol, plus if there should be a comma or dot or whitespace as delimiter. Or am I just missing something from skimming the MDN page?

src/client/components/checkout/Footer.stories.tsx
styrelius: OK so I fixed it in Storybook and put the `Footer` directly below `Checkout`. After some consideration I think the repo folder structure should also be changes to what I suggested above. But I will do that in another PR.

src/client/l10n/currency.ts
robinandeer: Hmm maybe it does -- that would be amazing :)

guilhermespopolin (dismissed):

src/client/l10n/currency.ts
styrelius: OK so I've made a try to use the JS Internationalization API for this and it seems to work. Only annoying thing is that the `Intl.NumberFormat()` constructor of course uses a format for the locale arguments than we don't use anywhere else. I solved this in the most "dumb" but easy way and added a `currencyLocale` to the `locales` objects. Felt that all "smart" solutions to use other locale properties that we already have would most likely be shooting ourselves in the foot.

src/client/utils/getFormattedPrice.ts
```diff
@@ -0,0 +1,26 @@
+import { LocaleData } from 'src/client/l10n/locales'
```
robinandeer, line 1: Fix import :)

src/client/utils/getFormattedPrice.ts
```diff
@@ -0,0 +1,26 @@
...
+
+type FormatCurrencyParams = {
+  locale: LocaleData['currencyLocale']
+  amount: string
```
robinandeer, line 5: hmm, why string? seems reasonable to expect it always as a number, no?

src/client/pages/CheckoutPayment/CheckoutPayment.tsx
```diff
@@ -1,6 +1,19 @@
...
-  return <CheckoutPageWrapper>{}</CheckoutPageWrapper>
+  const textKeys = useTextKeys()
+
+  return (
+    <CheckoutPageWrapper>
+      <Footer
+        buttonText={textKeys.CHECKOUT_FOOTER_COMPLETE_PURCHASE()}
+        buttonOnClick={() => {
+          console.log('click')
+        }}
```
robinandeer, line 15: I'm guessing this is just a placeholder for later 😅

src/client/pages/CheckoutDetails/CheckoutDetails.tsx
```diff
@@ -1,6 +1,23 @@
...
+import { useTextKeys } from 'utils/textKeys'
+import { useCurrentLocale } from 'l10n/useCurrentLocale'
 import { CheckoutPageWrapper } from 'components/checkout/CheckoutPageWrapper'
+import { Footer } from 'components/checkout/Footer'
 
 export const CheckoutDetails = () => {
-  return <CheckoutPageWrapper>{}</CheckoutPageWrapper>
+  const textKeys = useTextKeys()
+  const { path: localePath } = useCurrentLocale()
+  const { id: quoteCartId } = useParams<{ id: string }>()
```
robinandeer, line 11: Didn't we create a custom hook for getting quote cart ID from URL?

src/client/l10n/locales.ts
```diff
@@ -39,6 +39,7 @@ export type LocaleData = {
     formatRegex: RegExp
     backendFormatExample: string
   }
+  currencyLocale: 'en-SE' | 'sv-SE' | 'en-NO' | 'nb-NO' | 'en-DK' | 'da-DK'
```
robinandeer, line 42: isn't this really the standard/ISO locale? Maybe we should name it that.

And also use it as html-lang e.g.

src/client/components/checkout/PaymentInfo.tsx
```diff
@@ -0,0 +1,76 @@
...
+  const { data, loading: isLoading, error } = usePriceQuery({
+    variables: {
+      id: quoteCartId,
+      locale: isoLocale,
+    },
+  })
+  const quoteBundle = data?.quoteCart.bundle
+
+  if (error || !quoteBundle) {
+    return <div></div>
```
robinandeer, line 47: Do we really want to return this or e.g. `null`?

src/client/components/checkout/Footer.tsx
```diff
@@ -0,0 +1,68 @@
...
+  left: 0;
+  display: flex;
+  justify-content: center;
+  background-color: ${white};
+  box-shadow: 0px -4px 8px rgba(0, 0, 0, 0.05),
+    0px -8px 16px rgba(0, 0, 0, 0.05);
+`
+const InnerWrapper = styled.div`
+  width: 100%;
+  max-width: 628px;
```
robinandeer, line 34: nit: do we have a constant for this?

robinandeer (commented):
Okey lots of good stuff here!

I found one incorrect import so need to fix that and then I'll approve!

src/client/components/checkout/Footer.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const buttonProps: FooterButtonProps = {
+    background: purple500,
+    foreground: gray900,
+  }
+
+  return (
+    <Wrapper>
+      <InnerWrapper>
+        <PaymentInfo />
+        {buttonText && (
```
guilhermespopolin, line 51: [Discussion]

What do you think about an API like this?
```
type Props =  { actionButton: JSX.Element }

export const Footer = ({ actionButton }: Props) => {
   return (
     <Wrapper>
       <InnerWrapper>
         <PaymentInfo />
         {actionButton}
       </InnerWrapper>
     </Wrapper>
   )
}
```
I think it's easier to maintain since are just compose `Footer` with something that needs to be rendered: `actionButton`. With the current used approach, If you end needing to add more stuff specific to be button scenario, for example, you might need to accept to change `Footer` more button related props just so on pass them down, which end up making `Footer` more complex to work with IMHO.

guilhermespopolin (commented):
Nice work with `Intl` :)
Just added one comment

src/client/components/checkout/Footer.tsx
styrelius: No, not yet. Have thought about and will add that in another PR :)

src/client/components/checkout/PaymentInfo.tsx
styrelius: Oh, yeah, that's just to always keep the button on the right side of the footer, even if there's an error. But also doesn't make sense to render som error state since if this isn't working I'm pretty sure we'll have issues in other places as well and it makes more sense to render e.g. the error modal from the `CheckoutPageWrapper` for example. 
But I know this doesn't look good so note sure what's the best approach 🤔

src/client/pages/CheckoutDetails/CheckoutDetails.tsx
styrelius: I haven't seen anything like that, I think maybe we just talked about creating that? Or I missed it 🙃

src/client/pages/CheckoutPayment/CheckoutPayment.tsx
styrelius: No it's what should happen 🤪 (Yes, it's a placeholder)

src/client/utils/getFormattedPrice.ts
styrelius: Because we get a string from Giraffe, and I'd rather do the conversion to number here than everywhere we use this function.

src/client/utils/getFormattedPrice.ts
styrelius: Oops, thank you!

src/client/l10n/locales.ts
styrelius: Yes but we already have another `IsoLocale`, which is the `Locale` type from the GraphQL schema, which uses `_` instead of `-`. So I made the decision to name this based on what we use it for. Maybe we should go over the naming of lots of these properties in a separate PR? 🤔

styrelius: But if we can use it for html-lang then it's awesome but I thought you only wanted to provide the language tag for that and not include country/region?

src/client/components/checkout/Footer.tsx
styrelius: Hm, yes that's a good point. Probably won't do that remake now though, since I'm not sure it would be worth the effort. Or what do you think?

guilhermespopolin: Yeah, you're right.

src/client/utils/getFormattedPrice.ts
styrelius: But maybe we should accept a `number` as well? I.e. `string | number`? 🤔

mehdielket (approved):

src/client/utils/getFormattedPrice.ts
robinandeer: Ahaaaa, just makes me a little worried if we let the graphql API direct how our implementation looks like

In my world, it should be the other way around :P (our code should make sense even without the API context)

But this is a bit nit-picky so perhaps a progressive approach is to allow both string + number input.

src/client/pages/CheckoutDetails/CheckoutDetails.tsx
robinandeer: Hmm nope, you are right. When we do this a third time we can extract it into a hook then haha

src/client/l10n/locales.ts
robinandeer: aha, that might be the case indeed, okey -- let's keep it like this :)

src/client/components/checkout/PaymentInfo.tsx
robinandeer: Hmm, could be a case where adding a "why" comment is warranted -- if there's no other simple way around it.

robinandeer (approved):
I don't see any major hurdles left :D

PR #790, opened 2022-01-31
GRW-760 / Feature / Add origin Embark story to tracking

robinandeer:
## What?

Add new EmbarkStory interface.
Add current insurer to tracking.

## Why?

Highest prio analytics improvements from Sonny.

Long term I want to decouple publishing an analytics event from gathering data that should go along with it. For example in this case, we just read some data from the session directly in the tracking function instead of worrying about that in the React component render function.

**Ticket(s): [GRW-570]**

[GRW-570]: https://hedvig.atlassian.net/browse/GRW-570?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


mehdielket (approved):
Nicely done

guilhermespopolin (approved):
👍🏻

PR #793, opened 2022-02-01
Chore/debugger labels

dendoesit:
## What?

See title

## Why?

The new debugger is black and labels are black . 

image


styrelius (approved):
👍

PR #795, opened 2022-02-03
GRW-705 show invalid SSN error in checkout

mehdielket:
## Why?

**Ticket(s): []**

### DIFF
--- a/src/client/pages/Offer/Checkout/index.tsx
+++ b/src/client/pages/Offer/Checkout/index.tsx
@@ -1,5 +1,5 @@
 import React, { useEffect, useRef, useState, useCallback } from 'react'
-import { useFormik } from 'formik'
+import { useFormik, FormikHelpers } from 'formik'
 import { css } from '@emotion/core'
 import styled from '@emotion/styled'
 import { colorsV3 } from '@hedviginsurance/brand'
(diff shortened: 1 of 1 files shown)


src/client/pages/Offer/Checkout/index.tsx
```diff
@@ -185,6 +185,14 @@ const isManualReviewRequired = (errors: GraphQLError[]) => {
   return manualReviewRequiredError !== undefined
 }
 
+const isSsnInvalid = (errors: GraphQLError[]) => {
+  const invalidSsnError = errors.find((error) => {
+    return error?.extensions?.body?.errorMessage === 'Invalid SSN'
```
robinandeer, line 190: Hmmm this is never shown to users, right?

If it's more like an error code I would store it as a const first and reference that. Don't like hard coded strings 😅

robinandeer (approved):
Looks good, just one comment

src/client/pages/Offer/Checkout/index.tsx
mehdielket: Yes, it's never shown to users. I am planning to get in touch with the backend teams as this should be rather a code instead. We are relying on an error message because the errorCode is empty.

robinandeer (approved):

guilhermespopolin (approved):

robinandeer (approved):
Re-approve

PR #800, opened 2022-02-03
Fix / Error Modal margin

amandaflod:
## What?

Added a fix so that there will always be a margin on the Error Modal when the screen gets smaller than the modal width.

## Why?

To make it look like it should on mobile :D 

**Ticket(s): []**

Screenshot 2022-02-03 at 17 06 16


robinandeer (approved):
I think @styrelius raised a good point that this logic would live in the place that implements this `ErrorModal`.

However, since we already define max-width and this won't affect other elements on the page I guess it's a good solution for this use case.

src/client/components/ErrorModal.tsx
```diff
@@ -38,8 +38,10 @@ export const ErrorText = styled.div`
 `
 
 const CustomModal = styled(Modal)`
-  overflow-x: auto;
-  max-width: 36rem;
+  && {
```
guilhermespopolin, line 41: Just curious: what `&&` means?
Wouldn't just the following be enough? 🤔 

```
const CustomModal = styled(Modal)`
    overflow-x: auto;
    max-width: min(36rem, 100% - 2rem);
`
```

guilhermespopolin (approved):

src/client/components/ErrorModal.tsx
guilhermespopolin: Got it, thanks.

PR #802, opened 2022-02-07
Fix path locale for privacy policy link

robinandeer:
## What?

Use path locale instead of ISO locale.

## Why?

This will fix the URL path issue but the link is still invalid locale languages.

I suggest we add redirects to handle this in storyblok. What do you think of that, @gustaveen?

- /se/privacy-policy => /se/personuppgifter
- /dk/privacy-policy => /dk/personoplysninger
- /no/privacy-policy => /no/personvernpolicy

**Ticket(s): [GRW-774]**

[GRW-774]: https://hedvig.atlassian.net/browse/GRW-774?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gustaveen (approved):
Nice! 🙌 Is the problem that we don't get the path from backend most of the time so we fallback to the hardcoded one? I think it would be fine to create those redirects. But it would of course be good if we got the correct paths from the backend @robinandeer

robinandeer:
@gustaveen I'm not sure why we don't get it and have to use the fallback.

Perhaps Maggie can help us look into that :)

robinandeer:
Screenshot 2022-02-08 at 09 25 17

Added the redirect in Storyblok

PR #806, opened 2022-02-07
Fix / Checkout pages Footer styling

styrelius:
## What?

Refactoring of `Footer` and `PaymentInfo` components so that the footer button will always be rendered in the same place, regardless of whether the `PaymentInfo` component returns content, or if it returns `null`, in case of error.

There shouldn't be any visual changes.

## Why?

Since in case of error in `PaymentInfo` there will also be errors somewhere else it feels like overkill to return some sort of error component here. Instead I just don't want to render anything. But since we might want to use the error modal with a semi-transparent backdrop in case of error (the same way it looks in case of signing error, although with different error modal content) I don't want to mess up the layout of the footer.

### DIFF
--- a/src/client/components/checkout/Footer.tsx
+++ b/src/client/components/checkout/Footer.tsx
@@ -18,13 +18,14 @@ type FooterButtonProps = {
 
 const Wrapper = styled.div`
   width: 100vw;
-  height: 5rem;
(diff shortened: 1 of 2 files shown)


robinandeer (approved):
Oh nice! I like this approach 👌

PR #807, opened 2022-02-07
Add translation to birth date error message

robinandeer:
## What?

Screenshot 2022-02-07 at 15 23 17

## Why?

I added a generic "incorrect format" error.

We should be more specific but I still think this is an improvement that we can go live with.

**Ticket(s): [GRW-777]**


dendoesit (approved):
I agree we should have better error messages but these are a good start . 👍

PR #808, opened 2022-02-07
Fix / Checkout pages folder structure

styrelius:
## What?

Changed folder structure for new checkout pages.
Instead of having a `/checkout` folder below `/components` for the components shared between checkout pages we'd have all the checkout stuff in a folder called `/Checkout` in `/pages`.

New folder structure: 
```
/pages
|-- Checkout
      |-- CheckoutDetails
            |-- components
            |-- CheckoutDetails.tsx   //instead of an index.tsx
      |-- CheckoutPayment
            |-- components
            |-- CheckoutPayment.tsx   //instead of an index.tsx
      |-- shared    //all components shared between the checkout pages, e.g. Footer.tsx
```

## Why?

The idea is to make the components related to the checkout pages easier to find and let the `src/client/components` folder be for components that are meant to be used in several places.

**Ticket(s): [GRW-780]**

<hr />

### [Review app](https://web-onboardi-grw-780-fi-3ax57u.herokuapp.com/no/new-member/checkout/details/ecbe985f-e93d-45b5-afe9-5e5104eb7372)
_Nothing new here, just to see that it still works_ 🙃 

[GRW-780]: https://hedvig.atlassian.net/browse/GRW-780?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


guilhermespopolin (approved):
Good idea 💡 
I would just make one change tho: Keep `PageSection` component as a sibling of `CheckoutDetails` inside CheckoutDetails folder and remove `components` folder. Personally, I don't thing there is much gain of compartmentalising at this level deep. But that's just my humble opinion :)

PR #809, opened 2022-02-07
GRW-123 / Fix / Reset authorization header before `startCheckout`

robinandeer:
## What?

Before sending `startCheckout` mutation I make sure that the authorization header is removed.

## Why?

We accidentally sent the request with an authorization header using the `createSessionV2` token. This caused the backend to use it (the member ID) which caused the request to fail.

In Quote Cart we don't have a token until the cart has been signed (after checkout). However, in Embark, a session (old) is created before running ANY graphql query/mutation and the resulting token gets added to the client so it's hard to avoid from the start.

We probably should set up the authorization header according to the official guide: https://www.apollographql.com/docs/react/networking/authentication/#header. However, this is beyond the scope of this PR.

**Ticket(s): [GRW-778]**

## Screenshots / recordings 

https://user-images.githubusercontent.com/1220232/152840028-73fc375f-e38f-4327-8480-b62dfac1a918.mov

[GRW-778]: https://hedvig.atlassian.net/browse/GRW-778?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


guilhermespopolin (commented):
🤔 Wouldn't be the case of using a different graphql resolver [here](https://github.com/HedvigInsurance/web-onboarding/blob/master/src/client/pages/Embark/index.tsx#L448)? Because If I understood properly, we're getting an access token and adding it to the _GraphQL_ client inside this `setupSession` function, which is being called inside provided Embark graphql query/mutation resolvers..

robinandeer:
> 🤔 Wouldn't be the case of using a different graphql resolver [here](https://github.com/HedvigInsurance/web-onboarding/blob/master/src/client/pages/Embark/index.tsx?rgh-link-date=2022-02-08T09%3A33%3A23Z#L448)? Because If I understood properly, we're getting an access token and adding it to the _GraphQL_ client inside this `setupSession` function, which is being called inside provided Embark graphql query/mutation resolvers..

Hmmm, yes that would be a smart solution. But I try not to be smart here 😅  there are other cases where we setup the session (forever page e.g.) so we might add the token to the apollo client anyways.

That's why I still think this approach is what we should go with. And after the migration, we simply get rid of this logic!

mehdielket (approved):
🚀

PR #810, opened 2022-02-09
GRW-775 Replace `editQuote` with `createquoteBundle` for cancellation/start date updates

mehdielket:
## What?

Use `createQuoteBundle` for editing start date with Quote Cart.
Refactor logic a bit to reuse the same mutation handler, e.g. remove "offerData" usage.

## Why?

The edit endpoint in Giraffe doesn't work according to spec. Therefore we want to implement it this way until it is ready to be used in prod.

**Ticket(s): [GRW-775]**

[GRW-775]: https://hedvig.atlassian.net/browse/GRW-775?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/components/utils.ts
```diff
@@ -9,7 +9,6 @@ export const Spinner = styled.div`
   display: inline-block;
   width: 1em;
   height: 1em;
-  flex: 1;
```
robinandeer, line 12: This fixes the spinner, makes it look round.

src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
```diff
@@ -151,39 +163,39 @@ const getDateFormat = match<MarketLabel, string>([
 ])
 
 const DateForm: React.FC<{
-  quoteCartId: string
-  quote: OfferQuote
-  offerData: OfferData
-  isSingleStartDateBundle?: boolean
+  startDate: any
```
guilhermespopolin, line 166: Shouldn't be `Date | null`?

src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
```diff
@@ -76,59 +43,46 @@ const getLabelContent = (textKeys: TextKeyMap, quote: OfferQuote) => {
...
-          },
-        },
-      })
-    } catch (e) {
-      setShowError(true)
-    } finally {
-      setIsLoading(false)
-    }
-  }
+  const isDisabled = loadingQuoteIds.length > 0
```
guilhermespopolin, line 58: Now all _cancel my insurance_ checkbox will be disabled when there is one "cancellation" in progress. Not sure if we want that 🤔

guilhermespopolin (commented):
👍🏻 
Just added a couple of comments

src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
robinandeer: We do since we use the "createQuoteBundle" mutation. Otherwise you would possibly end up in unintended states.

src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
robinandeer: This is because it's a scalar in the schema so it's typed as `any`. Not a great solution but I think a good enough trade-off for now. What do you think?

robinandeer (approved):
🎉

guilhermespopolin (approved):

PR #811, opened 2022-02-09
fix(CancellationOptions): align styles between versions

guilhermespopolin:
## What?

Consolidates the styles used by the two versions of `CancellationOptions` and fixes a bug related with spinner used by this component (check image in the following section)

## Why?

Screenshot 2022-02-09 at 11 34 21

As a fix, I'm overriding `flex` CSS property while using `Spinner` inside `CancellationOptions` (link here). But maybe a more robust solution would be refactor `Spinner` by removing it's [`flex` property](https://github.com/HedvigInsurance/web-onboarding/blob/master/src/client/components/utils.ts#L12) since it might be complicating things. But `Spinner` is being used on several files which would require more effort to check that nothing was broken:

Screenshot 2022-02-09 at 12 06 19

So, for now I've decided leave as it is. Let me know your opinions about it.

## Relates to

CPS CancellationOptions redesign - #743 

**Ticket(s)**
_NA_

## Screenshots / recordings 

Screenshot 2022-02-09 at 12 16 55

Screenshot 2022-02-09 at 12 17 07

## How to test

Access the review app linked below and complete the switcher flow in _Embark_. In case you don't have an account with a different insurance provider, you can just skip the part of the flow which is about to collect info about your current insurance provider by selecting any insurance provider at the beginning of the flow and then clicking on [Skip] button:

Make sure to select a provider that we're able to cancel current insurance. E.g: In Sweden, Trygghansa

### DIFF
--- a/src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
+++ b/src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
@@ -24,13 +24,14 @@ const HandleSwitchingLabel = styled.label<{ isClickable: boolean }>`
 `
 
 const StyledSpinner = styled(Spinner)`
+  flex: none;
   height: 1.25rem;
   width: 1.25rem;
-  margin-right: 0.5rem;
+  margin-right: 1rem;
 `
 
 const StyledSwitch = styled(Switch)`
-  margin-right: 0.5rem;
+  margin-right: 1rem;
 `
 
 type CancellationOptionsProps = {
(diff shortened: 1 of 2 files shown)


robinandeer (approved):
Looks good! 🦖

dendoesit (approved):
👍

src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
```diff
@@ -24,13 +24,14 @@ const HandleSwitchingLabel = styled.label<{ isClickable: boolean }>`
 `
 
 const StyledSpinner = styled(Spinner)`
+  flex: none;
   height: 1.25rem;
   width: 1.25rem;
-  margin-right: 0.5rem;
+  margin-right: 1rem;
```
dendoesit, line 30: asdasd

PR #812, opened 2022-02-09
fix(GRW-642): disable start date when current insurance automatic cancellation is requested

guilhermespopolin:
## What?

Disable `StartDate` field in case users request the cancellation of their old insurance

## Why?

In such situations the start date should be right after the end date of the old insurance, which means there's no need to let them pick a start date.

**[Ticket](https://hedvig.atlassian.net/browse/GRW-642)**

## Screenshots / recordings 

Screenshot 2022-02-10 at 09 18 59
Screenshot 2022-02-28 at 11 47 32

## How to test it

Access the review app linked below and complete the switcher flow in _Embark_:

1. Select an insurance provider that we're able to cancel insurances from, E.g: Trygghasna. at the beginning of the flow;
2. Click on [Skip] button and finish the flow;
3. You should not be able to interact with Date picker fields for current insurances that should be cancelled by Hedvig while signing a new insurance. Please note that you can control which insurance should be cancelled by interacting with its corresponding _Cancel Insurance_ checkbox.


robinandeer (changes requested):
Something is not working here 🤔 

Now I can't change the date even if I'm not a switcher

Screenshot 2022-02-11 at 15 33 39

src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
```diff
@@ -213,6 +220,7 @@ const DateForm: React.FC<{
     <RowButtonWrapper isSplit={isSplit}>
       {isSplit && <StartDateRowLabel>{quoteDisplayName}</StartDateRowLabel>}
       <RowButton
+        disabled={currentInsurer && !startDate}
```
robinandeer, line 223: Why is `Boolean` not needed here?

robinandeer (approved):
This works when I test it now -- yaaay!

src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
guilhermespopolin: Good question. I'm no sure 😅 
Either way, I'm gonna add it.

robinandeer (approved):

PR #814, opened 2022-02-09
Fix / Message and email address for manual review signing errors

styrelius:
## What?

Bringing back the old special error message when we have a `'MANUAL_REVIEW_REQUIRED'` error (but upgraded to contain a `mailto:` link).

Also took the opportunity to add some nice things to the `SignFailModal` Storybook story :)

## Why?

Apparently IEX wants users who requires a manual review to sign to email a special address, which we previously displayed but removed after having decided that we should just have one generic signing error modal.

<hr />

**Ticket(s): [GRW-782]**

## Screenshots / recordings from review app
_The recordings first show what happens with an SSN that requires manual review, and then what happens with an SSN that is already signed_

image

**Current checkout:**

https://user-images.githubusercontent.com/42962286/153594537-6f9731aa-789f-4ef0-8161-460e9a6e4fdb.mov

<br />

**Quote cart checkout:**

https://user-images.githubusercontent.com/42962286/153594576-ea1119f1-3be2-454e-9a9e-2cb814f61308.mov

### [Review app](https://web-onboardi-grw-782-fi-in3hgm.herokuapp.com/no-en/new-member/debugger)

[GRW-782]: https://hedvig.atlassian.net/browse/GRW-782?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/pages/Offer/Checkout/index.tsx
```diff
@@ -422,9 +423,11 @@ export const Checkout = ({
         return
       }
     } catch (error) {
-      if (
-        !isManualReviewRequired((error.graphQLErrors || []) as GraphQLError[])
```
styrelius, line 426: @mehdielket I think this little `!` is the thing that made the error modal pop up when we didn't expect it to? But maybe you already figured that out 😊

mehdielket (commented):
Looks fine to me. What worries me is the infinite loader on the sign button even if the request has actually failed...

styrelius:
> Looks fine to me. What worries me is the infinite loader on the sign button even if the request has actually failed...

Hm, yes, that's not perfect. I think it might look the same in prod since what I think causes it is a `return` statement that I just left there from before. Anyway, removed it now and it should do the trick :)

dendoesit (dismissed):
lgtm 👍

amandaflod (approved):
Looks good to me! 😊

src/client/pages/Offer/Checkout/index.tsx
```diff
@@ -265,6 +265,7 @@ export const Checkout = ({
     checkoutStatusData?.quoteCart.checkout || {}
 
   const [isShowingFailModal, setIsShowingFailModal] = useState(false)
+  const [isManualReviewRequired, setIsManualReviewRequired] = useState(false)
```
robinandeer, line 268: What's the difference between this and "SignUiState"? Why do we need both?

src/client/pages/Offer/Checkout/index.tsx
```diff
@@ -422,10 +423,14 @@ export const Checkout = ({
...
-      ) {
+      const isManualReviewRequired = checkIsManualReviewRequired(
+        (error.graphQLErrors || []) as GraphQLError[],
+      )
+      if (isManualReviewRequired) {
+        setIsManualReviewRequired(isManualReviewRequired)
         setIsShowingFailModal(true)
+        setSignUiState('MANUAL_REVIEW')
+        return
       }
```
robinandeer, line 434: Do we need three separate pieces of state for this?

robinandeer (commented):
Great! Just wondering if we really need 3 separate pieces of state to handle this?

src/client/pages/Offer/Checkout/Sign.tsx
```diff
@@ -77,6 +77,11 @@ export const Sign: React.FC<Props> = ({
       ? textKeys.CHECKOUT_SIMPLE_SIGN_BUTTON_TEXT()
       : textKeys.CHECKOUT_SIGN_BUTTON_TEXT()
 
+  const hasSignStatusSection =
+    checkoutMethod &&
+    signUiState !== 'NOT_STARTED' &&
+    !(signUiState === 'STARTED' && checkoutMethod === CheckoutMethod.SimpleSign)
```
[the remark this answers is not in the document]
styrelius: It's rather that I don't want to render the sign status section when there's no text in it (because it looks weird) but this was not the best solution for that. I think we have a better solution in the old `Sign` component so maybe we could pick things from there. 🙃

src/client/pages/Offer/Checkout/index.tsx
styrelius: Probably not LOL 🙈 Let's look into that later.

PR #819, opened 2022-02-11
GRW-787 / Avoid using `OfferData` + `quoteDetails` with Quote Cart

robinandeer:
## What?

> Sorry for the huge PR. These things are tangled :P

Use `QuoteBundle` instead of `OfferData`.
Add a bunch more selectors.
Duplicate `trackOfferEvent`.
Duplicate `CheckoutSuccessRedirect`.
Duplicate `adtraction` event.

## Why?

`QuoteDetails` are deprecated in the API.
`OfferData` is hard to re-create from just a quote cart response so we should avoid it.

**Ticket(s): [GRW-787]**

[GRW-787]: https://hedvig.atlassian.net/browse/GRW-787?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robinandeer:
Tested locally
Screenshot 2022-02-11 at 15 15 06
Screenshot 2022-02-11 at 15 14 33

src/client/utils/tracking/trackSignedCustomerEvent.ts
```diff
@@ -0,0 +1,46 @@
...
+  memberId,
+  bundle,
+  campaignCode,
+  isDiscountMonthlyCostDeduction,
+}: TrackSignedEventParams) => {
+  const mainQuote = quoteBundleMainQuoteSelector(bundle)
+
+  if (variation === Variation.AVY) {
+    handleSignedEvent(memberId)
+  }
```
styrelius, line 30: Should we check whether we really need to keep all of these special Avy things in our codebase? From what I understand they _should_ only use our API's directly now and not the web onboarding.

src/client/api/quoteSelector.ts
```diff
@@ -0,0 +1,63 @@
+import { InsuranceType } from 'utils/hooks/useSelectedInsuranceTypes'
+import { BundledQuote, ApartmentType } from 'data/graphql'
+
```
styrelius, line 3: Maybe these functions would be more reusable if instead of a `quote` they just took `data` directly? Then they can be used even when we don't have access to the entire quote, but only the `data` object 🤔

src/client/api/quoteBundleHasAccident.ts
```diff
@@ -0,0 +1,5 @@
+import { QuoteBundle } from 'data/graphql'
+import * as quoteSelector from './quoteSelector'
```
styrelius, line 2: Just a question :) Why not directly import `isAccident` here?

src/client/api/quoteBundleIsBundleSelector.ts
```diff
@@ -0,0 +1,4 @@
+import { QuoteBundle } from 'data/graphql'
+
+export const quoteBundleIsBundleSelector = (bundle: QuoteBundle) =>
```
styrelius, line 3: The word `Selector` in this name is a little confusing to me. How about just `quoteBundleIsBundle` or something like `quoteBundleHasMultipleQuotes` or `quoteBundleIsTrueBundle`?

src/client/api/quoteBundleIsYouthOffer.ts
```diff
@@ -0,0 +1,4 @@
+import { QuoteBundle } from 'data/graphql'
+
+export const quoteBundleIsisYouthOffer = (bundle: QuoteBundle): boolean =>
```
styrelius, line 3: Typo 🙈 
```suggestion
export const quoteBundleIsYouthOffer = (bundle: QuoteBundle): boolean =>
```

src/client/utils/tracking/trackOfferEvent.ts
```diff
@@ -0,0 +1,74 @@
...
+          ),
+          switched_to: contractCategory,
+        }),
+        ...(memberId && { member_id: memberId }),
+        flow_type: EmbarkStory.get() ?? undefined,
+        current_insurer: mainQuote.currentInsurer?.id ?? undefined,
+      },
+      ...phoneNumberData,
+    })
+  } catch (e) {
```
styrelius, line 71: Nit:
```suggestion
  } catch (error) {
```

styrelius (commented):
G.R.E.A.T job with this 😅 👌 I'm pretty tired and probably need to check one more time but had a few comments for now.

src/client/api/quoteBundleHasAccident.ts
robinandeer: It's just another way to group the selectors so we don't have to name them all `quoteIsAccidentSelector`

src/client/api/quoteSelector.ts
robinandeer: Hmm do you have an example of such a use case? We are trying to find the right abstration-level.

Our best hunch now is to provide selectors for "bundle" and "quote".

guilhermespopolin (approved):
Nice refactoring

mehdielket (approved):
Nice cleanup and great solution

src/client/api/quoteSelector.ts
styrelius: I think there will be at least one case, in the "Summary"/"Your home" section on the checkout details page, where we only have access to the `data` object(s), but still want to be able to check for example if it's a student quote.

styrelius: But we can change that when we get there :)

PR #825, opened 2022-02-14
fix(GRW-788): displays NO bundle discount

guilhermespopolin:
## What?

- Fixes a bug on `CampaignBadge` component, responsible to skip the displaying of NO bundle discount (25%);
- Added support for the creation of _NO Home + Travel_ quotes to _Offer Debugger_

## Why?

The issue was happening because `CampaignBadge` was relying on the presence of an active campaign in order to build and display discount badges. However, that specific discount is not being added by the API (it feels like it should tho. Not sure 🤔)

## Screenshot(s)

Screenshot 2022-02-14 at 14 21 48
Screenshot 2022-02-14 at 14 21 55

## How to test it

1. Access provided review app;
2. Create a _Norwegian Home + Travel_ quote and then go to the offer page
3. Check if _25% Discount_ is visible both on the _InfoSidebar_ and _Checkout_ components

**Ticket(s)**
[GRW-788](https://hedvig.atlassian.net/browse/GRW-788)

### DIFF
--- a/src/client/components/CampaignBadge/CampaignBadge.tsx
+++ b/src/client/components/CampaignBadge/CampaignBadge.tsx
@@ -41,12 +41,12 @@ export const CampaignBadge: React.FC<CampaignBadgeProps> = ({
   })
   const campaignText = campaignData?.quoteCart.campaign?.displayValue ?? ''
 
-  if (loading || !campaignText) return null
-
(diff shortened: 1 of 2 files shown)


robinandeer (approved):
Amazing :D

mehdielket (approved):
🚀

PR #827, opened 2022-02-15
GRW-789-fix-total-price-discount

mehdielket:
## What?

It happens that after adding a discount code in the sidebar, the discounted price and the gross price are displayed as the same. This is due to the following:

* Sometimes the API adds the discount (in the campaign attribute) and does not update the prices. What is strange is a refresh of the page solves this issue... there could be some race condition. This was reported to the backend team.

Solutions:
* Moved away from `refetchQueries` that in my experience was a bit unreliable to leveraging the Apollo cache and extending the Add Campaign and remove campaign queries
* Hide the crossed gross price if the Gross and Net price are the same. This is how it is done in the checkout and it is a more reliable way to display the discount

**Ticket(s): [GRW-789]**

[GRW-789]: https://hedvig.atlassian.net/browse/GRW-789

https://user-images.githubusercontent.com/89857278/154061362-2c145cfd-16bb-495f-bfec-8f01524d9efb.mov

### [Review app](https://web-onboardi-grw-789-fi-ekzgpr.herokuapp.com/se-en/new-member)


src/client/graphql/RemoveCampaignCode.graphql
```diff
@@ -1,5 +1,22 @@
-mutation RemoveCampaignCode($quoteCartId: ID!) {
+mutation RemoveCampaignCode($quoteCartId: ID!, , $locale: Locale!) {
```
robinandeer, line 1: This just _looks_ odd => probably we can remove the extra "," (comma)

robinandeer (commented):
This looks even cleaner than the current solution! 🎉 

Let's just fix the extra comma :)

src/client/graphql/AddCampaignCode.graphql
```diff
@@ -1,8 +1,19 @@
-mutation AddCampaignCode($id: ID!, $code: String!) {
+mutation AddCampaignCode($id: ID!, $code: String!, $locale: Locale!) {
   quoteCart_addCampaign(id: $id, code: $code) {
     ... on QuoteCart {
       id
+      bundle {
```
guilhermespopolin, line 5: This addition produces the following warning while adding a new campaign code:

Screenshot 2022-02-15 at 10 56 00

Although it's not a current problem for us because a `QuoteCart` query is automatic triggered after `AddCampaignCode` mutation, it might be a problem in the future. The root of these problem is that `QuoteBundle`  type is not identifiable so every time we update this type we might be erasing some of the apollo cache. Not sure if we want to do something about it at this moment... 🤔

guilhermespopolin (commented):
Nice work. Added a couple of comments.

src/client/graphql/AddCampaignCode.graphql
robinandeer: Ahhh.... very tricky

guilhermespopolin (approved):

robinandeer (approved):

PR #828, opened 2022-02-16
Enable Quote Cart flows in NO 🎉

robinandeer:
## What?

Use quote cart feature flag to enable new Embark flows in NO

## Why?

So we can go live and roll back easily without having to change any code.

**Ticket(s): []**


dendoesit (approved):
🍾

guilhermespopolin (approved):

PR #829, opened 2022-02-16
GRW-803 / Fix Error Buttons Mobile

amandaflod:
## What?

 - Some fixes for the error modal to look a bit more cute on mobile screens 💅
 - New Outlined Warning Triangle added to the error modal. 

## Why?

**Ticket(s): [GRW-803]**

Screenshot 2022-02-16 at 16 27 20

[GRW-803]: https://hedvig.atlassian.net/browse/GRW-803?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/components/CloseButton/CloseButton.tsx
```diff
@@ -3,7 +3,7 @@ import styled from '@emotion/styled'
 import { colorsV3 } from '@hedviginsurance/brand'
 import { Cross } from '../icons/Cross'
 
-const Button = styled.button`
+export const StyledCloseButton = styled.button`
```
gustaveen, line 6: nit: Should we skip the Styled suffix for styled components? To me it feels a bit unnecessary and maybe confusing since we don't use it everywhere. I have probably written a few myself though :D

src/client/pages/Embark/ErrorModal.tsx
```diff
@@ -8,25 +8,43 @@ import { useCurrentLocale } from 'l10n/useCurrentLocale'
...
+  gap: 0.875rem;
 `
 
 const ButtonOutlined = styled(Button)`
   border: 1px solid ${gray900};
   background: ${white};
   color: ${gray900};
-  margin-left: 0.875rem;
+  padding: 0.375rem 0.75rem;
+  font-size: 0.875rem;
```
gustaveen, line 28: nit: maybe we could add this as a `sm` prop to button later on? :)

gustaveen (commented):
Looks great @amandaflod 🎉 Maybe we could add ErrorModal to Storybook as well? :)

src/client/components/CloseButton/CloseButton.tsx
amandaflod: Good one! I think I did it here just to be sure I hit the right button with the styling, but will fix 👍

src/client/pages/Embark/ErrorModal.tsx
amandaflod: Yes, we have some completely new button components coming along, but until then I decided to just do this tweak :)

src/client/pages/Embark/ErrorModal.tsx
```diff
@@ -8,25 +8,43 @@ import { useCurrentLocale } from 'l10n/useCurrentLocale'
...
   display: flex;
   justify-content: flex-start;
+  flex-wrap: wrap;
   margin-top: 1rem;
+  gap: 0.875rem;
 `
 
 const ButtonOutlined = styled(Button)`
   border: 1px solid ${gray900};
   background: ${white};
```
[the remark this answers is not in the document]
gustaveen: Haha set the comment in the wrong place 🙈 You're absolutely right, didn't think about the context of the modal :)

src/client/components/CloseButton/CloseButton.tsx
guilhermespopolin: If you don't mine I gonna add more fuel to the discussion ⛽ 

I would also prefer removing the _Styled_ prefix on this case, ending with something like `export const CloseButton = styled.button`. But what do you guys think it's a good name in the following scenario?

```javascript
const Button = styled.button``

const <insert-component-name-here> = styled(Button)``
```

I usually use the _Styled_ prefix on such cases, so I'd like to know your opinions about it.

guilhermespopolin (dismissed):
Looks good :)

src/client/components/CloseButton/CloseButton.tsx
amandaflod: I agree @guilhermespopolin, I am also a bit unsure about the conventions in those cases :)

styrelius (approved):
👍

PR #830, opened 2022-02-16
Remove auth token when loading offer page (Quote Cart)

robinandeer:
## What?

Make sure we don't include any auth token when loading quote cart offer page.

## Why?

This gets picked up by the API and causes strange things to happen like the campaign code not being applied when calculating the price.

This is because the auth token is linked to a member entity. However, on the quote cart offer page, we can be sure that there's no relevant member to keep track of. If there's an auth token present -- it's an artifact.

When we move over to quote cart for all markets, this will no longer be needed.

**Ticket(s): []**


dendoesit (approved):
👍

guilhermespopolin (approved):

PR #831, opened 2022-02-16
Fix campaign badge spacing

robinandeer:
## What?

Fix spacing between campaign badges.

There was an issue when they stacked on top of each other.

## Why?

I would like to use Flex gap but it doesn't have great support in browsers.

So I went with a margin hack I found online and references the source.
https://coryrylan.com/blog/css-gap-space-with-flexbox

**Ticket(s): []**

### Before

Screenshot 2022-02-16 at 17 38 25

### After

Screenshot 2022-02-16 at 17 33 58

Screenshot 2022-02-16 at 17 34 05

### DIFF
--- a/src/client/components/CampaignBadge/CampaignBadge.tsx
+++ b/src/client/components/CampaignBadge/CampaignBadge.tsx
@@ -12,7 +12,8 @@ const InlineFlexGap = styled.div`
   --gap: 0.5rem;
   display: inline-flex;
   flex-wrap: wrap;
-  margin: calc(-1 * var(--gap)) 0 0 calc(-1 * var(--gap));
+  margin-top: calc(-1 * var(--gap));
+  margin-left: calc(-1 * var(--gap));
   width: calc(100% + var(--gap));
 
   > * {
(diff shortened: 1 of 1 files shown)


src/client/components/CampaignBadge/CampaignBadge.tsx
```diff
@@ -7,15 +7,16 @@ import { useAppliedCampaignNameQuery } from 'data/graphql'
...
-  display: flex;
-  flex-direction: row;
-  justify-content: flex-start;
+/* https://coryrylan.com/blog/css-gap-space-with-flexbox */
+const InlineFlexGap = styled.div`
+  --gap: 0.5rem;
+  display: inline-flex;
   flex-wrap: wrap;
-  align-items: flex-start;
+  margin: calc(-1 * var(--gap)) 0 0 calc(-1 * var(--gap));
```
gustaveen, line 15: nit: Can be a bit hard to read long shorthands IMO and no need to set bottom and right margins :)
```suggestion
  margin-top: calc(-1 * var(--gap));
  margin-left: calc(-1 * var(--gap));
```

gustaveen (approved):
Awesome! 🙌
giphy (10)

gustaveen (approved):
🎉

PR #832, opened 2022-02-17
Enable NO quote cart routes in prod

robinandeer:
## What?

Disable the redirect for NO quote cart flows in production

## Why?

Luckily I noticed this already in the shadow environment! :D

**Ticket(s): []**


gustaveen (approved):
😌

PR #834, opened 2022-02-17
style(CancellationOptions): added vistual aid for labels

guilhermespopolin:
## What?

This ensures that only one current insurance can be requested to be cancel at a time.

## Why?

- It's not a big deal but Having visual aid on the cursor is a nice to have.
- Based in the code I had the sense that this should be the behaviour: cancel one at a time, since we're creating new bundles as way of editing quotes. But with the current implementation, one is still allowed to request the cancellation of multiple promises at a time. So I'm ensuring that should not be possible.

## Screenshots / recordings

https://user-images.githubusercontent.com/19200662/154912928-9c03913d-126d-4701-a47f-3bdc5dbd4492.mov


src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
```diff
@@ -10,12 +10,12 @@ const HandleSwitchingWrapper = styled.div`
...
 `
 
-const HandleSwitchingLabel = styled.label<{ isClickable: boolean }>`
+const HandleSwitchingLabel = styled.label<{ isEnabled: boolean }>`
   display: flex;
   align-items: center;
 
   &:hover {
-    cursor: ${({ isClickable }) => (isClickable ? 'pointer' : 'initial')};
+    cursor: ${({ isEnabled }) => (isEnabled ? 'pointer' : 'not-allowed')};
```
styrelius, line 18: I'm just thinking if the `default` cursor would be enough? Since that's what I think we use for all other disabled (otherwise clickable) elements, and I'm not sure how common it is to use `not-allowed` these days 🤔 

I guess it's clearer with the `not-allowed` but it's also... ugly and like, demands a lot of attention (🙈)? And yes, this second point is obviously mostly personal preference.

src/client/pages/OfferNew/Introduction/Sidebar/CancellationOptions.tsx
```diff
@@ -13,12 +13,12 @@ const HandleSwitchingWrapper = styled.div`
   padding: 0 0.25rem;
 `
 
-const HandleSwitchingLabel = styled.label<{ isClickable: boolean }>`
+const HandleSwitchingLabel = styled.label<{ isEnable: boolean }>`
```
styrelius, line 16: Typo 🙊 
```suggestion
const HandleSwitchingLabel = styled.label<{ isEnabled: boolean }>`
```

But also, would it make more sense to switch it and make it `isDisabled` instead of `isEnabled`?  Since it looks like that's more what we want to use it for (like when we pass it as `isLoading` and `disabled`).

styrelius (commented):
Really nice that you're looking at these types of UI/UX details 👌

src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
```diff
@@ -10,12 +10,12 @@ const HandleSwitchingWrapper = styled.div`
...
 `
 
-const HandleSwitchingLabel = styled.label<{ isClickable: boolean }>`
+const HandleSwitchingLabel = styled.label<{ isDisabled: boolean }>`
   display: flex;
   align-items: center;
 
   &:hover {
-    cursor: ${({ isClickable }) => (isClickable ? 'pointer' : 'initial')};
+    cursor: ${({ isDisabled }) => (isDisabled ? 'initial' : 'pointer')};
```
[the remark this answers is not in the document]
guilhermespopolin: No you're right. I've misunderstood the usage of `initial`

styrelius (approved):
🌟

PR #838, opened 2022-02-23
New translations

styrelius:
## What?

Downloaded translations.

## Why?

We need a fix for the `DETAILS_MODULE_TABLE_COINSURED_CELL_LABEL` text key which value should be equivalent to "Number of co-insured" but at some point was changed to "Number of insured people", which makes the value seem wrong in the details modal. This PR includes that fix (yes, it's a lot easier to change the label than to try to manipulate the value of the input field).

### Currently:
Screenshot 2022-02-23 at 16 01 45

### With these new translations:
Screenshot 2022-02-23 at 16 02 18


amandaflod (approved):
Looks good to me! 🤓

PR #843, opened 2022-03-01
GRW-858 Fix issues with editing bundles that include accident quotes

mehdielket:
## What?

How to reproduce the bug
* Create a bundle that includes Accident
* Try to edit a field (say first name) from the edit details form
* Expected: Update will be successful, modal closes
* Actual: Modal closes, but nothing is updated

This also affects the checkout form
* Edit first name in checkout form
* Click sign: Update fails, but we still proceed to signing

## Root cause

We were not passing the "isStudent" param as part of "data" in the `quoteCart_createQuoteBundle` mutation. The causes the API to return an error we do not handle.

## Solution

* Add isStudent as part of Data
* Handle these types of errors when they happen

**Ticket(s): [GRW-858]**

## Screenshots / recordings 

https://user-images.githubusercontent.com/89857278/156156641-4777a9fd-b662-481c-bf48-b304be56a063.mov

### [Review app](https://web-onboardi-grw-858-fi-9bg0ob.herokuapp.com/se/new-member/offer-debugger)

[GRW-858]: https://hedvig.atlassian.net/browse/GRW-858?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/pages/Offer/Checkout/index.tsx
```diff
@@ -477,7 +480,11 @@ export const Checkout = ({
         locale: locale.isoLocale,
         quoteCartId,
         quotes: getUniqueQuotesFromVariantList(quoteBundleVariants).map(
-          ({ startDate, currentInsurer, data: { type, typeOfContract } }) => {
+          ({
+            startDate,
+            currentInsurer,
+            data: { type, typeOfContract, isStudent },
+          }) => {
```
robinandeer, line 487: `isStudent` isn't returned for SE apartment/house quotes tho -- so is this really working?

My solution looked like this:

```
  const mainQuote = bundleSelector.getMainQuote(selectedQuoteBundle.bundle)
  ...

  const initialValues = {
    ...
    data: {
      ...mainQuoteData,
      isStudent: quoteSelector.isStudent(mainQuote),
    },
  } as QuoteInput
```

styrelius: I think `isStudent` is actually returned with the new quote cart API

Screenshot 2022-03-02 at 10 21 40

styrelius: Or maybe that's not what you meant @robinandeer 🙈

robinandeer: hmm, I didn't think it was.. but if it is, that makes things a lot easier!

src/client/pages/Offer/Checkout/index.tsx
```diff
@@ -416,6 +416,9 @@ export const Checkout = ({
       let quoteIds = getQuoteIdsFromBundleVariant(selectedQuoteBundleVariant)
       if (isFormDataUpdated) {
         const { data } = await submitForm()
+        const isUpdateQuotesFailed = isQuoteBundleError(data)
+        if (isUpdateQuotesFailed) throw Error('Updating quotes has failed')
```
robinandeer, line 420: Do we catch this error somewhere?

robinandeer (commented):
This works for me in testing :D

However, when I get an error I get multiple error messages...

Screenshot 2022-03-02 at 16 08 18

robinandeer:
Screenshot 2022-03-03 at 15 04 17

guilhermespopolin (approved):

PR #846, opened 2022-03-03
CPS / Check if quote cart is valid on checkout pages

styrelius:
## What?

- Check if quote cart ID is valid in `CheckoutPageWrapper` (i.e. running `QuoteCart` query and checking if we get an error - this should probably use its own request that only fetches the quote cart ID or something though, since `QuoteCart` query gets a LOT of data we don't need here).
- Add error modal if it's not valid, i.e. we get an error from quote cart query. This modal looks a lot like the one used on offer page but I build a new one since we might want to use different actions from the checkout page than from the offer page.
- Enable Intercom on checkout pages (although hiding the annoying "launcher" thingy since it will be places on top of CTA on mobile) - I created an `intercom.helpers.ts` file with one function for hiding the launcher and one for opening the chat).
- `Modal` prop `onClose` is made optional, and the close button is only rendered if that is passed.
- `useQuoteCartIdFromUrl` is added in a bunch of places, replacing `useParams<{ id: string }>`

## Why?

- We'd like to know immediately on the top level of the checkout pages whether or not the quote cart ID from the URL is a real one. And if not we should handle that error and provide the user with appropriate actions :)

**Ticket(s): [GRW-762]**

## Screenshots / recordings 

image

https://user-images.githubusercontent.com/42962286/156731330-a4b0d5c2-62c0-4c69-8f8c-e4f906c4a0ce.mov

### [Review app with valid quote cart ID](https://web-onboardi-grw-762-cp-zxisbe.herokuapp.com/dk-en/new-member/checkout/details/6789e5dd-ec58-4185-a5d7-d5aef4dbdbd0)

[GRW-762]: https://hedvig.atlassian.net/browse/GRW-762?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


amandaflod (approved):
Nice! Great that you can get to the chat directly from the modal 🙌

amandaflod (approved):
🥳

guilhermespopolin (approved):
👌🏻

PR #847, opened 2022-03-04
Skip `setupSession` if quote cart is enabled

robinandeer:
## What?

Remove "authorization" header resets if quote cart is enabled.

## Why?

Passing Authorization header with quote cart API:s confuses the backend so we should never do it.

My previous approach was to reset that header strategically. However, there are a lot of edge cases so better to avoid it entirely.

This new solution disables the `setupSession` function.

DRAWBACK: we need to go live with all embark flow in a market at once.

**Ticket(s): []**


amandaflod (dismissed):
Looks good to me! ⭐️

src/client/pages/Offer/Checkout/index.tsx
```diff
@@ -429,11 +429,6 @@ export const Checkout = ({
         }
       }
 
-      // clean up existing auth tokens
-      if (realApolloClient) {
-        realApolloClient.httpLink.options.headers.authorization = undefined
-      }
```
robinandeer, line 435: We already do this on offer page

dendoesit (approved):
looks good

guilhermespopolin (approved):

PR #849, opened 2022-03-04
fix(GRW-886): disabed start date inputs while there is an ongoing 'set start date' operation

guilhermespopolin:
## What?

Disable all _Start Date_ inputs while there is an ongoing "set start date" operation 

## Why?

As we're creating a new Quote Bundle when setting an start date, enabling setting it for two or more promises at the same time would not work as expected since one request would override the other.

**Ticket(s): [GRW-886](https://hedvig.atlassian.net/browse/GRW-886)**

## Screenshots / recordings 

https://user-images.githubusercontent.com/19200662/156765971-a205f3a1-40da-44a6-b702-2f77a28a2da4.mov

## Additional Notes

We might be experiencing what might be considering a performance issue while editing information on the _Offer Page_ (e.g: edit name, set _start date_, etc) because `creteQuoteBundle` mutation is taking up to 10s to finish.

Screenshot 2022-03-04 at 13 49 47

Another thing that I've noticed is that we enable users to choose _Start Date_ on two different places that don't share state between then: _Info Sidebar_ (what I could call Offer Summary) and _Checkout_. That might lead into "weird" behaviours. Maybe we could simplify things and enabling editing it in only one of the places, I dunno 🤔 

https://user-images.githubusercontent.com/19200662/156767267-5bfbdce2-efa0-4875-8c89-b33421e91db2.mov


styrelius (approved):
👌 

And yes, doesn't seem great with the extremely long loading times when you edit things on this page 🤔

robinandeer (approved):
Looks good! The only suggestion is to create some named variables so it's easier to understand this check

```
const whatIsCheckedHere = Boolean(quote.currentInsurer && !quote.startDate)
```

PR #851, opened 2022-03-08
GRW-889 / Fix / Make default body background light gray

robinandeer:
## What?

Remove overrides.
Update loading page to light gray.

## Why?

I think it's time. The default text color is already dark gray so makes sense that default background color it light 😅 

I tried verifying all pages.

This will also fix it so the scrollbar is dark and visible! ping @guilhermespopolin 

**Ticket(s): []**

## Screenshots / recordings 

Scrollbars:

Screenshot 2022-03-08 at 10 36 57

Loading page:

https://user-images.githubusercontent.com/1220232/157210931-ff647e50-6ff4-4735-9d70-e5deac0c6313.mov

### DIFF
--- a/src/client/components/LoadingPage.tsx
+++ b/src/client/components/LoadingPage.tsx
@@ -17,7 +17,7 @@ const OuterSpinnerWrapper = styled('div')`
   align-items: center;
   height: 100vh;
   width: 100%;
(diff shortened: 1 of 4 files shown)


amandaflod (approved):
🙌

PR #852, opened 2022-03-08
GRW-888 / Feature / Pickup saved campaign code on offer page

robinandeer:
## What?

Replace manual cookie handling with `CampaignCode` interface.
Don't overwrite campaign code if already added to quote cart.

## Why?

We want to send links directly to the offer page (magic link) with a campaign code when retargeting users that never signed.

**Ticket(s): []**

### Before
Screenshot 2022-03-08 at 11 27 24

### After (reload)
Screenshot 2022-03-08 at 11 27 30

### DIFF
--- a/src/client/pages/OfferNew/Introduction/Sidebar/DiscountCodeModal.tsx
+++ b/src/client/pages/OfferNew/Introduction/Sidebar/DiscountCodeModal.tsx
@@ -1,6 +1,5 @@
 import styled from '@emotion/styled'
 import { colorsV3 } from '@hedviginsurance/brand'
-import { CookieStorage } from 'cookie-storage'
 import { Form, Formik } from 'formik'
 import { motion } from 'framer-motion'
 import hexToRgba from 'hex-to-rgba'
@@ -11,6 +10,7 @@ import { useRedeemCodeMutation } from 'data/graphql'
 import { InputField } from 'components/inputs'
 import { Button } from 'components/buttons'
 import { CloseButton } from 'components/CloseButton/CloseButton'
+import { CampaignCode } from 'utils/campaignCode'
 
 interface Props {
   isOpen: boolean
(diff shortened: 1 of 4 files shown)


amandaflod (approved):
🪄

src/client/utils/campaignCode.ts
```diff
@@ -9,7 +9,7 @@ const getPersistedCampaignCode = () => {
 }
 
 const saveCampaignCode = (campaignCode: string) => {
-  cookieStorage.setItem(CAMPAIGN_CODE_COOKIE_KEY, campaignCode)
+  cookieStorage.setItem(CAMPAIGN_CODE_COOKIE_KEY, campaignCode, { path: '/' })
```
guilhermespopolin, line 12: Does `{ path: '/' }` make the cookie availability through the whole app (every subdoman)? 🤔

src/client/pages/OfferNew/Introduction/Sidebar/DiscountCodeModal.tsx
```diff
@@ -1,6 +1,5 @@
 import styled from '@emotion/styled'
```
guilhermespopolin, line 1: There's a `CampaignCodeModal` under `Offer` folder. Should we change there as well?

src/client/utils/campaignCode.ts
robinandeer: I don't think it applies to subdomains but the Cookie header will be sent on all URL paths:

https://developer.mozilla.org/en-US/docs/Web/HTTP/Cookies#path_attribute

This is just how it was setup before, that's why I included it.

src/client/pages/OfferNew/Introduction/Sidebar/DiscountCodeModal.tsx
robinandeer: It's already implemented there, no? In this function: `handleAddCampaignCode`

guilhermespopolin (approved):

PR #855, opened 2022-03-09
bug(GRW-890): apollo cache override

guilhermespopolin:
## What?

* Provide to _apollo_ InMemoryCache a way to generate an id for `QuoteBundle` types: it relies on quote ids;
* Update _graphql_ quoteCart related queries/mutations so we fetch quote ids while working with that type;

## Why?

`QuoteCart` type [defines a field called `bundle`](https://github.com/HedvigInsurance/giraffe/blob/master/src%2Fresolvers%2Fquote-cart%2FquoteCart.graphqls#L89) which type, [`QuoteBundle`](https://github.com/HedvigInsurance/giraffe/blob/master/src%2Fresolvers%2Fquote-bundle%2FquoteBundle.graphqls#L9), doesn't have a explicit form of identification. As a consequence, we might face some issues with _apollo_ caching as some queries might override the cache which may cause problems. 

To solve this issue we can provide to _apollo_ `InMemoryCache` a way to define a custom id for `QuoteBundle` type and make sure we fetch the necessary data for that id be built while working with that type.

**Ticket(s): [GRW-890](https://hedvig.atlassian.net/browse/GRW-890)**

## Drawbacks

* We need to keep in mind that it's **really important** to request `quotes { id }` while working with `QuoteBundle` type. Otherwise the custom id generation will not work;
* Apollo cache increases significantly  during editions because we don't clear cache stale data while creating a new quote bundle. I've tested with `editQuote` api real quick and it seems this problem doesn't happen with it (which make sense, since we're not creating new quotes but just updating them).

## How test it

You can access the [review app](https://web-onboardi-bug-grw-89-ynytlz.herokuapp.com/se-en/new-member/offer/3934f0ce-5118-4123-8d59-9a5bd28b9673) and check _Cache_ tab into _Apollo Dev Tools_. It should be able to see that `QuoteBundle` types are being normalised:

Screenshot 2022-03-09 at 13 23 59

### Related with

[chore: update due graphl api type changes](https://github.com/HedvigInsurance/web-onboarding/pull/856)

### DIFF
--- a/src/client/apolloClient.ts
+++ b/src/client/apolloClient.ts
@@ -11,6 +11,7 @@ import { SubscriptionClient } from 'subscriptions-transport-ws'
 import { getMainDefinition } from '@apollo/client/utilities'
 import possibleTypes from '../../possibleGraphqlTypes.json'
 import { createSession, Session } from '../shared/sessionStorage'
+import { QuoteBundle } from './data/graphql'
 
 export interface ApolloClientUtils {
   subscriptionClient: SubscriptionClient
@@ -61,6 +62,20 @@ export const apolloClient = (() => {
         },
         InProgressReferral: {
           keyFields: (referral) => referral.name as string,
+        },
+        QuoteBundle: {
+          keyFields: (quoteBundle) => {
+            const quotes = (quoteBundle as QuoteBundle).quotes ?? []
+
+            if (quotes.length === 0) return
+
+            // QuoteBundle:<id>+<id>
+            return quotes.reduce(
+              (id, quote, index) =>
+                (id += `${quote.id}${index < quotes.length - 1 ? '+' : ''}`),
+              'QuoteBundle:',
+            )
+          },
         },
       },
     }),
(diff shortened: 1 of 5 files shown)


src/client/apolloClient.ts
```diff
@@ -62,6 +63,20 @@ export const apolloClient = (() => {
...
+          keyFields: (quoteBundle) => {
+            const quotes = (quoteBundle as QuoteBundle).quotes ?? []
+
+            if (quotes.length === 0) return
+
+            // QuoteBundle:<id>+<id>
+            return quotes.reduce(
+              (id, quote, index) =>
+                (id += `${quote.id}${index < quotes.length - 1 ? '+' : ''}`),
+              'QuoteBundle:',
```
robinandeer, line 76: hehe isn't this overly complicated?

```suggestion
            // QuoteBundle:<id>+<id>
            return `QuoteBundle:${quotes.map((quote) => quote.id).join('+')}`
```

robinandeer (commented):
Neat enough solution I think since we are dealing with a resource that's missing an ID which from what I understand is different from how GraphQL is intended to work.

src/client/apolloClient.ts
guilhermespopolin: OMG hahhaha
So much better

guilhermespopolin:
We decided no go with this solution.

PR #845, opened 2022-03-01
Feature/grw 731 tooltip

guilhermespopolin:
## What?

- Updates `Tooltip` by using a dependency that takes care of best positioning  a tooltip on the screen
- Adds that tooltip to `CancellationOptions` as part of CPS

**Ticket(s): [GRW-731](https://hedvig.atlassian.net/browse/GRW-731)**

## Screenshots / recordings 

https://user-images.githubusercontent.com/19200662/156191033-ed11762f-abd3-426c-89e5-e91af0473134.mov

PR #857, opened 2022-03-09
Center align tooltip in `CancellationOptions`

robinandeer:
## What?

Screenshot 2022-03-09 at 16 11 06

## Why?

**Ticket(s): []**


guilhermespopolin:
This was implemented in that way by [design](https://github.com/HedvigInsurance/web-onboarding/pull/845#pullrequestreview-897515590).
Ellen mentioned about it in the PR that adds this tooltip and. So I've checked with Amy and she prefer aligned at the top.

robinandeer:
@guilhermespopolin haha oh 😅 but it can't be right that the left icon is centered and right is not?

guilhermespopolin:
> @guilhermespopolin haha oh 😅 but it can't be right that the left icon is centered and right is not?

I agree with you 🤷🏻

PR #858, opened 2022-03-09
Fix small cookie issues

robinandeer:
## What?

Remove cookie before removing campaign.
Include same option when removing cookie as setting.
Remove custom cookie settings for device id.

## Why?

I ran into an issue where you couldn't remove the campaign code without it getting added automatically again.

**Ticket(s): []**

### DIFF
--- a/src/client/pages/Offer/Introduction/Sidebar/index.tsx
+++ b/src/client/pages/Offer/Introduction/Sidebar/index.tsx
@@ -163,15 +163,8 @@ export const Sidebar: React.FC<SidebarProps> = ({
   )
 
   const handleRemoveCampaign = useCallback(async () => {
-    const { data: result } = await removeCampaignCode({
-      variables: { quoteCartId },
-    })
-    const hasError =
-      result?.quoteCart_removeCampaign.__typename === 'BasicError'
-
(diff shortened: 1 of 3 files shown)


guilhermespopolin (approved):

src/client/pages/Offer/Introduction/Sidebar/index.tsx
```diff
@@ -163,15 +163,8 @@ export const Sidebar: React.FC<SidebarProps> = ({
...
-      variables: { quoteCartId },
-    })
-    const hasError =
-      result?.quoteCart_removeCampaign.__typename === 'BasicError'
-
-    if (!hasError) {
-      CampaignCode.remove()
-    }
+    CampaignCode.remove()
+    await removeCampaignCode({ variables: { quoteCartId } })
```
guilhermespopolin, line 167: We were checking if the API worked before removing it from cookie storage. Wouldn't be a potencial problem not to check?

robinandeer: I don't really see that. It's more of a bonus that we remember the code in the first place :)

PR #860, opened 2022-03-10
Remove dismissing cookie banner in e2e tests

robinandeer:
## Why?

It was failing. @gustaveen turned off the banner in staging.

**Ticket(s): []**


guilhermespopolin (approved):

gustaveen (approved):
Nice! :D What do you think in the long run, should we have the cookie banner in staging or not?

robinandeer:
@gustaveen we can leave it turned off for me -- maybe treat it like norway credit check and make it possible to enable it for the few use cases where we want to test it

PR #863, opened 2022-03-11
DO NOT MERGE! Switch fonts from Favorit to Hedvig Letters Standard

gustaveen:
## What?
Switch fonts from Favorit to Hedvig Letters Standard. 
Will create a review app for Zak to look at. The timeline is to go live with the new fonts on april 18th

## Why?
💅


src/client/pages/ConnectPayment/components/TrustlyModal.tsx
```diff
@@ -9,11 +9,11 @@ import { useTextKeys } from 'utils/textKeys'
 const Header = styled('div')({
   width: '100%',
   height: 40,
-  background: colors.DARK_GREEN,
```
gustaveen, line 12: Let's see if this one should still be green to look more like Trustly

PR #868, opened 2022-03-14
Feature/grw-648/Checkout-details

dendoesit:
## What?

Added the checkout details page together under one single big request
Created one bigger request in useQuoteCartData that includes everything we need in the Checkout. 
Removed the useGetPriceData hook
Added z-index to StartDate component.
Changed the way components in Checkout Details get data : Footer, YourPlan, QuoteDetails.

## Why?

New CPS page for Checkout Details. 

**Ticket(s): []**
https://hedvig.atlassian.net/browse/GRW-648

## Testing: 

Using the offer-debugger create a new offer.
Go to the offer page and click next on NO/DK .
Alternative is to get to the offer page and replace the url part '/offer' with /checkout/details

image
image

P.S: sorry I know this is a bit long.
P.S.2 : yes, I am still working on types.


src/client/api/quoteDetailsDataTypes.ts
```diff
@@ -66,3 +66,34 @@ export type HomeQuoteDetails =
   | QuoteDataDkHomeContents
 
 export type HouseQuoteDetails = QuoteDataSeHouse
+
+export type GenericQuoteData = {
```
dendoesit, line 70: I noticed this in @robinandeer 's PR so I'll import it from there.

robinandeer: haha yeah I know it's not the most generic to type the field that way but it's better than just using "any", right? 😅

robinandeer: Long term we should not access any of this data client side!

src/client/pages/Checkout/CheckoutDetails/CheckoutDetails.tsx
```diff
@@ -2,21 +2,40 @@ import React from 'react'
...
 
 export const CheckoutDetails = () => {
   const textKeys = useTextKeys()
   const { path: localePath } = useCurrentLocale()
   const { quoteCartId } = useQuoteCartIdFromUrl()
-
+  const data = useQuoteCartData()
+  if (!data) {
+    return null
+  }
```
robinandeer, line 21: How does this look for users?

src/client/pages/Checkout/CheckoutDetails/components/QuoteDetails/QuoteDetails.tsx
```diff
@@ -56,26 +54,13 @@ const getValueText = ({
...
-  const quoteDetailsData = data?.quoteCart.bundle?.quotes?.map(
-    ({ data }) => data,
-  )
-
-  if (!quoteDetailsData) {
-    return null // TODO: Do something more sophisticated
-  }
-
-  const quoteDetails = getQuoteDetails({ quoteDetailsData })
+  const quoteDetails: any[] = quoteData.quoteData
```
robinandeer, line 59: Let's try to avoid `any` when we are writing new features/code, right?

src/client/pages/Checkout/CheckoutDetails/components/YourPlan/YourPlan.tsx
```diff
@@ -40,15 +40,14 @@ const HorizontalDivider = styled(Divider)`
   margin: 0.5rem 0;
 `
 
-export const YourPlan = () => {
-  const priceData = useGetPriceData()
+export const YourPlan = (priceData: any) => {
```
robinandeer, line 43: Strongly consider typing these props ;)

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
```diff
@@ -1,18 +1,24 @@
...
+    return null
+  }
+  const priceData = data.priceData
   return (
     <CheckoutPageWrapper>
       <Footer
         buttonText={textKeys.CHECKOUT_FOOTER_COMPLETE_PURCHASE()}
         buttonOnClick={() => {
           console.log('click')
         }}
```
robinandeer, line 20: Is this a placeholder?

src/client/pages/Checkout/shared/Footer.tsx
```diff
@@ -55,7 +62,7 @@ export const Footer = ({ buttonText, buttonOnClick, buttonLinkTo }: Props) => {
     <Wrapper>
       <InnerWrapper>
         <PaymentInfoWrapper>
-          <PaymentInfo />
+          <PaymentInfo {...priceData} />
```
robinandeer, line 65: Any way we can use composition to avoid props drilling here?

src/client/pages/Checkout/CheckoutDetails/components/QuoteDetails/QuoteDetails.tsx
dendoesit: yeah, still working on that one.

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
dendoesit: yes.

src/client/pages/Checkout/CheckoutDetails/CheckoutDetails.tsx
dendoesit: Will be handled in the error + loading state PR.

mehdielket (commented):
Nicely done: A couple of comments here

src/client/utils/hooks/useQuoteCartData.ts
```diff
@@ -1,21 +1,218 @@
...
+
+  const getPriceData = (): PriceData => {
+    return {
+      prices: prices,
+      discount: selectedQuoteBundle.bundle.bundleCost.monthlyDiscount.amount,
+      currency: selectedQuoteBundle.bundle.bundleCost.monthlyNet.currency,
+      totalBundleCost: selectedQuoteBundle.bundle.bundleCost.monthlyNet.amount,
+      campaignName: data?.quoteCart.campaign?.displayValue,
+    }
+  }
```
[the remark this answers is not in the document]
dendoesit: changed this to an object.

src/client/utils/hooks/useQuoteCartData.ts
```diff
@@ -1,21 +1,218 @@
...
+  const prices = selectedQuoteBundle?.bundle.quotes.map((item) => {
+    return { displayName: item.displayName, price: item.price.amount }
   })
 
+  const mainQuote = getMainQuote(selectedQuoteBundle.bundle)
+
+  const getPriceData = (): PriceData => {
+    return {
+      prices: prices,
+      discount: selectedQuoteBundle.bundle.bundleCost.monthlyDiscount.amount,
```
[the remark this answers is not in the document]
dendoesit: sure, I can do something like this. 
`export const getDiscountAmount = (bundle: QuoteBundle) => {
  return bundle.bundleCost.monthlyDiscount.amount
}`

I do think that kind of adds some more complexity to the file as we add more lines and instead of maybe directly getting a value from the scope we do a function hosted in another file 🤔 . what do you think ?

src/client/pages/Checkout/CheckoutDetails/components/QuoteDetails/QuoteDetails.tsx
```diff
@@ -56,48 +54,31 @@ const getValueText = ({
...
-  const quoteDetailsData = data?.quoteCart.bundle?.quotes?.map(
-    ({ data }) => data,
-  )
-
-  if (!quoteDetailsData) {
-    return null // TODO: Do something more sophisticated
-  }
-
-  const quoteDetails = getQuoteDetails({ quoteDetailsData })
+  console.log(groups)
```
robinandeer, line 60: Remove :)

src/client/pages/Checkout/CheckoutDetails/components/YourPlan/YourPlan.tsx
```diff
@@ -40,20 +40,13 @@ const HorizontalDivider = styled(Divider)`
   margin: 0.5rem 0;
 `
 
-export const YourPlan = () => {
-  const priceData = useGetPriceData()
+export const YourPlan = (data: PriceData) => {
```
robinandeer, line 43: Consider adding like a regular React component:
```suggestion
export const YourPlan = ({ data }: Props) => {
```

src/client/pages/Checkout/CheckoutDetails/components/YourPlan/YourPlan.tsx
```diff
@@ -65,13 +58,13 @@ export const YourPlan = () => {
...
-            <Label>{displayName}</Label>
-            <Value>{formattedPrice(price.amount)}</Value>
+        {prices.map((item: Price) => (
+          <Row key={item.displayName}>
+            <Label>{item.displayName}</Label>
+            <Value>{formattedPrice(item.price)}</Value>
           </Row>
         ))}
-        {Number(discount) > 0 && (
+        {Number(discount) > 0 && discount && (
```
robinandeer, line 67: Shouldn't this be:

```suggestion
        {discount && Number(discount) > 0 && (
```

Why isn't it being passed as a bumber to the component?

src/client/utils/hooks/useQuoteCartData.ts
```diff
@@ -1,21 +1,212 @@
...
+  const priceData = {
+    prices: prices,
+    discount: selectedQuoteBundle.bundle.bundleCost.monthlyDiscount.amount,
+    currency: selectedQuoteBundle.bundle.bundleCost.monthlyNet.currency,
+    totalBundleCost: selectedQuoteBundle.bundle.bundleCost.monthlyNet.amount,
+    campaignName: data?.quoteCart.campaign?.displayValue,
+  }
+  const quoteDetails = mainQuote.data
+
+  const getHomeDetails = (): DetailsGroup => {
```
robinandeer, line 59: Redefining these inside the hook will not be good for performance (at least not without `useCallback`) - consider moving them outside the hook :)

src/client/utils/hooks/useQuoteCartData.ts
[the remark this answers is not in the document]
mehdielket: The idea behind the selectors was to have one place that is tightly coupled to the GQL api and that the components are API agnostic as possible. Getting the amount is something we do in a lot of other places, so I am sure this function can/will be reused.

src/client/pages/Checkout/CheckoutDetails/CheckoutDetails.tsx
```diff
@@ -2,22 +2,43 @@ import React from 'react'
...
+  const priceData = data.priceData
+  const quoteDetails = data.quoteDetails
   const paymentPageLink = `/${localePath}/new-member/checkout/payment/${quoteCartId}`
-
   return (
     <CheckoutPageWrapper>
+      <PageSection>
+        <YourPlan {...priceData} />
+        <SubSection headlineText={textKeys.SIDEBAR_STARTDATE_CELL_LABEL()}>
+          <StartDate quoteCartId={quoteCartId} modal size="sm" />
```
amandaflod, line 31: This start date components now takes up full width on desktop which looks a bit crazy 😅 Maybe would be good to set a max-width?

dendoesit: ah. good catch. 👍

amandaflod (approved):
Nice job Dan! I think this is good to go now 🚢🎉

mehdielket (commented):
I noticed a couple of fixes needed for the design:

- The Edit Information link is missing
- The Terms links section is missing
- Desktop: Padding around the date field & the top of the summary seems a bit off compared to design
- Mobile: The design does not seem to be working properly at the moment.


Screenshot 2022-03-17 at 14 51 12
Screenshot 2022-03-17 at 14 53 41

src/client/utils/hooks/useQuoteCartData.ts
```diff
@@ -1,21 +1,222 @@
...
   })
 
+  const mainQuote = getMainQuote(selectedQuoteBundle.bundle)
+
+  const priceData = {
+    prices: prices,
+    discount: getDiscountAmount(selectedQuoteBundle.bundle),
+    currency: getBundleCurrency(selectedQuoteBundle.bundle),
+    totalBundleCost: getTotalBundleCost(selectedQuoteBundle.bundle),
+    campaignName: data?.quoteCart.campaign?.displayValue,
```
mehdielket, line 199: There is already a selector that gets you the campaign from the query under `api/quoteCartQuerySelectors.ts`

So something like this should work
```suggestion
    campaignName: getCampaign(data)?.displayValue,
```

dendoesit: that's great 👍

dendoesit:
@mehdielket Fixed all issues with padding and mobile layout. 

The Edit Information link is missing
The Terms links section is missing
These two were removed from the V0 version in order to be able to go to market faster. Also, you have both these option in the previous Offer screen. I added a separate ticket for them in Jira but they should not be blockers for this PR.

mehdielket (approved):
Looks good! 🚀

PR #870, opened 2022-03-15
Default to selecting least expensive bundle

robinandeer:
## What?

If no bundle is explicitly selected, default to the one with the fewest quotes.

## Why?

Seems like a dark pattern to select the most expensive bundle by default.

**Ticket(s): []**


robinandeer:
I discussed this with Amy and @zakkjellin and they agree that this is the best approach

PR #872, opened 2022-03-16
Fix/apple pay error

FredrikAreschoug:
While testing apple pay with @robinandeer we got this error:
`Blocked a frame with origin "https://www.dev.hedvigit.com/" from accessing a frame with origin "https://checkoutshopper-test.adyen.com/". Protocols, domains, and ports must match.`
I wonder if it is this easy to fix it


gustaveen (approved):
Hm, this might do the trick. But I'm unsure if that's a warning because of the content security policy. But let's try it!

PR #874, opened 2022-03-17
Skip validation of campaign code

robinandeer:
## Why?

Since the backend returns error for all forever codes.

**Ticket(s): []**

### DIFF
--- a/src/client/pages/Forever/index.tsx
+++ b/src/client/pages/Forever/index.tsx
@@ -61,25 +61,9 @@ export const Forever: React.FC<ForeverProps> = ({
   const textKeys = useTextKeys()
   const history = useHistory()
 
-  const { refetch: fetchCampaign } = useCampaignQuery({ skip: true })
-
-  const handleSubmit = async (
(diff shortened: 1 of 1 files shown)


amandaflod (approved):
🧹

src/client/pages/Forever/index.tsx
```diff
@@ -61,25 +58,9 @@ export const Forever: React.FC<ForeverProps> = ({
   const textKeys = useTextKeys()
   const history = useHistory()
 
-  const { refetch: fetchCampaign } = useCampaignQuery({ skip: true })
```
guilhermespopolin, line 64: So it seems `useCampaignQuery` is not being used anymore and, therefore, we could remove it completely. That means get rid of `Campaign.graphql` file, righ?!

guilhermespopolin (approved):

src/client/pages/Forever/index.tsx
robinandeer: Yep, will remember to clean that up later!

PR #876, opened 2022-03-18
Grw 650/checkout payment

dendoesit:
## What?

Set up the checkout-payment page so we can start working on forms and API.

## Why?

Part of CPS

**Ticket(s): []**
https://hedvig.atlassian.net/browse/GRW-650

image


src/client/pages/Checkout/CheckoutPayment/ContactInformation/ContactInformation.tsx
```diff
@@ -10,13 +10,14 @@ import { MEDIUM_SMALL_SCREEN_MEDIA_QUERY } from 'utils/mediaQueries'
...
 export type ContactInfoData = {
-  firstName?: string
-  lastName?: string
-  ssn?: string
-  email?: string
+  firstName?: string | null | undefined
+  lastName?: string | null | undefined
+  ssn?: string | null | undefined
+  email?: string | null | undefined
 }
```
robinandeer, line 17: Isn't this just exposing strange things from the API? I think we should avoid letting it through to the React components. Let's sort that out in the hook :)

Seems like in this component we only want defined strings. So just type it that way - we should not care about API stuff here.

src/client/pages/Checkout/CheckoutPayment/PaymentDetails/PaymentDetails.tsx
```diff
@@ -5,10 +5,15 @@ import { useTextKeys } from 'utils/textKeys'
...
 import {
   AdyenCheckout,
-  Wrapper,
+  Wrapper as AdyenWrapper,
 } from '../../../ConnectPayment/components/AdyenCheckout'
 const { gray100, gray600, gray700, gray300, gray900 } = colorsV3
 
+const Wrapper = styled.div`
+  margin: 0 auto;
+  max-width: 628px;
```
robinandeer, line 14: Do we have a constant for this? Maybe we should use rems at least?

robinandeer (commented):
Wop wop!

src/client/pages/Checkout/CheckoutPayment/PaymentDetails/PaymentDetails.tsx
dendoesit: rems is a bit strange for this. it would be 39.25. 

I will add a const WrapperWidth and we can use it in all files part of the Checkout

src/client/pages/Checkout/CheckoutPayment/ContactInformation/ContactInformation.tsx
```diff
@@ -49,7 +47,19 @@ const Spacer = styled.div`
...
-export const ContactInformation = (data: ContactInfoData) => {
+const HorizontalDivider = styled(Divider)`
+  margin: 1.5rem 0;
+  ${MEDIUM_SMALL_SCREEN_MEDIA_QUERY} {
+    margin: 3rem 0;
+  }
+`
+
+export const ContactInformation = ({
+  firstName,
```
mehdielket, line 58: You could default here to empty string if you want I believe. Same with last name
```suggestion
  firstName = '',
```

amandaflod (approved):
I think this is ready to merge! :D

PR #878, opened 2022-03-18
Fix address display without floor

robinandeer:
## What?

Update `GenericQuoteData`: floor is always a number.
Refactor `getAddress` to not include floor if "0".

## Why?

The floor was displayed for all addresses (even house) but shouldn't.

**Ticket(s): [GRW-924]**

[GRW-924]: https://hedvig.atlassian.net/browse/GRW-924?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dendoesit (approved):
great stuff.

PR #883, opened 2022-03-25
Grw 920/feat/close button in details modal

simonauner:
## What?

* Add Close button and info text to Offer's Customer Info Modal
* Add `border` prop to `Button` so that we can easily style it to same color as `foreground`
* Add `ButtonGroup` component that creates a flex div for buttons to be in with column-gap (can be extended to be vertical in the future) 
* Add ModalFooter component with `sticky` prop - this enables us to reuse the interaction design from the Figma sketch (https://www.figma.com/file/F6P9t3lzTsmvADTT5MPHXN/Quote-in-10-seconds?node-id=404%3A5288)
* Make Modal's background color "off-white" according to above Figma sketch

**Waiting for:**

- [x] translations

## Why?

**Ticket(s): [GRW-920]**

image

image

[GRW-920]: https://hedvig.atlassian.net/browse/GRW-920?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/components/ModalNew.tsx
```diff
@@ -157,3 +157,26 @@ export const Modal: React.FC<ModalProps> = ({
...
+    bottom: 0;
+`}
+`
+type ModalFooterProps = {
+  sticky?: boolean
+}
+
+export const ModalFooter: React.FC<ModalFooterProps> = ({ children }) => {
+  return <Footer>{children}</Footer>
+}
```
robinandeer, line 182: Nit: we avoid using `React.FC`. You can read more here: https://www.notion.so/hedviginsurance/React-components-props-f13a8f7ebfae4ee0b57fe47b55b6a156#7ad1d210a58a4ea4a4814c3b682e13be

At least for new code :)

src/client/components/buttons.tsx
```diff
@@ -37,7 +50,10 @@ export const Button = styled(UnstyledButton)<ButtonProps>`
   background: ${({ background }) => background ?? gray900};
   color: ${({ foreground }) => foreground ?? white};
   border-radius: 8px;
-  border: none;
+  ${({ border, foreground }) =>
+    css`
+      border: ${border ? `1px solid ${foreground ?? white}` : 'none'};
+    `}
```
robinandeer, line 56: haha wow that's a lot of nesting 😅

robinandeer (approved):
Looks good! We should probably have some secondary/primary styles for buttons when we implement the UI kit but this works for now :)

src/client/components/buttons.tsx
simonauner: 🤯

robinandeer (dismissed):
re-approve

dendoesit (approved):
looks good 👍

PR #884, opened 2022-03-25
fixed StartDate jumping

marian-buc:
## What?

This prevents the date element from jumping around when it's rerendered

## Why?

React re-renders the component everytime date changes, and since the Error component was not init'ed with display:'none' it would waste a render cycle when displaying it and then hiding, and also causing a visual bug when CPU performance is lower (I could only replicate the bug with the x6 CPU throttle in dev tools, but then it appears everytime). Fix was very simple after figuring out exactly what was causing that.

**Ticket(s): [GRW-917]**

Previously:
image

[GRW-917]: https://hedvig.atlassian.net/browse/GRW-917?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
```diff
@@ -357,7 +357,7 @@ export const StartDate = ({
     <>
       <ErrorMessage
         aria-hidden={!showError}
-        initial={{ height: 0, opacity: 0 }}
+        initial={{ height: 0, opacity: 0 , display:'none'}}
```
robinandeer, line 360:
```suggestion
        initial={{ height: 0, opacity: 0 , display: 'none' }}
```

robinandeer (commented):
Looks good, just that linting fails in CI

You can test locally with `yarn lint` I think

dendoesit (approved):
looks good.

guilhermespopolin (approved):

robinandeer (approved):

PR #888, opened 2022-03-28
GRW-918/chat-button-checkout-position

marian-buc:
## What?

This is my (good/bad?) attempt at providing a better UX with the Intercom button on the Checkout pages. The button is now raised above the bar, as well as automatically hiding and showing up on mouse move and on scroll. 
I've also added a footerMargin prop to CheckoutPageWrapper, so now the layout fits perfectly for both pages

## Why?

Previously the Intercom chat button was overlapping the footer bar and buttons, as well as form elements on mobile. There's no way to hook into Intercom loaded events or React component so I had to use good old querySelectors and eventListeners.

**Ticket(s): [GRW-918]**

https://user-images.githubusercontent.com/101566411/160474189-d6eb4d37-88d6-46e1-9158-2b888fa8dd96.mp4

CheckoutPayment
image

CheckoutDetails
image

[GRW-918]: https://hedvig.atlassian.net/browse/GRW-918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonauner:
I like the idea! Maybe we should only do the hiding when the user has scrolled to the bottom of the page? I'm thinking that users can scroll the page to see whatever is hidden behind the button in all other situations than when they're at the bottom of the page. As such we don't need to disturb the user with hiding/showing elements unless absolutely necessary. 

I'm guessing that the chatbutton is absolutely positioned - perhaps instead of `opacity` we can just move it _almost_ off-screen so that users only see a bit of the circle? With `opacity` we could end up in the unfortunate situation that the user clicks the invisible element at least on mobile when we don't have mousemove events.

src/client/utils/Intercom.tsx
```diff
@@ -47,3 +47,40 @@ export const Intercom: React.FC = () => {
...
+        clearTimeout(intercomTimeout.current)
+      }
+      buttonEl.style.opacity = '1'
+      intercomTimeout.current = setTimeout(() => {
+        buttonEl.style.opacity = '0'
+      }, 800)
+    }
+  }
+  useEffect(() => {
+    window.addEventListener('scroll', onInteraction)
```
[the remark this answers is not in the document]
marian-buc: Yeah, I guess we should debounce them a little

src/client/utils/Intercom.tsx
```diff
@@ -47,3 +47,40 @@ export const Intercom: React.FC = () => {
...
+    <style>
+      {`
+    ${intercomBtnSelector}{
+      bottom: 87px!important;
+      transition: opacity 150ms ease-out;
+      opacity: 1;
+    }
+   `}
+    </style>
+  )
```
robinandeer, line 85: I've never seen this pattern before. I think it would be better to use the built-in global styles from Emotion.

robinandeer (commented):
Interesting approach!

Perhaps we can name the component something more descriptive: `<IntercomHideOnScroll>` or something?

Maybe we can use the official JS API to hide/show the bubble? https://developers.intercom.com/installing-intercom/docs/intercom-javascript

src/client/utils/Intercom.tsx
```diff
@@ -47,3 +47,40 @@ export const Intercom: React.FC = () => {
...
+      buttonEl.style.opacity = '1'
+      intercomTimeout.current = setTimeout(() => {
+        buttonEl.style.opacity = '0'
+      }, 800)
+    }
+  }
+  useEffect(() => {
+    window.addEventListener('scroll', onInteraction)
+    window.addEventListener('mousemove', onInteraction)
+    // onInteraction()
```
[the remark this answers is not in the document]
marian-buc: Yep, also the unmount. Regarding the previous comment, currently, the element is visible until the first interaction, and afterwards it gets hidden. On longer mobile screens there's no scroll, but the mousemove does trigger an interaction tick event on touchscreens so they can't click the invisible icon, it will be shown on tap

src/client/utils/Intercom.tsx
```diff
@@ -47,3 +50,67 @@ export const Intercom: React.FC = () => {
 
   return null
 }
+
+export const IntercomHideOnScroll: React.FC = () => {
+  const intercomTimeout = useRef<null | NodeJS.Timeout>()
+  const paused = useRef<boolean>(false)
```
simonauner, line 56: Does `<boolean>` need to be specified when `false` is passed? I believe the type will be inferred anyway.

guilhermespopolin (commented):
Nice initiative

Although I'd try to use their official API ([hide](https://developers.intercom.com/installing-intercom/docs/intercom-javascript#intercomhide) / [show](https://developers.intercom.com/installing-intercom/docs/intercom-javascript#intercomshow)) or a third party like [react-use-intercom](https://www.npmjs.com/package/react-use-intercom).

It might be more work now but maybe a better solution in the long run

robinandeer:
Let's invest more in Racoon - maybe it's simpler to just disable it on mobile and show a link to open it like what @dendoesit wrote :)

dendoesit:
Can we have this as part of the CheckoutPageWrapper ?
Connect payment at sign has 2 pages : Checkout Details and Checkout Payment which should both have the same intercom logic so the best place would be to put it inside the CheckoutPageWrapper .

marian-buc:
@dendoesit I have decided to add the component to both pages as they have slight differences I believe this is better than using it in CheckoutPageWrapper. I've also modified the marginBottom on the wrappers to match both pages nicely. Lmk what you think

src/client/pages/Checkout/CheckoutDetails/CheckoutDetails.tsx
```diff
@@ -1,12 +1,14 @@
 import React from 'react'
-import { useTextKeys } from 'utils/textKeys'
-import { useQuoteCartIdFromUrl } from 'utils/hooks/useQuoteCartIdFromUrl'
+
 import { useCurrentLocale } from 'l10n/useCurrentLocale'
 import { useQuoteCartData } from 'utils/hooks/useQuoteCartData'
+import { useQuoteCartIdFromUrl } from 'utils/hooks/useQuoteCartIdFromUrl'
+import { useTextKeys } from 'utils/textKeys'
```
dendoesit, line 6: we don't need to import these ones right ? 
import { useQuoteCartIdFromUrl } from 'utils/hooks/useQuoteCartIdFromUrl'

marian-buc: they are used within the file

dendoesit (approved):
looks good 👍

PR #890, opened 2022-03-29
bug(GRW-952): details modal save

guilhermespopolin:
## What?

Fix a bug where Details Modal wasn't triggering a API request when "Save" button is clicked;

**Ticket(s): [GRW-952](https://hedvig.atlassian.net/browse/GRW-952)**

## Screenshots / recordings 

https://user-images.githubusercontent.com/19200662/160577492-85ea90d2-ee20-47c8-a756-ab5322341270.mov

### DIFF
--- a/src/client/components/ModalNew.tsx
+++ b/src/client/components/ModalNew.tsx
@@ -41,7 +41,7 @@ type ModalContainerProps = {
 }
 
 const ModalContainer = styled(motion.div)<ModalContainerProps>`
-  position: relative;
+  position: absolute;
   left: 50%;
   top: 50%;
   width: 100%;
@@ -55,7 +55,6 @@ const ModalContainer = styled(motion.div)<ModalContainerProps>`
   max-height: 56rem;`}
   background: ${colorsV3.gray100};
   border-radius: 8px;
-  position: absolute;
   transform: translateX(-50%) translateY(-50%);
   box-shadow: 0 0 14px rgba(0, 0, 0, 0.06);
   overflow-y: auto;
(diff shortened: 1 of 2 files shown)


src/client/components/ModalNew.tsx
```diff
@@ -171,7 +170,7 @@ const Footer = styled.div<ModalFooterProps>`
   align-items: center;
   justify-content: center;
   ${({ sticky }) =>
-    !sticky &&
+    sticky &&
```
robinandeer, line 173: haha just wondering why this changes?

robinandeer (approved):
Makes sense to me -- a little strange that <Formik> doesn't render a <form> element on it's own tho 😅

src/client/components/ModalNew.tsx
guilhermespopolin: It's just that before `position: sticky` (and "associate" styles) were being applied when `sticky` prop was `false` which is misleading is my opinion: you want to add the styles when "you ask" for them - `sticky` being `true`.

It was working before because `sticky` prop wasn't being provided to <Footer /> component: `!undefined && ...`. As you can see [here](https://github.com/HedvigInsurance/web-onboarding/pull/890/files#diff-e9fa83e568fe0a1210f47c062d4d43bb35e20fdcef8ac7f7d0ef6978fa685207R183)

simonauner: I forgot to pass `sticky` to the styled element, so `sticky` was `falsy`. Now that @guilhermespopolin forwards `sticky` the condition must be switched

PR #891, opened 2022-03-29
removed amount object in apple pay

FredrikAreschoug:
## What?

Small change after talking to adyen support

## Why?

Let's try to get apple pay out soon

**Ticket(s): []**


dendoesit (approved):
👍

PR #893, opened 2022-03-29
Apple pay: Amount and currencyCode instead of value and currecy

FredrikAreschoug:
## What?

Amount and currencyCode instead of value and currecy

## Why?

After more support help

**Ticket(s): []**


guilhermespopolin (approved):
LGTM

robinandeer (approved):

PR #895, opened 2022-03-31
bumped to adyen version 5.12.0

FredrikAreschoug:
## What?

Bumped to adyen drop in version 5.12.0

## Why?

We are trying to bump the version form checkout version 65 to 68 in the backend hopefully this will fix the redirect problems we found when we did this last time we tried to do this.

**Ticket(s): []**


FredrikAreschoug:
Yes maybe I should have put this as a draft I'm testing the redirects for this. But currently it is not working at all

PR #896, opened 2022-04-01
Draft / added ExternalInsuranceData  on trackOffer

marian-buc:
## What?
I've added the current_insurance_price data on the track offer function using the apollo cache.

## Why?
To get detailed metrics for every trackOffer() call without query-ing the gql api everytime

**Ticket(s): [GRW-627]**

[GRW-627]: https://hedvig.atlassian.net/browse/GRW-627?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/utils/tracking/gtm.ts
```diff
@@ -132,6 +133,10 @@ export const trackOfferGTM = (
   const grossPrice = Math.round(Number(offerData.cost.monthlyGross.amount))
   const netPrice = Math.round(Number(offerData.cost.monthlyNet.amount))
 
+  const currentInsurer = offerData.quotes[0].currentInsurer?.id ?? undefined
+  const externalInsuranceData = getExternalInsuranceDataFromGQLCache(
+    offerData.quotes[0]?.dataCollectionId,
+  )
```
robinandeer, line 139: We should use our selectors for this - or is it not possible in this case?

src/client/utils/tracking/tracking.ts
```diff
@@ -162,6 +165,31 @@ export const setInitialOfferToSessionStorage = (
   sessionStorage.setItem('initial_offer', contractCategory)
 }
 
+export const getExternalInsuranceDataFromGQLCache = (
+  dataCollectionId?: string | null,
+) => {
```
robinandeer, line 170: Why would this function get called if `dataCollectionId === null`?

robinandeer (commented):
I like this approach, we should not complicate the client code just to enable more tracking data!

Just left a few comments :)

src/client/utils/tracking/gtm.ts
marian-buc: i've went ahead and made a new one

robinandeer (approved):

PR #903, opened 2022-04-05
Feature/Submit-Contact-Information

dendoesit:
GRW-967

## What?

Added option to submit data inside Connect payment at sign 
Add formik and submit function

image

### DIFF
--- a/src/client/pages/Checkout/CheckoutPayment/ContactInformation/ContactInformation.stories.tsx
+++ b/src/client/pages/Checkout/CheckoutPayment/ContactInformation/ContactInformation.stories.tsx
@@ -2,6 +2,7 @@ import React from 'react'
 import { Meta, Story } from '@storybook/react'
 import { MemoryRouter, Route, RouterProps } from 'react-router-dom'
 import styled from '@emotion/styled'
+import { FormikProps } from 'formik'
 import { TextKeyProvider } from 'utils/textKeys'
 import { LocaleLabel } from 'l10n/locales'
 
@@ -11,12 +12,13 @@ import {
 } from 'utils/storybook/storyHelpers'
 import { localePathPattern } from 'l10n/localePathPattern'
 import { mockedQuoteCartId } from 'utils/testData/quoteDetailsDataMock'
+import { QuoteInput } from 'src/client/pages/Offer/Introduction/DetailsModal/types'
 import { ContactInfoData } from '../../shared/types'
 import { ContactInformation } from './ContactInformation'
 
 type StoryProps = {
   localePath: LocaleLabel
-  data: ContactInfoData
+  data: any
 }
 
 const Wrapper = styled.div`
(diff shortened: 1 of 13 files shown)


src/client/pages/Checkout/CheckoutPayment/ContactInformation/ContactInformation.stories.tsx
```diff
@@ -11,12 +11,11 @@ import {
 } from 'utils/storybook/storyHelpers'
 import { localePathPattern } from 'l10n/localePathPattern'
 import { mockedQuoteCartId } from 'utils/testData/quoteDetailsDataMock'
-import { ContactInfoData } from '../../shared/types'
 import { ContactInformation } from './ContactInformation'
 
 type StoryProps = {
   localePath: LocaleLabel
-  data: ContactInfoData
+  data: any
```
simonauner, line 18: Isn't this also `FormikProps<QuoteInput>` like you have in the file below?

src/client/pages/Checkout/CheckoutPayment/ContactInformation/ContactInformation.tsx
```diff
@@ -32,89 +41,94 @@ const InputFieldsWrapper = styled.div`
...
+  const [isShowingCreditCheckInfo, setIsShowingCreditCheckInfo] = useState(
+    false,
+  )
+  const {
+    values: { ssn },
+    initialValues,
+    submitForm,
+  } = formikProps
+
+  const debouncedSubmitForm = useCallback(
```
simonauner, line 74: How come this is needed? Doesn't `disabled={isLoading}` on the Footer Button make it impossible to click many times?

simonauner (approved):
Don't know much about the context but code-wise it looks good to me!

src/client/pages/Checkout/CheckoutPayment/ContactInformation/ContactInformation.stories.tsx
dendoesit: yes it is but then formik requires another 10 values or something 😞 and this felt easier instead of giving empty strings. may have to rethink the way this story is created.

src/client/pages/Checkout/CheckoutPayment/ContactInformation/ContactInformation.tsx
mehdielket: There is a useEffect below that submits the form on SSN change. We submit the SSN immediately (without waiting for users to click the Footer Button, because SSN changes can affect the price. So they need to see the new price before they click continue.

mehdielket: The debouncing is to make sure when users type in the SSN, that the form is not submitted many times

mehdielket (approved):
Looks good! I only see some jumpiness on the footer on mobile, if we can fix that it would be great

https://user-images.githubusercontent.com/89857278/161763968-12e2e460-617c-4c01-b57e-0f3ff8b1cb73.mov

dendoesit:
ahh. @mehdielket  will add it as a fix in the next PR on the Checkout Paymnet page.

PR #905, opened 2022-04-07
feat(GRW-955): expose NO onboarding flow

guilhermespopolin:
## What?

Expose `onboarding-NO` embark flow to support selling _Accident Insurance_ in _NO_.

## Why?

As part of _Pentagon Accident NO_ project, we have:

- Merged all Norwegian _Embark_ flows into a single one: `onboarding-NO` (which already covers the selling of _Accident_ insurance)
- Created a new _Landing Page_ for _NO_ which is implemented under `racoon` project. However, instead of moving all _load Embark story_ logic into `racoon` right now we've decided to accomplish _Pentagon Accident NO_ project on it's early stages by redirecting from `racoon` -> `web-onboarding`. But first we need to make the new flow available on `web-onboarding`: That's what this PR is all about. 

image 

**Ticket(s): [GRW-955](https://hedvig.atlassian.net/browse/GRW-955)**


mehdielket (approved):
🚀

PR #908, opened 2022-04-12
GRW-977 / Fix / Fix how cookies are decoded

robinandeer:
## What?

Avoid decoding `undefined` as "undefined" (string).

## Why?

The previous implementation caused the value `undefined` to be decoded and returned as the string "undefined".

**Ticket(s): []**

### DIFF
--- a/src/client/utils/storage/ServerCookieStorage.ts
+++ b/src/client/utils/storage/ServerCookieStorage.ts
@@ -11,9 +11,8 @@ export class ServerCookieStorage implements MinimalStorage {
       return this.changedItems[item]
     }
 
-    return decodeURIComponent(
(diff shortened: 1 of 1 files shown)


src/client/utils/storage/ServerCookieStorage.ts
```diff
@@ -11,9 +11,8 @@ export class ServerCookieStorage implements MinimalStorage {
       return this.changedItems[item]
     }
 
-    return decodeURIComponent(
-      this.requestCtx.cookies.get(item, { signed: false })!,
```
robinandeer, line 15: Interesting to note is the TS would've complained if we hadn't told it not to with the `!`

mehdielket (approved):
🚀

PR #912, opened 2022-04-14
Bugfix/cpas

dendoesit:
## What?

Added loading state to button when start checkout is initiated.
Added option for google pay and VIPS in NO. 

## Why?

Making the connect payment at sign flow better

**Ticket(s): []**

## Screenshots / recordings 
image


robinandeer (dismissed):
Nice!

I wonder tho if we should import `checkIsManualReviewRequired` from the solution that will be removed. Maybe it's better to do it the other way around?

robinandeer (approved):

PR #913, opened 2022-04-15
Bugfix/error handling for CPAS

dendoesit:
## What?

Error state for CPAS pages.

## Why?

We need to handle errors and loading state.


simonauner (dismissed):
I do not know enough about the context to say if the logic "fits", but coding-wise it looks good to me 🤷

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
```diff
@@ -123,6 +123,22 @@ const Terms = styled.div`
...
+}
+
+export const isSsnInvalid = (errors: GraphQLError[]) => {
+  const invalidSsnError = errors.find((error) => {
+    return error?.extensions?.body?.errorCode === LimitCode.INVALID_SSN
+  })
+
+  return invalidSsnError !== undefined
+}
+
```
mehdielket, line 141: These probably should be in some helper file or utils I think.

src/client/pages/Checkout/CheckoutDetails/CheckoutDetails.tsx
```diff
@@ -20,18 +20,14 @@ export const CheckoutDetails = () => {
...
-  }
-  const onRetry = () => {
-    window.location.reload()
-    return false
-  }
 
-  if (data.error) {
-    console.error('Quote cart data error:', data.error.message, data.error)
+  if (data?.error) {
+    console.error('Quote cart data error: no data')
```
mehdielket, line 25: Maybe we should remove the console.log?

mehdielket (commented):
Looks good, just a couple of small comments

src/client/pages/Checkout/CheckoutDetails/CheckoutDetails.tsx
dendoesit: I think it helps to know what exactly failed since we display a really generic error everywhere.

robinandeer (approved):

PR #915, opened 2022-04-19
 update footer button

dendoesit:
## What?

Update old sticky button in offer page

## Why?

Same button as sidebar. CPAS updates.

Before
image

After
image


mehdielket (approved):
🚀 Tested on staging

PR #919, opened 2022-04-20
GRW-990 Fix adyen simple sign

mehdielket:
## What?

Fix issue with simple sign for adyen not working. The root cause is that we were trying to call `startCheckout` mutation with outdated quote ids that did not belong to the cart.

The solution was to not make `startCheckout` part of the success handler passed to Adyen but rather call it when the payment info is added to the cart successfully

**Ticket(s): [GRW-990]**

[GRW-990]: https://hedvig.atlassian.net/browse/GRW-990?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/api/quoteCartQuerySelectors.ts
```diff
@@ -48,3 +48,12 @@ export function getMonthlyCostDeductionIncentive(
     ? incentive
     : undefined
 }
+
+export function isPaymentConnected(quoteCartQuery: QuoteCartQuery | undefined) {
+  const id = quoteCartQuery?.quoteCart.paymentConnection?.id
+  console.log(
```
dendoesit, line 54: this should be removed.

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
```diff
@@ -163,51 +165,61 @@ export const CheckoutPayment = ({
...
-        })
-      } catch (error) {
-        const isManualReviewRequired = checkIsManualReviewRequired(
-          (error.graphQLErrors || []) as GraphQLError[],
-        )
-        if (isManualReviewRequired) {
-          throw new Error('Manual Review required')
-        }
-        setIsLoading(false)
+  const checkout = useCallback(async () => {
```
robinandeer, line 185: I think this variable name is too generic -- can we be more descriptive?

robinandeer (approved):
Looks good, just had a couple of small comments

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
mehdielket: I meant `checkout` as a verb as in "do checkout", since this is calling `startCheckout`. Maybe this is unclear since it can be a noun and a verb. Do you have a suggestion?

robinandeer: Hmmm... maybe just `handleCheckout` or `performCheckout`

dendoesit (dismissed):
🎉

dendoesit (approved):

robinandeer (approved):
Guess this good looks good! Let's test in staging :D

PR #920, opened 2022-04-20
fix: unify scroll lock solution

robinandeer:
## What?

Use the same hook for all scroll lock behavior.
Remove `useLockBodyScroll` hook.

Tested successfully locally.

## Why?

1. We've had issues with scroll lock persisting after you close checkout/modal.
2. We should only have one hook to achieve this behavior.

**Ticket(s): [GRW-957]**

[GRW-957]: https://hedvig.atlassian.net/browse/GRW-957?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dendoesit (approved):
looks good :)

PR #921, opened 2022-04-21
GRW-986 / Disabled form submissions with incomplete fields and added initial validation

marian-buc:
## What?

Added the validateOnMount prop on formik to make sure the form is always validated, and disabled the footer button when there are form errors

## Why?

To prevent users from submitting the form without any empty or incorrect fields

**Ticket(s): []**

Now, after arriving without a ssn:
image

### DIFF
--- a/src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
+++ b/src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
@@ -239,7 +239,7 @@ export const CheckoutPayment = ({
       },
     } as QuoteInput,
     validationSchema: getCheckoutDetailsValidationSchema(locale, textKeys),
-    validateOnMount: true,
+    validateOnChange: false,
     onSubmit: async (
       form: QuoteInput,
       { setErrors }: FormikHelpers<QuoteInput>,
@@ -341,8 +341,12 @@ export const CheckoutPayment = ({
     })
   }
 
-  const startSign = async () => {
-    const { submitForm, dirty: isFormDataUpdated } = formik
+  const handleClickCompletePurchase = async () => {
+    const { validateForm, submitForm, dirty: isFormDataUpdated } = formik
+
+    const errors = await validateForm()
+    if (Object.keys(errors).length > 0) return
+
     if (isFormDataUpdated) {
       const { data } = await submitForm()
       const isUpdateQuotesFailed = isQuoteBundleError(data)
@@ -393,9 +397,7 @@ export const CheckoutPayment = ({
       {mainQuote && (
         <Footer
           buttonText={textKeys.CHECKOUT_FOOTER_COMPLETE_PURCHASE()}
-          buttonOnClick={() => {
-            startSign()
-          }}
+          buttonOnClick={handleClickCompletePurchase}
           isLoading={isBundleCreationInProgress || isDataLoading}
           disabled={isFormikError}
         >
--- a/src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
+++ b/src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
@@ -239,6 +239,7 @@ export const CheckoutPayment = ({
       },
     } as QuoteInput,
     validationSchema: getCheckoutDetailsValidationSchema(locale, textKeys),
+    validateOnMount: true,
     onSubmit: async (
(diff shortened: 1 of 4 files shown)


src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
```diff
@@ -224,6 +224,7 @@ export const CheckoutPayment = ({
       },
     } as QuoteInput,
     validationSchema: getCheckoutDetailsValidationSchema(locale, textKeys),
+    validateOnMount: true,
```
simonauner, line 227: I'm not sure what the use case looks like, but isn't it generally a good idea to allow the user to get the chance to fill in information before showing the error? Now it will look like the user has entered incorrect information or tried to submit the form even before they've had a chance to enter anything? Or am I misinterpreting the flow?

I totally agree that we shouldn't submit when there are errors, but Formik validates on submit, so it won't send any requests if the form isn't valid.

mehdielket: I agree with Simon that we should not show the error right away. That said, the validation on submit is not automatic I believe. We use the following piece of code in the `startSign` function in the old offer page to force validation. I think we should do something similar

```
const errors = await validateForm()

if (Object.keys(errors).length) {
  setSignUiState('FAILED')
  return
}
```

https://github.com/HedvigInsurance/web-onboarding/blob/62d429c13430dafe0a8616e5b6739bf715689e87/src/client/pages/Offer/Checkout/index.tsx#L392

simonauner: Perhaps we're not using Formik's submit then? It _should_ validate: https://formik.org/docs/guides/validation#when-does-validation-run

robinandeer: ☝️ yeah let's look into why that's not working out of the box

robinandeer:
Built on the initial solution but changes a few things:

- Validate the form when user clicks "complete purchase" instead of on mount.
- Refactor automatic form submission to only happen after user has stopped typing

https://user-images.githubusercontent.com/1220232/164491806-4b51ca46-8b0a-4945-a63b-7a61cc9f47ca.mov

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
robinandeer: @simonauner the "complete purchase" button didn't belong to that form so that's why it wasn't validating the inputs

simonauner: Ah, good find!

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
```diff
@@ -239,6 +239,7 @@ export const CheckoutPayment = ({
       },
     } as QuoteInput,
     validationSchema: getCheckoutDetailsValidationSchema(locale, textKeys),
+    validateOnChange: false,
```
simonauner, line 242: It makes a lot more sense with the new solution! How come we do not want to validate on change though?

If a user submits the form without an SSN, the error will show on the SSN field. And when the user has entered a valid SSN the error will still show until they press the submit button again if we do not validate on change?

robinandeer: `validateOnChange` causes an error to show up as soon as the user starts typing into the input field 🙄 I just think that's annoying

For the SSN-field we actually submit when the user stops typing so that error will go away automatically :)

simonauner: Ah, yes, I agree that's annoying. It's possible to "configure" Formik so that error messages only show when the field is blurred and dirty, but of course that's out of scope for this! 👌

simonauner (approved):

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
robinandeer: Aha yeah that would make more sense.

For now I thought I'd go with standard web behavior (validate on submit) and then let's see if we need something more fancy ;)

PR #925, opened 2022-04-21
Minor/fix vips

dendoesit:
https://hedvig.atlassian.net/browse/GRW-989

## What?

Fix Vipps implementation in CPAS and add apple pay option.

## Why?

FYI: It is hard for me to test the apple pay implementation so not 100% sure it works. 

image


src/client/l10n/locales.ts
```diff
@@ -46,6 +46,7 @@ export type LocaleData = {
   }
   callCenter?: CallCenterData
   currencyCode: 'SEK' | 'NOK' | 'DKK'
+  market: 'SWEDEN' | 'NORWAY' | 'DENMARK'
```
robinandeer, line 49: This look identical to `apiMarket` - is there a difference?

robinandeer (commented):
Nice stuff - guess we can test apple pay in staging

src/client/l10n/locales.ts
dendoesit: no. I just did not notice we have that. will revert.

PR #928, opened 2022-04-22
Add CI workflow using GitHub Actions

robinandeer:
## What?

See title.

## Why?

We are making a push to move off CodeFresh so this aims to replace it and use GitHub Actions for our CI pipeline.

See: https://hedviginsurance.slack.com/archives/C9K7XRQ0G/p1650626436696869

**Ticket(s): []**

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,9 +39,3 @@ jobs:
 
       - name: Test
         run: yarn test
-
-      - name: Build client
-        run: yarn build-client
-
-      - name: Build server
-        run: yarn build-server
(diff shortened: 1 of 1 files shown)


simonauner (dismissed):

robinandeer:
Not sure we need to run the build scripts in the CI pipeline -- they are taking an awful long time to complete 🙄 

What do you say?

simonauner:
> Not sure we need to run the build scripts in the CI pipeline -- they are taking an awful long time to complete 🙄
> 
> What do you say?

Does it add any value? I feel like ts check, lint and testing is enough 🤔

robinandeer:
I mean it could be possible to mess up the `build` while everything else passes. But that means we still won't be able to deploy anything.

Let's remove it from here :)

PR #934, opened 2022-04-26
Feat / GRW-984/ handle 3Ds error redirect

marian-buc:
## What?
This PR adds handling for failed 3Dsecure flows when redirected back to checkout/payment. Changed to use history API in order to cleanup the URL params before reload .

## Why?
To let the user know that his payment attempt failed and he can try again.

- Should we use a different modal?

**Ticket(s): [GRW-984]**

https://user-images.githubusercontent.com/101566411/165206725-329e5f68-fe1e-4aa6-9bfe-c0cbf5b2b810.mp4

![

[GRW-984]: https://hedvig.atlassian.net/browse/GRW-984?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dendoesit (approved):
looks good.

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
```diff
@@ -163,11 +163,25 @@ export const CheckoutPayment = ({
   const [getStatus] = useCheckoutStatusLazyQuery({
     pollInterval: 1000,
   })
-  const { search: is3DsComplete } = useLocation<{ search: string }>()
+  const history = useHistory()
```
robinandeer, line 166: Why not still use `useLocation`?

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
```diff
@@ -163,11 +163,25 @@ export const CheckoutPayment = ({
...
+
   const [isDataLoading, setIsDataLoading] = useState(false)
   const [isPageLoading, setIsPageLoading] = useState(false)
   const [isError, setIsError] = useState(false)
 
+  //handle 3ds error redirect
+  useEffect(() => {
+    if (is3DsError) {
+      history.replace('?')
+      setIsError(true)
```
robinandeer, line 181: I'm not sure if it's a good idea to reuse this state.

Could we just use `is3DsError` instead?

robinandeer (commented):
Good work, I just wondered if we really should use such a big modal takeover?

Would be nice if we can handle it in a more subtle way, what do you say? Or have you discussed with amy/zak?

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
guilhermespopolin: I guess he needs `history` so he can replace `error` search param below.

robinandeer: Aha I see, but I meant for accessing search params. That seems a bit more complicated now, no?

```
const {
    location: { search },
  } = history
```

But just a nit

el-buku:
What do you suggest? I could add a smaller modal without the backdrop, I'll ping Zak

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
marian-buc: yeah, I think it's better performance to destructure that object rather than using 2 hooks

robinandeer (approved):
On second thought I think we can merge this as a step-improvement

But let's check in with Zak later for a more subtle solution :)

marian-buc:
Will merge this and add the new modal in a future PR
https://www.figma.com/file/GQ1b2GBtO3SxGW2YaH7qng/?node-id=2646%3A10018

PR #936, opened 2022-04-26
Bugfix/vipps redirect

dendoesit:
## What?

Fix for Vipps

## Why?

Correct redirect to download page instead of offer

**Ticket(s): []**
https://hedvig.atlassian.net/browse/GRW-1017


robinandeer (approved):
Is this all we need to do? 😅 

Nice

dendoesit:
> Is this all we need to do? 😅
> 
> Nice

pretty much. I spoke with Fredrik and Carl about it and apparently we do not need to do anything else on FE.

PR #938, opened 2022-04-26
Fix / Add border to images to increase contrast

robinandeer:
## What?
 
localhost_8040_se-en_new-member_download(iPad Air)

## Why?

**Ticket(s): []**


gustaveen:
Cool! Could a drop shadow on the images itself be a solution as wll? Or did Amy or Zak suggest this? :)

robinandeer:
@gustaveen the screenshots are grouped together so they would have to be re-exported in that case :P

robinandeer:
Zak will provide new images instead

PR #940, opened 2022-04-27
useTrackOfferEvent hook

marian-buc:
## What?
Added useTrackOfferEvent hook for checkout tracking, using an eventQueue state array to defer event tracking until the data is loaded. I've also replaced the tracking function calls in https://github.com/HedvigInsurance/web-onboarding/pull/923 with the hook.

## Why?
In order to avoid fetching the gql data inside each element where we want to use trackOfferEvent() and call each event atomically.

**Ticket(s): [GRW-1016, GRW-631,GRW-758]**


src/client/pages/Checkout/CheckoutDetails/CheckoutDetails.tsx
```diff
@@ -28,42 +22,16 @@ import { DocumentLinks } from './components/DocumentLinks'
...
-        {
-          quoteCartId,
-        },
-      )
-    }
-  }, [isReferralCodeUsed, quoteCartId, selectedBundleVariant])
+  const trackOfferEvent = useTrackOfferEvent()
+  const data = useQuoteCartData()
 
+  useEffect(() => trackOfferEvent({ eventName: EventName.CheckoutOpen }), [])
```
robinandeer, line 34:
```suggestion
  useEffect(() => trackOfferEvent({ eventName: EventName.CheckoutOpen }), [trackOfferEvent])
```

src/client/pages/Checkout/CheckoutPayment/ContactInformation/ContactInformation.tsx
```diff
@@ -59,39 +53,12 @@ type Props = {
...
 
-  const isReferralCodeUsed =
-    getMonthlyCostDeductionIncentive(quoteCartQueryData) !== undefined
-
-  const selectedBundleVariant = getSelectedBundleVariant(
-    quoteCartQueryData,
-    selectedInsuranceTypes,
+  useEffect(
+    () => trackOfferEvent({ eventName: EventName.ContactInformationPageOpen }),
+    [],
```
robinandeer, line 60:
```suggestion
    [trackOfferEvent],
```

src/client/utils/tracking/trackOfferEvent.ts
```diff
@@ -69,3 +83,62 @@ export const trackOfferEvent = (
...
+  )
+
+  const trackOfferHandler = useCallback(
+    (eventParams: EventParameters) => {
+      selectedBundleVariant
+        ? trackOfferCallback(eventParams)
+        : setEventQueue(eventQueue.concat([eventParams]))
+    },
+    [selectedBundleVariant, eventQueue, trackOfferCallback],
+  )
```
robinandeer, line 133: Can we do this instead? Then it will only do one thing - add a new event to be tracked to the queue.

It also means we get a stable reference so it won't cause `useEffect` calls to rerun.

```suggestion
  const trackOfferHandler = useCallback(
    (eventParams: EventParameters) => {
      setEventQueue((prevQueue) => [...prevQueue, eventParams])
    },
    [],
  )
```

amandaflod (approved):
Great work! 👍 Will make my PR into a draft again and await these changed so I can implement them directly 🥳

src/client/utils/tracking/trackOfferEvent.ts
```diff
@@ -69,3 +83,62 @@ export const trackOfferEvent = (
...
+
+  //cleanup the queue
+  useEffect(() => {
+    if (selectedBundleVariant && eventQueue.length) {
+      eventQueue.map(trackOfferCallback)
+      setEventQueue([])
+    }
```
robinandeer, line 140: I would avoid map here since we don't care about the return. 

```suggestion
    if (selectedBundleVariant) {
      const event = eventQueue[0]
      if (event) {
        trackOfferCallback(event, selectedBundleVariant)
        setEventQueue((prevQueue) => [...prevQueue].splice(1))
      }
    }
```

simonauner (approved):
I like the solution 🙌 

Github actions added some comments about warnings that would need to be fixed but other than that two thumbs up

src/client/utils/tracking/trackOfferEvent.ts
```diff
@@ -69,3 +83,62 @@ export const trackOfferEvent = (
...
+    selectedInsuranceTypes,
+  )
+
+  const [eventQueue, setEventQueue] = useState<EventParameters[]>([])
+
+  const trackOfferCallback = useCallback(
+    ({ eventName, options = {} }: EventParameters) => {
+      trackOfferEvent(
+        eventName,
+        selectedBundleVariant!.bundle,
```
robinandeer, line 115: Let's pass this into the function instead of using this escape hatch :)

robinandeer (commented):
Fantastic approach -- really like it :D 

I had some thought on simplifying the logic in the hook tho - let me know if you wanna go over it together @marian-buc?

src/client/utils/tracking/trackOfferEvent.ts
el-buku: Right, this was my first thought as well, but figured I'd go with a quick map rather than an effect for each call. It is cleaner like this though, since all events will go through the queue and the callback is only called in one place

marian-buc: Right, this was my first thought as well, but figured I'd go with a quick map rather than an effect for each call. It is cleaner like this though, since all events will go through the queue and the callback is only called in one place

robinandeer: Yes exactly, better to keep each function doing one thing :)

robinandeer (approved):
So clean! Awesome work - haven't tested but looks good

PR #941, opened 2022-04-27
GRW-983 / Fix / Handle error in insurance summary details

robinandeer:
## What?

Handle uncaught error, show generic error message, and send error to sentry.

Screenshot 2022-04-27 at 13 51 06

## Why?

We noticed this error occurs if backend sends back a birth date from the 1800s.
Therefore I think it's best we still throw the exception but catch it and log it.

**Ticket(s): []**


amandaflod (approved):
Great stuff 👍 Didn't know about the ErrorBoundary from Sentry before!


https://en.wikipedia.org/wiki/Emma_Morano 👵

PR #944, opened 2022-04-28
feat/GRW-1016/CPAS error tracking

marian-buc:
## What?
Added error tracking using the new hook on CheckoutPayment as well as useAdyenCheckout

## Why?
To track checkout errors on Google Analytics besides Datadog
Thoughts:

- Should we have a separate trackErrorEvent() and have the hook call the correct function?
- What data needs to be sent along with each error? Do we need the selectedBundle data like in trackOfferEvent()?
- error optional parameter typedef? 

**Ticket(s): [GRW-1016]**

[GRW-1016]: https://hedvig.atlassian.net/browse/GRW-1016?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/utils/tracking/gtm.ts
+++ b/src/client/utils/tracking/gtm.ts
@@ -107,7 +107,7 @@ export const pushToGTMDataLayer = (obj: DataLayerObject) => {
   castedWindow.dataLayer.push(obj)
 }
 
-export enum EventName {
+export enum OfferEventName {
   OfferCreated = 'offer_created',
   SignedCustomer = 'signed_customer',
   InsuranceSelectionToggle = 'insurance_selection_toggle',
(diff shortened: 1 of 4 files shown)


src/client/utils/tracking/gtm.ts
```diff
@@ -117,6 +117,20 @@ export enum EventName {
   ContactInformationPageOpen = 'contact_information_page_open',
 }
 
+export enum ErrorEventName {
+  CheckoutErrorPaymentTokenMutation = 'checkout_error_payment_token_mutation',
+  CheckoutErrorPaymentTokenID = 'checkout_error_payment_token_id_missing',
```
robinandeer, line 122:
```suggestion
  CheckoutErrorPaymentTokenIDMissing = 'checkout_error_payment_token_id_missing',
```

src/client/utils/tracking/trackOfferEvent.ts
```diff
@@ -28,6 +28,8 @@ type OptionalParameters = {
   phoneNumberData?: GTMPhoneNumberData
   quoteCartId?: string
   memberId?: string
+  // TODO Better typedef?
+  error?: Error | any
```
robinandeer, line 32: why is any needed?

robinandeer (approved):
I think this looks great - just some small comments!

src/client/utils/tracking/trackOfferEvent.ts
marian-buc: well, Error is a generic interface and catch blocks return (error: unknown) objects which are different depending on what's called inside, so I figured it's a good measure for usability. Looking forward to a better solution

robinandeer: I would prefer `Error | unknown` if simply `Error` doesn't work -- to better reflect what this actually us :)

robinandeer (dismissed):
Consider narrowing the type - otherwise looks good!

src/client/utils/tracking/trackOfferEvent.ts
marian-buc: cool, makes sense

dendoesit (approved):
⭐

PR #945, opened 2022-04-28
GRW-1001 / Feature / Adapt offer page to show car insurance

simonauner:
## What?

Adapt offer page to be able to show insurances that are not house/apartment

This PR mainly makes sure that the page doesn't crash when a non-house/apartment quote is provided, so that we can fan out the remaining tasks.

* Use contract type to determine which quote bundle that is selected, and what is shown in the URL. This is because all car quote variations have the same quote type, but not same contract type. This also happens to work for other bundles.
* Adapt SummaryDetails to only show information that's available with the current data.

Do we need some regression testing on this before I ship it to production? I have tested to render the offer page with all variations that exist in Sweden and it looks good.

## Why?

Since we need to sell car insurance in web-onboarding we need to be able to show it in the offer screen. 

**Ticket(s): [GRW-1001]**

House + accident SE local (just to show that it still works 😁 ):
image

Rendering a car quote cart:

image

[GRW-1001]: https://hedvig.atlassian.net/browse/GRW-1001?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/pages/Offer/Introduction/HeroOfferDetails.tsx
+++ b/src/client/pages/Offer/Introduction/HeroOfferDetails.tsx
@@ -121,7 +121,7 @@ export const HeroOfferDetails: React.FC<Props> = ({
     return query.get('showEdit') === 'true'
   })
   const { person } = offerData
-  const numberCoInsured = person.householdSize - 1
+  const numberCoInsured = (person.householdSize ?? 0) - 1
 
   const address = getAddress(allQuotes)
 
--- a/src/client/pages/Offer/Introduction/HeroOfferDetails.tsx
+++ b/src/client/pages/Offer/Introduction/HeroOfferDetails.tsx
@@ -121,7 +121,7 @@ export const HeroOfferDetails: React.FC<Props> = ({
     return query.get('showEdit') === 'true'
   })
   const { person } = offerData
-  const numberCoInsured = (person.householdSize ?? 0) - 1
+  const numberCoInsured = (person.householdSize ?? 1) - 1
 
   const address = getAddress(allQuotes)
 
--- a/src/client/pages/OfferNew/utils.ts
+++ b/src/client/pages/OfferNew/utils.ts
@@ -25,6 +25,8 @@ import { InsuranceType } from 'utils/hooks/useSelectedInsuranceTypes'
 
 export const getOfferData = (quoteBundle: QuoteBundle): OfferData => {
   const firstQuote = quoteBundle.quotes[0]
+  const householdSize = getHouseholdSizeFromBundledQuotes(quoteBundle.quotes)
+
   return {
     person: {
       firstName: firstQuote.firstName,
@@ -33,7 +35,7 @@ export const getOfferData = (quoteBundle: QuoteBundle): OfferData => {
       ssn: firstQuote.ssn,
       phoneNumber: firstQuote.phoneNumber,
       birthDate: firstQuote.birthDate,
-      householdSize: getHouseholdSizeFromBundledQuotes(quoteBundle.quotes),
+      ...{ householdSize },
       address: getAddressFromBundledQuotes(quoteBundle.quotes),
     },
     quotes: quoteBundle.quotes.map((bundleQuote) => {
@@ -59,7 +61,9 @@ export const isOfferDataAvailable = (
   return offerData !== null
 }
 
-const getHouseholdSizeFromBundledQuotes = (quotes: BundledQuote[]): number => {
+const getHouseholdSizeFromBundledQuotes = (
+  quotes: BundledQuote[],
+): number | undefined => {
   const quoteDetails = quotes.map((quote) => quote.quoteDetails)
 
   for (const details of quoteDetails) {
@@ -67,11 +71,7 @@ const getHouseholdSizeFromBundledQuotes = (quotes: BundledQuote[]): number => {
     if ('coInsured' in details) return details.coInsured + 1
   }
 
-  throw new Error(
-    `quoteDetails ${JSON.stringify(
-      quoteDetails,
-    )} must include one of the following: "householdSize" or "coInsured".`,
-  )
+  return
 }
 
 export const getHouseholdSize = (quoteDetails: QuoteDetails): number => {
@@ -534,7 +534,7 @@ const isBundleVariantMatchingInsuranceTypes = (
   variant: QuoteBundleVariant,
   insuranceTypes: Array<InsuranceType>,
 ) => {
-  const variantInsuranceTypes = getInsuranceTypesFromBundleVariant(variant)
+  const variantInsuranceTypes = getTypeOfContractFromBundleVariant(variant)
   return (
     variantInsuranceTypes.sort().join(',') ===
     insuranceTypes
@@ -552,23 +552,18 @@ export const getBundleVariantFromInsuranceTypesWithFallback = (
     isBundleVariantMatchingInsuranceTypes(variant, insuranceTypes),
   )
 
-  if (!matchingVariant) {
-    // Fallback to the variant with the fewest quotes
-    return [...variants].sort(
-      (a, b) => a.bundle.quotes.length - b.bundle.quotes.length,
-    )[0]
-  }
+  if (matchingVariant) return matchingVariant
 
-  return matchingVariant
+  // Fallback to the variant with the fewest quotes
+  return [...variants].sort(
+    (a, b) => a.bundle.quotes.length - b.bundle.quotes.length,
+  )[0]
 }
-
-export const getInsuranceTypesFromBundleVariant = (
-  bundleVariant: QuoteBundleVariant,
-) =>
-  bundleVariant.bundle.quotes.map<InsuranceType>(
-    (quote) => quote.data.type as InsuranceType,
-  )
 
 export const getQuoteIdsFromBundleVariant = (
   bundleVariant: QuoteBundleVariant,
 ) => bundleVariant.bundle.quotes.map((quote) => quote.id)
+
+export const getTypeOfContractFromBundleVariant = (
+  bundleVariant: QuoteBundleVariant,
+) => bundleVariant.bundle.quotes.map((quote) => quote.typeOfContract)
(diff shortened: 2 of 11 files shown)


src/client/pages/OfferNew/utils.ts
```diff
@@ -59,19 +61,17 @@ export const isOfferDataAvailable = (
...
     if ('householdSize' in details) return details.householdSize
     if ('coInsured' in details) return details.coInsured + 1
   }
 
-  throw new Error(
-    `quoteDetails ${JSON.stringify(
-      quoteDetails,
-    )} must include one of the following: "householdSize" or "coInsured".`,
-  )
+  return
```
simonauner, line 74: Can't throw here since neither `householdSize` nor `coinsured` exists on a car insurance. Sure, we could check before calling `getHouseholdSizeFromBundledQuotes` if any of those two exists, but then there's no need to throw either since one of them will exist every time we call the function.

src/client/pages/Offer/Introduction/HeroOfferDetails.tsx
```diff
@@ -121,7 +121,7 @@ export const HeroOfferDetails: React.FC<Props> = ({
     return query.get('showEdit') === 'true'
   })
   const { person } = offerData
-  const numberCoInsured = person.householdSize - 1
+  const numberCoInsured = (person.householdSize ?? 0) - 1
```
robinandeer, line 124: Wait... if we display this, won't it show up as `-1` co-insured now?

src/client/pages/OfferNew/utils.ts
```diff
@@ -33,7 +35,7 @@ export const getOfferData = (quoteBundle: QuoteBundle): OfferData => {
       ssn: firstQuote.ssn,
       phoneNumber: firstQuote.phoneNumber,
       birthDate: firstQuote.birthDate,
-      householdSize: getHouseholdSizeFromBundledQuotes(quoteBundle.quotes),
+      ...{ householdSize },
```
robinandeer, line 38: just curious why this object destructing is needed?

src/client/utils/hooks/useSelectedInsuranceTypes.ts
```diff
@@ -13,7 +14,7 @@ export enum InsuranceType {
   DANISH_TRAVEL = 'DANISH_TRAVEL',
 }
 
-const ALL_INSURANCE_TYPES = Object.values(InsuranceType)
+const ALL_INSURANCE_TYPES = Object.values(TypeOfContract)
```
robinandeer, line 17: How will this work? Because we send in insurance type still.

If we use "type of contract" we should probably rename the hook as well, no?

robinandeer (commented):
Great progress :)

src/client/pages/Offer/Introduction/HeroOfferDetails.tsx
simonauner: Yeah, it's not easy to spot, but on row 136 we do a conditional render if `numberCoInsured > 0`

src/client/utils/hooks/useSelectedInsuranceTypes.ts
simonauner: It works because `setSelectedInsuranceTypes` is now called with a `TypeOfContract`, see `src/client/pages/Offer/index.tsx`. But yes, I agree on the name. I wanted to inflict as small a change as possible, but I will look into renaming it 👍

simonauner: Ok, seeing as this could inflict a lot of changes in other parts of the web, I will create a new hook `useSelectedContractTypes` instead and use that in the checkout.

src/client/pages/OfferNew/utils.ts
simonauner: Seems to work totally fine as it was, probably some refactoring that got stuck =) Will revert!

src/client/utils/hooks/useSelectedInsuranceTypes.ts
simonauner: Touched a lot of files... https://github.com/HedvigInsurance/web-onboarding/pull/945/commits/5a1391fbc6be7bba174db9e77ae83bf6815ae2ae

src/client/pages/Checkout/shared/types.ts
```diff
@@ -1,5 +1,16 @@
 import { ExtraBuildingInput } from 'data/graphql'
-import { InsuranceType } from 'utils/hooks/useSelectedInsuranceTypes'
+
+// TODO: we should get this from giraffe schema
```
simonauner, line 3: This was the only thing left in `useSelectedInsuranceTypes.ts` and it didn't make sense to keep that unused hook file just for some type definitions so I moved them here. Not sure if there's a better place to put them?

robinandeer (commented):
haha I think I still don't get it 😅 

how will this work with the existing Embark-flows that select bundle based on "insurance type" instead of "type of insurance"?

What I mean is that this solution doesn't seems backwards compatible, or am I missing something?

simonauner:
> haha I think I still don't get it 😅
> 
> how will this work with the existing Embark-flows that select bundle based on "insurance type" instead of "type of insurance"?
> 
> What I mean is that this solution doesn't seems backwards compatible, or am I missing something?

Ok, yes, that is something I know nothing about obviously :) I thought this was an isolated solution in the offer page and checkout.

Then I'm not really sure how we're gonna get this to work, since the select from Car needs to be made on ContractType, not TypeOfInsurance.

robinandeer:
> > haha I think I still don't get it 😅
> > how will this work with the existing Embark-flows that select bundle based on "insurance type" instead of "type of insurance"?
> > What I mean is that this solution doesn't seems backwards compatible, or am I missing something?
> 
> Ok, yes, that is something I know nothing about obviously :) I thought this was an isolated solution in the offer page and checkout.
> 
> Then I'm not really sure how we're gonna get this to work, since the select from Car needs to be made on ContractType, not TypeOfInsurance.

Right, I think we should aim to keep this backwards compatible - meaning that existing Embark flows don't need to change.

My initial thought was to update `getSelectedBundleVariant` so that it first tries to find the bundle based on "type of contract" and if it comes up with nothing, falls back to the existing behaviour by matching with "insurance type" => do you think that could work?

simonauner:
> > > haha I think I still don't get it 😅
> > > how will this work with the existing Embark-flows that select bundle based on "insurance type" instead of "type of insurance"?
> > > What I mean is that this solution doesn't seems backwards compatible, or am I missing something?
> > 
> > 
> > Ok, yes, that is something I know nothing about obviously :) I thought this was an isolated solution in the offer page and checkout.
> > Then I'm not really sure how we're gonna get this to work, since the select from Car needs to be made on ContractType, not TypeOfInsurance.
> 
> Right, I think we should aim to keep this backwards compatible - meaning that existing Embark flows don't need to change.
> 
> My initial thought was to update `getSelectedBundleVariant` so that it first tries to find the bundle based on "type of contract" and if it comes up with nothing, falls back to the existing behaviour by matching with "insurance type" => do you think that could work?

Yeah, I remember you said that previously, and I tried to get that to work.. But also the select-action on the cards need to change to put the ContractType in the URL, not InsuranceType, and that's what breaks that solution. But I can revisit this today and see if I can make it work. Maybe the cards can put the ContractType in the URL but Embark can put InsuranceType. Then it's perhaps just an issue with TypeScript, and that can be solved.

simonauner:
New solution!

* When selecting cards, if it's a car insurance, set `type` in URL to its `TypeOfContract` (yes, not very scalable, but we only have the one instance of a quote so we can't determine if they all share the same `InsuranceType`)
* `getBundleVariantFromInsuranceTypesWithFallback` now first tries to find bundle variant on `TypeOfContract` if all possible bundles share the same `InsuranceType`. If they do not share the same `InsuranceType` it will use `InsuranceType` to determine bundle variant.
* `useSelectedInsuranceTypes` now uses both `InsuranceType` and `TypeOfContract`

src/client/pages/Offer/Introduction/HeroOfferDetails.tsx
robinandeer: Would still consider changing to:

```suggestion
  const numberCoInsured = (person.householdSize ?? 1) - 1
```

...or does that not work? Just since "-1" co-insured or household size "0" simply is not valid.

src/client/pages/OfferNew/utils.ts
```diff
@@ -544,3 +571,25 @@ export const getInsuranceTypesFromBundleVariant = (
 export const getQuoteIdsFromBundleVariant = (
   bundleVariant: QuoteBundleVariant,
 ) => bundleVariant.bundle.quotes.map((quote) => quote.id)
+
+export const getTypeOfContractFromBundleVariant = (
+  bundleVariant: QuoteBundleVariant,
+) => bundleVariant.bundle.quotes.map((quote) => quote.typeOfContract)
+
+function allBundleVariantsHaveSameInsuranceType(
```
robinandeer, line 579: nit: consider using a function expression for consistency

src/client/pages/OfferNew/utils.ts
```diff
@@ -544,3 +571,25 @@ export const getInsuranceTypesFromBundleVariant = (
...
+  bundleVariants: Array<QuoteBundleVariant>,
+) {
+  if (bundleVariants.length <= 1) {
+    return false
+  }
+
+  const insuranceType = bundleVariants?.[0].bundle.quotes?.[0]?.data.type
```
robinandeer, line 586: guess we don't have a selector for this? 😅

robinandeer (commented):
Nicely solved, certainly not scalable but not sure if this is the time to dig in a really solve it either 🙃

simonauner:
> Nicely solved, certainly not scalable but not sure if this is the time to dig in a really solve it either 🙃

Thanks! Well, probably not, right? Let's call it a day and save any further ideas to the next solution in Racoon?

src/client/pages/OfferNew/utils.ts
simonauner: Yeah, thanks. This is what happens when I tell VSCode to create missing function declarations =)

src/client/pages/OfferNew/utils.ts
simonauner: Nope, but I can add one!

simonauner:
Fixed Robin's comments and rebased to solve conflict.

robinandeer (approved):
Good stuff, we sure we don't need to feature flag any of this, right?

robinandeer (approved):

PR #946, opened 2022-04-28
feat/TrackingContext

marian-buc:
## What?
Added tracking context that manages tracking hook instances

## Why?
To keep the same hook instance and references across pages

**Ticket(s): []**


robinandeer (dismissed):
Looks good!

My first idea was to only store the "queue" in context but I guess this would also work.

Perhaps a nice thing would be to co-locate `useTrackOfferEvent` with the context. But you decide if you wanna do it here or in another PR.

marian-buc:
I could do that, but I thought it makes more sense to have them separated, as context would be supposed to hold multiple hooks, not just that one

PR #947, opened 2022-04-28
GRW-1031 / Feature / Adapt checkout panel to car

simonauner:
## What?

Add car information to checkout panel, if exist.

Replace deprecated `useCurrentLocale`.

## Why?

**Ticket(s): [GRW-1031]**

English:`1,000 mil/year` (I have a discussion ongoing about how to treat this translation):

image

Swedish: `1 000 mil/år`:
image

[GRW-1031]: https://hedvig.atlassian.net/browse/GRW-1031?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/pages/Offer/Checkout/InsuranceSummaryDetails.tsx
+++ b/src/client/pages/Offer/Checkout/InsuranceSummaryDetails.tsx
@@ -6,7 +6,11 @@ import {
   OfferQuote,
   GenericQuoteData,
 } from 'pages/OfferNew/types'
-import { parseAddress } from 'pages/Offer/utils'
+import {
+  formatNumber,
+  formatCarRegistrationNumberSE,
+  parseAddress,
+} from 'pages/Offer/utils'
 import {
   getFormattedBirthdate,
   typeOfResidenceTextKeys,
@@ -14,9 +18,10 @@ import {
 } from 'pages/OfferNew/utils'
 import { formatPostalNumber } from 'utils/postalNumbers'
 import { TextKeyMap, useTextKeys } from 'utils/textKeys'
-import { useCurrentLocale } from 'components/utils/CurrentLocale'
 import { InsuranceType } from 'utils/hooks/useSelectedInsuranceTypes'
 import { TypeOfContract } from 'src/client/data/graphql'
+import { useCurrentLocale } from 'l10n/useCurrentLocale'
+import { LocaleLabel } from 'l10n/locales'
 import { Group, Row } from './InsuranceSummary'
 
 const Label = styled.div`
@@ -53,7 +58,7 @@ export const InsuranceSummaryDetails: React.FC<Props> = ({
         {getPersonalDetails({
           person: personalDetails,
           textKeys,
-          currentLocale,
+          currentLocale: currentLocale.path,
         }).map(({ key, label, value }) => (
           <Row key={key}>
             <Label>{label}</Label>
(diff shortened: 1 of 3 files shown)


src/client/pages/Offer/Checkout/InsuranceSummaryDetails.tsx
```diff
@@ -226,6 +202,94 @@ const getHouseholdSizeValue = (householdSize: number, textKeys: TextKeyMap) => {
...
+    ? [
+        {
+          key: 'registrationNumber',
+          label: textKeys.CHECKOUT_DETAILS_REGISTRATION_NUMBER(),
+          value: formatRegistrationNumber(data.registrationNumber),
+        },
+        {
+          key: 'mileage',
+          label: textKeys.CHECKOUT_DETAILS_MILEAGE(),
+          value: formatMileage(data.mileage, currentLocale),
```
simonauner, line 261: Should be in a translation: "1 000 mil"

src/client/pages/Offer/Checkout/InsuranceSummaryDetails.tsx
```diff
@@ -198,18 +205,16 @@ const getHouseholdSizeValue = (householdSize: number, textKeys: TextKeyMap) => {
 const getCoInsuredMaybe = (
```
robinandeer, line 205: funny that it returns "household size" => but still agree that we should name it like this 😅

src/client/pages/Offer/utils.ts
```diff
@@ -32,3 +35,18 @@ export const getAddress = (quotes: BundledQuote[]) => {
       })
     : ''
 }
+
+export const formatRegistrationNumber = (registrationNumber: string) => {
```
robinandeer, line 39: should this also be named like `formatCarRegistrationNumber`?

robinandeer (approved):
Nice nice!

Just a note that we would need the same changes in the new `Checkout` page that @dendoesit is working on. Just to keep them in sync for when we go live with that in Sweden.

src/client/pages/Offer/utils.ts
simonauner: Yes, more explicit is good 👍  will fix!

src/client/pages/Offer/utils.ts
```diff
@@ -32,3 +35,18 @@ export const getAddress = (quotes: BundledQuote[]) => {
...
+  if (!SE_CAR_REGISTRATION_NUMBER_REGEX.test(registrationNumber))
+    return registrationNumber
+
+  const letters = registrationNumber.match(/[A-Za-z]{3}/)?.[0] || ''
+
+  return registrationNumber.replace(letters, `${letters} `)
+}
+
+export const formatNumber = (number: number, locale: LocaleLabel) => {
+  if (typeof number !== 'number') return number
```
guilhermespopolin, line 49: Disclaimer: I'm not saying this is wrong, I'm just taking this opportunity to initiate a discussion and hopefully collect some opinions. 

This reminds a general question of mine: Isn't one of the benefits of using Typescript not having to write "type validation code" like this one?

I understand that you just want to make sure the value provided is a number since there's nothing that will help us ensure that during runtime, but in situations like that I always find myself wondering: I'm using TS and there is a "contract" about this type (GenericQuote if I'm not wrong) so if I received something different than what's defined in the "contract" wouldn't be ok "to let the app break"? 

Or maybe the idea is to find a balance... 🤔

guilhermespopolin (approved):

src/client/pages/Offer/utils.ts
simonauner: I totally see your point, but at the same time (as you also point out): what happens when we call this function with something that might not be a number, sent from the server? 

I guess I want to safeguard when dealing with data that we do not control, i.e from a server. I would rather have a resilient code base with some type checks here and there than to have the web application crash because a BE server sends some value somewhere which is undefined, I guess. Then on the other hand, that would lead us to have type checks everywhere, which I'm not an advocate of. I think this "slipped in" because `toLocaleString` only exists on `Number` (and `Date`). 

If this was a server, I would probably not do the type check but instead assume that a type validation on the controller would do that for me 🤔

guilhermespopolin: Yep, you've made some valid points :)

PR #948, opened 2022-04-29
GRW-1031 / Feature / Format numbers

simonauner:
## What?

Just a minor change to adjust formatting of (larger) numbers now that we have prices that might exceed 1,000.

## Why?

Different languages have different thousand separators. For `sv` the number `1000` is rendered as `"1 000"` but for `en` it's `"1,000"`. This becomes important for pricing as we might render prices above 1,000 for car insurance.

**Ticket(s): [GRW-1031](https://hedvig.atlassian.net/browse/GRW-1031)** (not really specified in the ticket, just a UX consideration from my part)

Swedish:

image

English:

image

### DIFF
--- a/src/client/l10n/useLocalizeNumber.ts
+++ b/src/client/l10n/useLocalizeNumber.ts
@@ -1,0 +1,29 @@
+import { useCallback } from 'react'
+import { useCurrentLocale } from './useCurrentLocale'
+
+/**
+ * Returns a localized number string
+ * @param number
+ * @param lang ISO two-letter language code, for example 'sv'
+ * @returns Number formatted according to browser's language settings
+ */
+export const localizeNumber = (number: number, lang: string): string => {
+  if (typeof number !== 'number') return number
+
+  return number.toLocaleString(lang)
+}
+
+/**
+ * Returns a function that returns a localized number string
+ */
+export const useLocalizeNumber = () => {
+  const locale = useCurrentLocale()
+  const lang = locale.htmlLang
+
+  const localize = useCallback(
+    (number: number) => localizeNumber(number, lang),
+    [lang],
+  )
+
+  return localize
+}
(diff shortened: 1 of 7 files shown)


robinandeer (approved):
Nice, perhaps we can take it one step further and create `useFormatNumber` that automatically gets the locale information. But it's fine to merge like this.

src/client/pages/Offer/utils.ts
```diff
@@ -45,8 +44,8 @@ export const formatRegistrationNumber = (registrationNumber: string) => {
   return registrationNumber.replace(letters, `${letters} `)
 }
 
-export const formatNumber = (number: number, locale: LocaleLabel) => {
+export const formatNumber = (number: number, lang: string) => {
```
guilhermespopolin, line 47: nit: maybe call it `localizeNumber`?

guilhermespopolin (approved):
Noice

src/client/pages/Offer/utils.ts
simonauner: Ah, yes, good idea!

src/client/l10n/useLocalizeNumber.ts
```diff
@@ -0,0 +1,33 @@
...
+  const localizeNumber = useCallback(
+    (number: number) => {
+      if (typeof number !== 'number') return number
+
+      return number.toLocaleString(lang)
+    },
+    [lang],
+  )
+
+  return localizeNumber
```
robinandeer, line 32:
```suggestion
  return useCallback(
    (number: number) => localizeNumber(number, lang),
    [lang],
  )
```

robinandeer (approved):
Looks good to me, just added a "nit" suggestion to keep things DRY ;)

src/client/l10n/useLocalizeNumber.ts
simonauner: Ah, good thinking! Fixed it and added a test as well

robinandeer (approved):

PR #952, opened 2022-04-29
isDisabled hook fix

marian-buc:
## What?
Fixes the gql bug by disabling the tracking hook on pages that do not expose a quoteCartId in the url

## Why?

**Ticket(s): []**

### DIFF
--- a/src/client/utils/tracking/trackOfferEvent.ts
+++ b/src/client/utils/tracking/trackOfferEvent.ts
@@ -96,13 +96,14 @@ export const useTrackOfferEvent = () => {
 
   const { quoteCartId } = useMatchedQuoteCartIdFromUrl()
 
+  const isDisabled = quoteCartId === undefined
   const [selectedInsuranceTypes] = useSelectedInsuranceTypes()
-
   const { data: quoteCartQueryData } = useQuoteCartQuery({
     variables: {
-      id: quoteCartId!,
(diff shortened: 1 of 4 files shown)


src/client/utils/tracking/trackOfferEvent.ts
```diff
@@ -96,13 +96,14 @@ export const useTrackOfferEvent = () => {
...
   const [selectedInsuranceTypes] = useSelectedInsuranceTypes()
-
   const { data: quoteCartQueryData } = useQuoteCartQuery({
     variables: {
-      id: quoteCartId!,
+      id: quoteCartId || '',
       locale: isoLocale,
     },
+    skip: isDisabled,
   })
```
mehdielket, line 107: This query will still be called many times and fail in the landing page and during the embark flow. Can we try to find a solution where we do not call it to start with?

marian-buc: no, it will not be called, that's what skip param does :D

mehdielket (dismissed):
🚀

dendoesit (commented):
when following the normal flow the param will be 'ownership' in this case as we still do not have any quoteCartId yet.
image

dendoesit (changes requested):
I believe this can be closed as we have the other fix ticket.

PR #953, opened 2022-04-29
fix/Tracking hook, context

marian-buc:
## What?
Removed tracking context and changed hook to call the callback directly after data is loaded

## Why?
To avoid querying the data where the hook is not used or having the queue go stale after rendering

**Ticket(s): []**


src/client/pages/Checkout/CheckoutDetails/CheckoutDetails.tsx
```diff
@@ -26,14 +26,14 @@ export const CheckoutDetails = () => {
...
-  const { trackOfferEvent } = useTrackingContext()
+  const trackOfferEvent = useTrackOfferEvent()
   const data = useQuoteCartData()
 
-  useEffect(() => trackOfferEvent({ eventName: EventName.CheckoutOpen }), [
-    trackOfferEvent,
-  ])
+  const { quoteCartId } = useQuoteCartIdFromUrl()
+
+  useEffect(() => trackOfferEvent({ eventName: EventName.CheckoutOpen }), [])
```
robinandeer, line 35: I think we should add `trackOfferEvent` to the dependency array

src/routes.tsx
```diff
@@ -425,5 +425,3 @@ export const routes: Route[] = [
     },
   },
 ]
-
-export const routePaths = routes.map((route) => route.path.toString())
```
robinandeer, line 429: is this not imported anywhere?

robinandeer (commented):
I'm not really sure what the motivation is here.

Didn't we already find some race conditions where we never send the analytics events?

src/routes.tsx
dendoesit: this was added so we can use it in useQuoteCartIdFromUrl but we are rolling all that logic back.

robinandeer (approved):
Looks good, now I understand!

PR #955, opened 2022-05-02
GRW-1024 / Fix / General terms checkout details

amandaflod:
## What?

Remove the filtering out of General terms document on the Checkout details page. 

## Why?

The original figma did not include the General terms which is why they were excluded, although they are supposed to be there.

**Ticket(s): [GRW-1024]**

Screenshot 2022-05-02 at 10 22 19

[GRW-1024]: https://hedvig.atlassian.net/browse/GRW-1024?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dendoesit (dismissed):
looks good. I am curious however why this was filtered out in the first place and if we should remove the filter 🤔 . can we look a bit into that before we release? maybe @robinandeer has a clue ?

amandaflod:
I assume it was because they were not included in the Figma before and we didn't question it? 🤔 @dendoesit

robinandeer (approved):
Hmm seems like a good change!

Why would we do any filtering here?

amandaflod:
Hmm are you guys thinking we should include the privacy policy as well? 🙈 I can ask Zak! @dendoesit @robinandeer

robinandeer:
They way I see it is that we just display all the documents in the list -- we don't worry about what they are. That should be controlled in PCMS.

Perhaps we looked a little too close to matching the Figma file ;)

amandaflod:
That sounds good to me! Should I remove the filtering both here and on the offer page then? :))

robinandeer:
@amandaflod I would say so. In general we should just render what the API gives us.

robinandeer (commented):
Nice guess you just need to remove the unused import :)

amandaflod:
Yessir! Already done ;) @robinandeer

dendoesit (approved):
looks good.

PR #956, opened 2022-05-02
GRW-1027 / Feature / Add car header image

simonauner:
## What?

Add support for `BackgroundImage` to render different images based on insurance type

## Why?

Since we want to show a different header image for Car insurance sometimes we need to enable passing a `variant` to the `BackgroundImage` component.

**Ticket(s): [GRW-1027]**

The original image is quite low-res, but I scaled it up to match the pixel dimensions of the living room header images, we can add the high-res images later (GRW-1041).

image

[GRW-1027]: https://hedvig.atlassian.net/browse/GRW-1027?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/pages/Offer/utils.ts
```diff
@@ -32,3 +32,7 @@ export const getAddress = (quotes: BundledQuote[]) => {
       })
     : ''
 }
+
+export const bundleContainsCarQuote = (bundle: BundledQuote[]) => {
```
simonauner, line 36: Maybe this is not a "bundle"? I can happily change name if anyone has any suggestions.

guilhermespopolin: Maybe this can leave inside [quoteBundleSelectors](https://github.com/HedvigInsurance/web-onboarding/blob/master/src/client/api/quoteBundleSelectors.ts#L21)

guilhermespopolin (approved):

src/client/pages/Offer/Introduction/index.tsx
```diff
@@ -14,6 +14,7 @@ import { ExternalInsuranceProvider } from '../../OfferNew/Introduction/ExternalI
 
 import { Sidebar } from './Sidebar'
 import { HeroOfferDetails } from './HeroOfferDetails'
+import { bundleContainsCarQuote } from 'src/client/api/quoteBundleSelectors'
```
robinandeer, line 17:
```suggestion
import { bundleContainsCarQuote } from 'api/quoteBundleSelectors'
```

robinandeer (commented):
Aaaaaaaalmost looks _great_! ;)

Had a suggestion to update name of selector -- and just clean up one import statement

src/client/api/quoteBundleSelectors.ts
```diff
@@ -84,3 +84,7 @@ export const getBundleCurrency = (bundle: QuoteBundle) => {
 export const getTotalBundleCost = (bundle: QuoteBundle) => {
   return bundle.bundleCost.monthlyNet.amount
 }
+
+export const bundleContainsCarQuote = (bundle: BundledQuote[]) => {
+  return bundle.some((quote) => quote.data.type === 'SWEDISH_CAR')
+}
```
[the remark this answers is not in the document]
simonauner: Yes, I guess, but the Introduction component from which this is called isn't fed a `QuoteBundle`, just an array of `BundledQuote` :/ 
I' not sure how much time I want to put into refactoring that. What do you think?

But the renaming I can do!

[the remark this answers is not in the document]
robinandeer: right, no I think that's good reasoning :)

PR #957, opened 2022-05-02
Fix / Enable CPAS routes in production

marian-buc:
## What?
This enables the checkout/details and checkout/payment routes in production

## Why?
CPAS rollout :D

**Ticket(s): []**


dendoesit (approved):
👍

PR #958, opened 2022-05-02
bugifx/minor copy changes

dendoesit:
## What?

Removed helper text for email. Added space to error modal. 

## Why?

Design changes


robinandeer (approved):
haha why don't we need the helper text?

great on the error modal spacing - that was bugging me!

dendoesit:
@robinandeer Design and copy changed their mind about the helper text as we do not need to explain it since everyone asks for email 🧷

PR #959, opened 2022-05-02
GRW-1000 / Feature / Add car route to Embark

simonauner:
## What?

Add `/car` route to Embark section of site so that the Car flow can be launched.

## Why?

**Ticket(s): [GRW-1000]**

image

[GRW-1000]: https://hedvig.atlassian.net/browse/GRW-1000?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robinandeer (commented):
Not sure if we need to disable in prod => maybe just put it behind a feature flag?

simonauner:
> Not sure if we need to disable in prod => maybe just put it behind a feature flag?

Good idea! Added this now

robinandeer (approved):

PR #960, opened 2022-05-02
feat(LanguageSwitcher): bind it to feature flag for Embark

guilhermespopolin:
## What?

Hide `LanguagePicker` under a feature flag for Embark (`FEATURE_LANGUAGE_SWITCHER_AT_EMBARK`)

## Why?

We make a client side navigation while using the Language Picker inside Embark and currently that would mean redirect to the old Landing Page. So while we don’t fully migrate to the new Landing Page we should disable Language Picker in Embark.

## How to test it

**Ticket(s): [GRW-1036](https://hedvig.atlassian.net/browse/GRW-1036)**


src/client/pages/Embark/index.tsx
```diff
@@ -171,17 +184,7 @@ const Embark = (props: EmbarkProps) => {
                 passage={currentPassage}
                 storyData={state.data}
                 startPageLink={props.startPageLink}
-                customTrailingContent={
```
guilhermespopolin, line 174: So, by design, there shouldn't be a case where both `LanguagePicker` and `CallCenterPhoneNumber` should be displayed?! 🤔

robinandeer: Exactly, they occupy the same space on the page

robinandeer (commented):
Hmmm could the other solution be to opt-out of client-side navigation for the language picker?

mehdielket (approved):
🚀  Let's try to remember to fix this properly asap if we can

mehdielket:
> Hmmm could the other solution be to opt-out of client-side navigation for the language picker?

Definitely, depending on the urgency of the fix.

src/client/pages/Embark/index.tsx
```diff
@@ -179,7 +181,10 @@ const Embark = (props: EmbarkProps) => {
                       onClick={handleClickPhoneNumber}
                     />
                   ) : (
-                    <LanguagePicker color="black" />
+                    <LanguagePicker
+                      color="black"
+                      performClientSideNavigation={market !== Market.No}
```
robinandeer, line 186: Avoid "market checks", they don't scale => I think this case is better solved by a feature flag

src/client/components/LanguagePicker/LanguagePicker.tsx
```diff
@@ -71,10 +79,14 @@ export const LanguagePicker: React.FC<Props> = ({
...
             <ActiveOption color={color}>{shortTitle}</ActiveOption>
-          ) : (
+          ) : performClientSideNavigation ? (
             <LinkOption color={color} to={`/${linkTo}${path}`}>
               {shortTitle}
             </LinkOption>
+          ) : (
+            <NativeLinkOption color={color} href={`/${linkTo}${path}`}>
+              {shortTitle}
+            </NativeLinkOption>
```
robinandeer, line 89: nit: Wouldn't it be possible to use `as`-prop here?

```
<LinkOption color={color} to={`/${linkTo}${path}`} as={performClientSideNavigation ? undefined : 'a'}>
  {shortTitle}
</LinkOption>
```

robinandeer (commented):
Nice problem solving -- I just wanna push for our "no market checks" policy 😉

src/client/pages/Embark/index.tsx
guilhermespopolin: 🤔 what would you call that feature flag?

src/client/components/LanguagePicker/LanguagePicker.tsx
guilhermespopolin: I guess it doesn't work for this case 😞 

> The as prop is only used by styled when it’s not forwarded to the underlying element. By default, this means that the as prop is used for html tags and forwarded for components. To change this, you can pass a custom [shouldForwardProp](https://emotion.sh/docs/styled#customizing-prop-forwarding) which returns true for 'as' to forward it or returns false for 'as' to use it and not forward it.

image

robinandeer (approved):
Appreciate the attention to detail :)

PR #966, opened 2022-05-04
Chore / Remove Lottie animation from dev-only 404 page

robinandeer:
## What?

See title.
Remove all lottie dependencies.

## Why?

Lottie is a big lib and the only place we use it is on the 404 page - which is not visible in production.

**Ticket(s): []**


dendoesit (approved):
⭐

PR #967, opened 2022-05-04
Bugfix/cpas mutation

dendoesit:
## What?

Start checkout mutation is being called twice inside connect payment when the button is called resulting in a second StartCheckout mutation that ends up in failure showing the error modal for a couple of seconds and causing a Member mutation failure.

## Why?

I believe the start checkout mutation takes too long so the even is fired twice. 

This may not be the best approach so happy to hear other options to fix this. 

Now : 

image


src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
```diff
@@ -242,6 +246,8 @@ export const CheckoutPayment = ({
         })
         throw new Error('Manual Review required')
       }
+      setIsCheckoutLoading(false)
```
robinandeer, line 249: should this be in a "finally"-block?

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
```diff
@@ -224,6 +227,7 @@ export const CheckoutPayment = ({
           eventName: EventName.CheckoutErrorBasicError,
         })
         setIsError(true)
+        setIsCheckoutLoading(true)
```
robinandeer, line 230: why "true" - shouldn't it be false?

robinandeer (commented):
Nice find - let's get it out

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
dendoesit: yeah, meant to write false.

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
dendoesit: good point. will add.

robinandeer (approved):
Nice, I hope this will do the trick :)

PR #968, opened 2022-05-04
Feat / GRW-1039 / CPAS tracking final

marian-buc:
## What?

- Removed quoteCartId from outside offer_data 
- Refactored sign_error event
- Added payment_details_confirmed event
- Refactored tracking hook to use async apolloClient.query instead of useQuery, removing race conditions and keeping the same API

## Why?
Overhauling tracking for CPAS 

**Ticket(s): [GRW-1039]**

[GRW-1039]: https://hedvig.atlassian.net/browse/GRW-1039?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/utils/tracking/trackOfferEvent.ts
```diff
@@ -100,62 +105,51 @@ export const useTrackOfferEvent = () => {
...
-    (eventParams: EventParameters) => {
-      if (selectedBundleVariant) {
-        trackOfferCallback(eventParams, selectedBundleVariant)
-      } else {
-        setEventQueue((prevQueue) => [...prevQueue, eventParams])
+    ({ eventName, options = {} }: EventParameters) => {
+      const runQuery = async () => {
+        if (apolloClient && quoteCartId) {
+          try {
+            const quoteCartQuery = await apolloClient!.client.query<
```
robinandeer, line 113: Why do we need the hashbang here if we do the if statement above?

src/client/utils/tracking/trackOfferEvent.ts
```diff
@@ -100,62 +105,51 @@ export const useTrackOfferEvent = () => {
...
+              variables: {
+                id: quoteCartId,
+                locale: isoLocale,
+              },
+            })
+            return quoteCartQuery.data
+          } catch (e) {
+            return
+          }
+        } else return
```
robinandeer, line 126: Guess you could handle this with "finally" block... but shouldn't we do something/log something is the query fails?

robinandeer (commented):
Nice work on this. I had some comments - let me know if you wanna go over together.

I guess you discussed the updated event names with Sonny?

So with this approach, we skip the queue altogether but call the API/cache as soon as we call the API. I like it I think. The only thing is that this approach is not resilient if the query fails. Perhaps we should at least log that to datadog maybe?

marian-buc:
@robinandeer @dendoesit addressed the err handling in https://github.com/HedvigInsurance/web-onboarding/pull/972

src/client/utils/tracking/trackOfferEvent.ts
robinandeer: @marian-buc bump ;)

marian-buc: https://github.com/HedvigInsurance/web-onboarding/blob/0c888ab09114125b3ad47a5b7c76724a5e11d461/src/client/apolloClient.ts#L89

Query still might fail event if quoteCartId is supplied

robinandeer: hmmm not sure what you mean there.. I was referring to the non-null assertion operator `!` here:

```
apolloClient!.client.query
```

Why is it still necessary? :)

robinandeer (approved):
Consider my comments on the non-null assertion operator -- otherwise I think it looks great!

robinandeer (approved):

src/client/utils/tracking/trackOfferEvent.ts
marian-buc: haha, yup, removed it in the new PR as well
[by passing the client as a param](https://github.com/HedvigInsurance/web-onboarding/pull/972/commits/9e89fe0c8438dd94f416448a38a84a8927907c26#diff-baa1c36a82a2c7876dc53d4df01709b66f9257464f8359b7f700d8e628abe6a5R32)

PR #970, opened 2022-05-04
Fix / Add google pay domain to CSP

robinandeer:
## What?

Allow scripts from Google Pay domain.

## Why?

Needed for adyen google pay option. Currently, we block that script.

**Ticket(s): []**


dendoesit (approved):
👍

PR #971, opened 2022-05-05
change offer-debugger button position

dendoesit:
## What?

Changed button in offer-debugger to bring it up.

## Why?

(make life easier ? ) I hate scrolling to the bottom of the page and I never change any of the inputs. This is a personal preference so please let's discuss if it is not good. 

Image : 
image


robinandeer (approved):
Haha yeah I totally agree with you -- maybe add some padding below but this is better already!

PR #974, opened 2022-05-05
Feature / Bring back Adyen buttons to submit form

robinandeer:
Show success status if "isPaymentConnected".
Skip returning dropin component from hook.

## Why?

**Ticket(s): []**


dendoesit (approved):
looks great. 👍  tested it on mine and works as expected.

PR #975, opened 2022-05-06
GRW-1062 / feat / add table component

simonauner:
## What?

This PR:
- Adds `ThinTick` icon which is in Figma files but missing from the repository (will happily take suggestions on naming...)
- Adds margins util so that we can add `mt`, `mx` etc margins for components instead of adding custom styling everywhere
- Adds Table component with API heavily inspired by [MUI](https://mui.com/material-ui/react-table/) and Kent C. Dodds [Compound Components](https://kentcdodds.com/blog/compound-components-with-react-hooks)

## Why?

GRW-1062 ticket says that we should add a comparison table to the offer page. This is a component for that use-case. One of the stories mimics what we need to show visually.

**Ticket(s): [GRW-1062]**

Our many ways to show checkmarks:

image

Basic data:

image

As a comparison table:

image

This is the [Figma sketch](https://www.figma.com/file/HEQXHw3ptHZTltJz38hZ9T/Car-Insurance-%C2%B7-EXPLORATION?node-id=1239%3A18556) that it was based on:
image

[GRW-1062]: https://hedvig.atlassian.net/browse/GRW-1062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/components/Table/Table.stories.tsx
+++ b/src/client/components/Table/Table.stories.tsx
@@ -1,0 +1,170 @@
+import React from 'react'
+import { Story } from '@storybook/react'
+import { colorsV3 } from '@hedviginsurance/brand'
+import { ThinTick } from '../icons/ThinTick'
+import {
+  Table,
+  TableBody,
+  TableCell,
+  TableHead,
+  TableProps,
+  TableRow,
+} from './Table'
+
+export default {
+  title: 'Components/Table',
+  component: Table,
+  parameters: {
+    backgrounds: {
+      default: 'gray100',
+    },
+  },
+  args: {
+    minWidth: '650px',
+  },
+}
+
+function createData(
+  name: string,
+  calories: number,
+  fat: number,
+  carbs: number,
+  protein: number,
+) {
+  return { name, calories, fat, carbs, protein }
+}
+
+const rows = [
+  createData('Frozen yoghurt', 159, 6.0, 24, 4.0),
+  createData('Ice cream sandwich', 237, 9.0, 37, 4.3),
+  createData('Eclair', 262, 16.0, 24, 6.0),
+  createData('Cupcake', 305, 3.7, 67, 4.3),
+  createData('Gingerbread', 356, 16.0, 49, 3.9),
+]
+
+const Template: Story<TableProps> = (args) => (
+  <Table {...args}>
+    <TableHead>
+      <TableRow>
+        <TableCell>Dessert (100g serving)</TableCell>
+        <TableCell align="right">Calories</TableCell>
+        <TableCell align="right">Fat&nbsp;(g)</TableCell>
+        <TableCell align="right">Carbs&nbsp;(g)</TableCell>
+        <TableCell align="right">Protein&nbsp;(g)</TableCell>
+      </TableRow>
+    </TableHead>
+    <TableBody>
+      {rows.map((row) => (
+        <TableRow key={row.name}>
+          <TableCell>{row.name}</TableCell>
+          <TableCell align="right">{row.calories}</TableCell>
+          <TableCell align="right">{row.fat}</TableCell>
+          <TableCell align="right">{row.carbs}</TableCell>
+          <TableCell align="right">{row.protein}</TableCell>
+        </TableRow>
+      ))}
+    </TableBody>
+  </Table>
+)
+
+export const Default = Template.bind({})
+Default.args = {}
+
+function createDataCarInsuranceData(
+  name: string,
+  traffic: boolean,
+  half: boolean,
+  full: boolean,
+) {
+  return { name, traffic, half, full }
+}
+
+const comparisonRows = [
+  createDataCarInsuranceData('Personskador', true, true, true),
+  createDataCarInsuranceData('Skador på annans egendom', true, true, true),
+  createDataCarInsuranceData('Stöld och inbrott', false, true, true),
+  createDataCarInsuranceData('Brand', false, true, true),
+  createDataCarInsuranceData('Glasskador', false, true, true),
+  createDataCarInsuranceData('Bärgning', false, true, true),
+  createDataCarInsuranceData('Skador på bilen från olycka', false, false, true),
+]
+
+const premiumRows = [
+  createDataCarInsuranceData('Personskador', false, true, true),
+  createDataCarInsuranceData('Skador på annans egendom', false, true, true),
+]
+
+const ComparisonTemplate: Story<TableProps> = (args) => (
+  <>
+    <Table {...args}>
+      <TableHead>
+        <TableRow>
+          <TableCell></TableCell>
+          <TableCell align="center">Trafik</TableCell>
+          <TableCell align="center">Halv</TableCell>
+          <TableCell align="center">Hel</TableCell>
+        </TableRow>
+      </TableHead>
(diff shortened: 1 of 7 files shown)


src/client/components/Table/Table.stories.tsx
```diff
@@ -0,0 +1,171 @@
...
+
+const ComparisonTemplate: Story<TableProps> = (args) => (
+  <>
+    <Table {...args}>
+      <TableHead>
+        <TableRow>
+          <TableHeaderCell></TableHeaderCell>
+          <TableHeaderCell align="center">Trafik</TableHeaderCell>
+          <TableHeaderCell align="center">Halv</TableHeaderCell>
+          <TableHeaderCell align="center">Hel</TableHeaderCell>
```
gustaveen, line 106: Would be nice if we could create a data structure for this as well. :) Thinking if we move this to PCMS you could first create the columns (Trafik, Halv, Hel) and then you add rows with a label and a boolean for each column

src/client/components/utils/margins.ts
```diff
@@ -0,0 +1,26 @@
...
+  mr?: string
+  // margin-top
+  mt?: string
+  // margin-bottom
+  mb?: string
+  // all margins
+  m?: string
+}
+
+export const getMargins = (margins: Margins) => {
```
gustaveen, line 18: 👌

gustaveen (commented):
Really nice @simonauner ! 🙌 Added some comments but in general I think we can start with this for sure! Saw that Mange added kind of an accordion to read more about each row, but add that in upcoming iterations.

src/client/components/Table/Table.stories.tsx
simonauner: Yeah, I'm thinking that the alternative to using the MUI API design is to do like [Ant](https://ant.design/components/table) and pass in the data structure instead. But I'm thinking that that will be another component that translates whatever we get from giraffe/PCMS into this structure. Makes sense?

src/client/components/Table/Table.stories.tsx
```diff
@@ -0,0 +1,171 @@
...
+      </TableHead>
+      <TableBody>
+        {premiumRows.map((row) => (
+          <TableRow key={row.name}>
+            <TableCell>{row.name}</TableCell>
+            <TableCell
+              align="center"
+              color={row.traffic ? undefined : colorsV3.gray300}
+            >
+              {row.traffic ? <ThinTick /> : <>&mdash;</>}
```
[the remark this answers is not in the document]
simonauner: Yes, for sure! I took the easy way out and just added a UTF8 character instead 😴

src/client/components/Table/Table.stories.tsx
gustaveen: Ah yes, the API for the component is great, I was thinking more on the datastructure that we would get from PCMS later on, so no need to change anything here. 🙌

src/client/components/Table/Table.tsx
```diff
@@ -0,0 +1,59 @@
...
+  maxWidth?: string
+  fullWidth?: boolean
+} & Margins
+
+const Table = styled.table(
+  ({ minWidth, maxWidth, fullWidth, ...props }: TableProps) => ({
+    borderCollapse: 'collapse',
+    borderSpacing: 0,
+    minWidth,
+    maxWidth,
```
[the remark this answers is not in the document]
gustaveen: That was my initial thought to hav it full-width as default and leave the width to the container. But tables are a bit special so it might be a good thing to have on the actual component. Let's evaluate :)

src/client/components/Table/Table.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+const TableCell = styled.td((props: TableCellProps) => ({
+  ...getTableCellStyles(props),
+  padding: '0.5rem 0',
+
+  borderBottom: `1px solid ${colorsV3.gray300}`,
+}))
+
+const TableHeaderCell = styled.th((props: Omit<TableCellProps, 'color'>) => ({
```
robinandeer, line 51: Just for consistency maybe we can call this `TableHeadCell`?

Or go all the way and render based on context like MUI: https://mui.com/material-ui/api/table-cell/ ;)

> The component renders a <th> element when the parent context is a header or otherwise a <td> element.

robinandeer (approved):
Nice nice! Really like the API - I think it's gonna be super flexible :)

Was curious about the `getMargins` utility - have you used it anywhere yet (just to see an example)? Generally, I think we said we prefer using layout components over pushing adjacent elements with margins - but sometimes it will def be useful!

simonauner:
> Nice nice! Really like the API - I think it's gonna be super flexible :)
> 
> Was curious about the `getMargins` utility - have you used it anywhere yet (just to see an example)? Generally, I think we said we prefer using layout components over pushing adjacent elements with margins - but sometimes it will def be useful!

Thanks! The utility is the way to get margins added to the component's styling, o that a component can support attributes like `mx="4rem"`, `mt="2rem"`, etc:

https://github.com/HedvigInsurance/web-onboarding/pull/975/files#diff-1bc6263a87c56d7c8e3aa963b54040c1b3a33d38c12c75c1e80cdc7f6701e126R135

 I guess it's a matter of taste and if we prefer to have layout components in the example above all we need to do is to add a wrapping `<div>` with a `margin-top`, I guess? I can do either.

src/client/components/Table/Table.tsx
simonauner: Yeah, I was considering the conditional rendering based on parent.. I can try it with a simple `createContext` perhaps

simonauner:
@robinandeer removed `TableHeaderCell` -> `TableCell` now renders a `th` when inside a `thead` and a `td` otherwise. Let me know what you think!

robinandeer:
> Does it sound reasonable to you?

Great updates! :D

...and yes sounds very reasonable. I'm thinking of the case where we (probably) will need different margins for different viewports. Do you see something like this supporting that in the future? Or is that an advanced use case where we would wrap the element inside another div? (Totally fine to solve that problem later)

robinandeer (approved):

simonauner:
> > Does it sound reasonable to you?
> 
> Great updates! :D
> 
> ...and yes sounds very reasonable. I'm thinking of the case where we (probably) will need different margins for different viewports. Do you see something like this supporting that in the future? Or is that an advanced use case where we would wrap the element inside another div? (Totally fine to solve that problem later)

I find it easier to maintain and read that kind of code with just CSS and media queries.. I'm not a big fan of Bootstrap's `mt={xl : "4rem", m: "2rem"}` syntax to be honest, so I wouldn't say that we should support that with this utility 🤷

gustaveen:
Just to add to the margins/paddings discussion @robinandeer @simonauner  :) Simon and I discussed it when Simon added it to racoon. I think it's a neat to have that capability on components and not always have to reach for a layout component or a wrap it in a div to have some padding/margin. It's also a well adopted pattern on ui libs out there 🙏 If we choose to not go with a ui lib, I think we can adopt some of the good parts of the API they use 🎉

PR #977, opened 2022-05-06
Refactor / Move auto form submit logic to Checkout

robinandeer:
## What?

Delete debounce function in favor of custom hook.
Extract SSN auto submit logic to a hook.
Show errors to user when form submit fails.

## Why?

Code scouting. ESLint shows a warning with the current solution.

We also want to show an error to the user if the mutation fails.

**Ticket(s): []**


simonauner (approved):
Nice! 🙌

guilhermespopolin (approved):
Nice

dendoesit (approved):
great stuff 👍

PR #979, opened 2022-05-06
Feature / Add configuration for google pay

robinandeer:
## What?

Add types for AdyenEnvironment.

Documentation: https://docs.adyen.com/payment-methods/google-pay/web-drop-in

## Why?

We are getting an error in the console when you try to use the google pay option.

Still doesn't seem to work but this will still be required.

**Ticket(s): []**


simonauner (approved):

dendoesit:
great work. from what I talked with Fredrik we do not have that enabled in prod right now 👍

robinandeer:
@dendoesit yes, I saw a ticket for it! I can maybe leave this PR open until we get to that task :)

PR #980, opened 2022-05-09
 added payment confirmed tracking

dendoesit:
## What?

Added tracking to payment confirmed.

## Why?

New changes to the flow of payment requires changing the tracking implementation

image


robinandeer (approved):
Nice!

PR #981, opened 2022-05-09
fix(GRW-1048): NO Accident - Current insurer skip button

guilhermespopolin:
## What?

Changes the parameters used to determine the necessity of 'locking start date' selection based on _current insurer_

## Why?

The logic used previously in Offer Page for that goal can be described as follow:

> _An insurance that should be cancelled by Hedvig is the one which its quote has a **current insurer associated with it**. With that, there's no point of letting users picking a start date since it can be automatically defined: starts when your old insurance expires_

However that might not be so accurate since there's the possibility of having quotes which its current insurer aren't switchable, which means Hedvig can't cancel them automatically. Therefore, the logic has changed for something that can be described as following:

> _An insurance that should be cancelled by Hedvig is the one which its quote has a **switchable current insurer associated with it**. With that, there's no point of letting users picking a start date since it can be automatically defined: starts when your old insurance expires_

**Ticket(s): [GRW-1048](https://hedvig.atlassian.net/browse/GRW-1048)**

## Screenshots / recordings

image

### DIFF
--- a/src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
+++ b/src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
@@ -160,7 +160,7 @@ const getDefaultDateValue = (
     return parse(startDate, 'yyyy-MM-dd', new Date())
   }
 
-  if (currentInsurer) {
+  if (currentInsurer?.switchable) {
     return null
   }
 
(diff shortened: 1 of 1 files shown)


robinandeer (approved):
Nice! I wonder where a good place to document this would be 🤔

guilhermespopolin:
> Nice! I wonder where a good place to document this would be 🤔

Good question. It sounds too specific for Notion and in the code it's not so easy to figure it out 🤔

PR #982, opened 2022-05-09
Grw-1078 / Cpas add paymentTokenId

dendoesit:
## What?

- Added payment token inside of Adyen.
-  Changed the logic of IsPaymentConnected to get triggered when payment has been succesful instead of if payment token is added to the quoteCart

## Why?

In order to get better tracking backend now return a paymentTokenID even if connect_payment fails. The id gets added to the quoteCart . This enables us to see what happened to the payment based on the quoteCart.

### DIFF
--- a/src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
+++ b/src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
@@ -16,7 +16,6 @@ import {
   QuoteBundleVariant,
   BundledQuote,
   CheckoutStatus,
-  useAddPaymentTokenMutation,
 } from 'data/graphql'
 import { MEDIUM_SMALL_SCREEN_MEDIA_QUERY } from 'utils/mediaQueries'
 import { Headline } from 'components/Headline/Headline'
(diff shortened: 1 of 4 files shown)


src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
```diff
@@ -213,26 +210,9 @@ export const CheckoutPayment = ({
...
-          options: { error, errorType: ErrorEventType.PaymentTokenMutation },
-        })
-        console.error('Failed to add Payment Token :', error.message, error)
-      }
-    },
-    [addPaymentTokenMutation, quoteCartId, trackOfferEvent],
-  )
+  const addPaymentToCart = useCallback(async () => {
+    setIsPaymentConnected(true)
+  }, [])
```
robinandeer, line 215: Don't think this is necessary - you don't need to add `useState` function to the dependency array :)

I think the name of this function is confusing. I would just call it inside the `useEffect`.

src/client/pages/ConnectPayment/components/useAdyenCheckout.tsx
```diff
@@ -151,6 +155,7 @@ interface AdyenCheckoutProps {
   currentLocale: LocaleData
   paymentMethodsResponse: Scalars['PaymentMethodsResponse']
   connectPaymentMutation: any
+  addPaymentTokenMutation: any
```
robinandeer, line 158: Why not type this properly? ;)

robinandeer (approved):
I think it looks good! Just minor comments

Let's continue iterating on this after we launch :D

PR #985, opened 2022-05-10
GRW-1062 / feat / Add comparison table

simonauner:
## What?

This PR adds a comparison table to the offer screen when there are more than one variant on display.

By finding all unique perils for all quotes on display we can create a table and see which peril that exists in each bundle.

I checked with Mange and Zak and they approved to have this on all offer screens, not just Car. That being said, we need to await final copy and design for where the buttons should go before switching on the feature flag.

This is how it looks for Swedish Home + Accident:

https://user-images.githubusercontent.com/1343979/167628533-d98da071-ff56-4c04-8091-54f46dd39386.mov

For Car on mobile (ignore the wrong perils, it's incorrect from BE):

https://user-images.githubusercontent.com/1343979/167629510-801bcc8a-7176-4bfb-ae17-6c73e71da29f.mov

When only one variation, no button shows:

image

## Why?

**Ticket(s): [GRW-1062]**

[GRW-1062]: https://hedvig.atlassian.net/browse/GRW-1062?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simonauner:
Just rebased from `master` and fixed a failing snapshot test...

gustaveen (approved):
Yaay, great stuff! 🎉

One thing to consider might be if we can get some shorter names for the table columns

PR #986, opened 2022-05-10
Fix / Prevent clean-up function from crashing app

robinandeer:
## What?

Log to datadog if unexpected error occurs.

## Why?

The app crashes in production with error:

```
NotFoundError: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
```

https://datadoghq.eu/rum/error-tracking?issueId=c6b5d57c-cf96-11ec-8ff0-da7ad0900005&start=1652189692835&end=1652193292835&paused=false

**Ticket(s): []**


dendoesit (approved):
✔️

PR #988, opened 2022-05-10
Feature / Add workflow to upload source maps

robinandeer:
## What?

Add dependency on Datadog CLI.

## Why?

This is needed to improve stack traces in the error tracking with Datadog.

**Ticket(s): []**


simonauner (approved):
Wow, nice solution!

I guess that we ideally would like to run this as a post-hook in the Heroku promote-to-prod action. Since now we might have diffs for some time in-between when someone pushes to master vs promotes to production. But I think we should try this first and see that it works before investing more time perhaps?

https://devcenter.heroku.com/articles/release-phase

robinandeer:
Good point! I kind of assumed the build output would be deterministic but guess it still makes sense to do it as part of the deploy pipeline

simonauner (approved):

simonauner:
For transparency's sake: me and Robin talked and I didn't know that Datadog connects the git SHA of the application with the source map: https://github.com/HedvigInsurance/web-onboarding/pull/988/files#diff-d04565070ef767bb8a5203e487673da9866d232090b480e5b6e64f12b8d6eaf3R33

This solves my concern above.
Two thumbs up from me!

gustaveen (approved):
Awesome!

PR #991, opened 2022-05-11
GRW-1021 Add support for house NO in GTM tracking and small refactoring

mehdielket:
## What?

Adds the following and removes the following fields in the GTM event tracking
(-) insurance_type: Does not scale, can be deduced from the has_home, has_house, has_accident, has_travel, ownership_type
(+) has_house: To track if bundle has house insurance
(+) ownership_type: RENT | OWN
(-) initial offer & current_offer: Do not scale and will be deduced from the quoteCart id
(-) switch_to: not needed
(+) switch_from: does not scale, was changed to a list of booleans instead of specific strings for each type
(-) current_insurer: does not make sense we can have more than one (eg. NO house + travel) and does not scale

This has been sinced with @sonnyandersson1 

## Why?

Single strings for each bundle combination does not scale.

### DIFF
--- a/src/client/api/quoteBundleSelectors.ts
+++ b/src/client/api/quoteBundleSelectors.ts
@@ -24,6 +24,26 @@ export const getQuotes = (bundle: QuoteBundle | undefined) => {
   return bundle?.quotes ?? []
 }
 
+export const hasHomeContents = (bundle: QuoteBundle) =>
+  bundle.quotes.some(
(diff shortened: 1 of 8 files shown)


guilhermespopolin (commented):
Nicely done
Just fix the TS error and we're good to go :)

guilhermespopolin (approved):

PR #992, opened 2022-05-11
 cleanup

dendoesit:
## What?

Code comments that I did not get the chance to fix. 

## Why?

Cleaning time

### DIFF
--- a/src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
+++ b/src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
@@ -221,9 +221,9 @@ export const CheckoutPayment = ({
     }
   }, [is3DsError, history, trackOfferEvent])
 
-  const addPaymentToCart = useCallback(async () => {
+  const addPaymentToCart = () => {
     setIsPaymentConnected(true)
-  }, [])
+  }
 
   const performCheckout = useCallback(async () => {
     try {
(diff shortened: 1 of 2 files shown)


src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
```diff
@@ -221,9 +221,9 @@ export const CheckoutPayment = ({
     }
   }, [is3DsError, history, trackOfferEvent])
 
-  const addPaymentToCart = useCallback(async () => {
+  const addPaymentToCart = () => {
```
dendoesit, line 224: not sure if this is what you meant in the previous pr @robinandeer

robinandeer: hmm... not sure. Now it seems like we are missing a dependency in the dependency array for `performCheckout`

I think I commented on the fact that the name of the function doesn't represent what it actually does. One way would be to just skip it and call `setIsPaymentConnected(true)` directly inside `performCheckout`, no?

src/client/pages/ConnectPayment/components/useAdyenCheckout.tsx
```diff
@@ -271,21 +272,23 @@ const createAdyenCheckout = ({
         }
 
         const paymentTokenId =
-          result.data.paymentConnection_connectPayment.paymentTokenId
+          result.data?.paymentConnection_connectPayment.paymentTokenId
```
robinandeer, line 275: hmmm why doesn't type-checking pick this up? 🤔

robinandeer (approved):
Looks good :)

PR #993, opened 2022-05-12
Add husky and pre-commit hooks

simonauner:
## What?

- Updates husky to latest version (the one we had didn't seem to work anyway)
- Update lint-staged to run Prettier and ESLint only on staged files

Same setup as in Racoon, basically.

## Why?

Because I like to know that I have errors in my code before it reaches the CI server. Many times I create PR:s only to find out that I have some TSC error somewhere in some code that I didn't touch, and it's a bit annoying. 

The existing setup didn't run any pre-commit hooks for me at least.

**Ticket(s): []**


gustaveen (approved):
Sweet! 👌 I'm happy with running it pre-commit. Sometimes though I want to commit things that is unfinished instead of stashing. So an alternative would be to run it pre-push.

simonauner:
> Sweet! 👌 I'm happy with running it pre-commit. Sometimes though I want to commit things that is unfinished instead of stashing. So an alternative would be to run it pre-push.

Yeah, either works for me. When it want to commit without using hooks I use `git commit -m "stuffs" --no-verify`

package.json
```diff
@@ -84,9 +85,9 @@
     "eslint-plugin-prettier": "^3.1.4",
     "eslint-plugin-react": "^7.21.2",
     "eslint-plugin-react-hooks": "^4.2.0",
-    "husky": "^0.14.3",
+    "husky": "^8.0.1",
```
robinandeer, line 88: woho, jumping 7 major version 😅

robinandeer (approved):
Also agree that we could explore running tests/typecheck only before push 🤔 

But great to align projects!

package.json
simonauner: Can't hurt, right :D ?

simonauner:
Ok, I hear you.

I made a change now so that we run `lint-staged` (Prettier and ESLint) on commit, and `tsc` and `test` on push.

gustaveen:
> When it want to commit without using hooks I use `git commit -m "stuffs" --no-verify`

True that @simonauner! 👌

PR #995, opened 2022-05-12
Fix / Remove auth header from API calls

robinandeer:
## What?

This is a quick fix. Starting to get a little hard to maintain this -- but we can get rid of it when we remove the old offer page :P

## Why?

We currently send "undefined" as a header which makes the API confused.

**Ticket(s): []**


FredrikAreschoug (approved):

dendoesit (approved):
👍

PR #998, opened 2022-05-12
GRW-1021 Fix Ownership Type and tracking problem with insurance toggle

mehdielket:
## What?

* Fix `ownership_type` to be 'own' or 'rent'
* Fix incorrect tracking for `switch_from` value for insurance toggle and cross sell


src/client/api/quoteSelector.ts
```diff
@@ -82,4 +82,6 @@ const HOME_INSURANCE_TYPES = [
 export const isHomeContentsOrHouse = (quote: BundledQuote) =>
   HOME_INSURANCE_TYPES.includes(quote.data.type)
 
-export const getSubType = (quote: BundledQuote) => quote.data.subType
+type QuoteDataSubTypes = 'RENT' | 'OWN' | 'STUDENT_RENT' | 'STUDENT_OWN'
```
robinandeer, line 85: Will this conflict with car that adds additional sub types? Right, @simonauner?

robinandeer (commented):
Nice :)

src/client/utils/tracking/hooks/useTrackOfferEvent.ts
```diff
@@ -31,15 +31,20 @@ export const useTrackOfferEvent = () => {
             locale: isoLocale,
           },
         })
-        const selectedBundleVariant = getSelectedBundleVariant(
-          quoteCartQueryData,
-          selectedInsuranceTypes,
-        )
 
-        if (selectedBundleVariant) {
+        let bundle = selectedBundle
```
marian-buc, line 35: What about
```suggestion
        const bundle= selectedBundle || getSelectedBundleVariant(
          quoteCartQueryData,
          selectedInsuranceTypes,
        )?.bundle

        if (bundle) {
```

marian-buc (approved):

src/client/api/quoteSelector.ts
mehdielket: I can also set it to string for the time being. Since it comes from the data blob it's not typed

src/client/utils/tracking/hooks/useTrackOfferEvent.ts
mehdielket: Fixed

marian-buc:
we can use this to get over the selectedInsuranceTypes race condition and the selectedBundle override 
I've tested it seems to be working fine
`export const getSelectedInsuranceTypes = () =>
  deserializeSearchParams(new URLSearchParams(window.location.search))
  `

marian-buc:
fixed useTrackPageViewEvent to change only when location.pathname changes, ignoring the search params

src/client/utils/tracking/hooks/useTrackPageViewEvent.ts
```diff
@@ -13,6 +13,8 @@ export const useTrackPageViewEvent = () => {
   const market = useMarket().toLowerCase()
   const location = useLocation()
 
+  const [currentPathname, setCurrentPathname] = useState<string>()
```
robinandeer, line 16:
```suggestion
  const [currentPathname, setCurrentPathname] = useState<string>(location.pathname)
```

...but I'm not sure I like this. We should avoid state unless we really care about it. This is a bit murky. Isn't it possible to subscribe to some event that fires every time there's a page navigation? Like:

```
history.listen((location, action) => {
    // location is an object like window.location
    console.log(action, location.pathname, location.state)
});
```

robinandeer (commented):
Nice progress! I think I noticed a small bug tho

And had a comment about avoid React state

src/client/api/quoteSelector.ts
simonauner: It is correct that we use sub-types also for Car, but this doesn't really seem to matter since when Car sends different types in runtime TS won't know about it anyway. 

But if we were to add some logic based on any Car value in the future we need to add it. But as Mehdi says, we're trying to type values in an `any` object anyway, so it's just developer ergonomics 🤷

src/client/utils/tracking/hooks/useTrackPageViewEvent.ts
marian-buc: problem is location object can trigger changes with location.search as well, but we're interested in pathname changes only, so we need a way to keep track of the previous pathname

marian-buc:
Looking good, we'll fix the pageview in a new PR

src/client/utils/tracking/hooks/useTrackPageViewEvent.ts
mehdielket: @robinandeer I reverted this change so this fix will come in a future PR where we can discuss the solution in more details

PR #999, opened 2022-05-12
GRW-1029 / feat / Load FAQ from PCMS

simonauner:
## Why?

In order to launch Car we want to have Car-specific FAQ content on the Offer page. This means that we must get FAQ content based on what type of insurance(s) we're showing on the Offer page, not just a generic one like today. This should also work for all other data types on the offer page, not just Car.

These are the conditions:
- We have a number of bundles on the offer page, some will have different _insurance type_ (SE_CAR, SE_HOME etc), and some will have different _contract type_ (SE_CAR_HALF, SE_CAR_FULL etc)
- FAQs in PCMS are organized by _contract type_
- It's will be the same FAQ content for all of Car's contract types
- We only want one FAQ for Car on the offer page

## What?

We request FAQ:s for all insurance types from Giraffe/PCMS. Giraffe will take care of filtering duplicates (based on title 🙄 , not the best solution, but works for now), and only returning those items that have a title. FAQ items without title are considered not complete for web's sake and will not be returned from Giraffe. When we do not get a response or empty response from PCMS we fetch from GraphCMS instead, so it's a fallback to use old content.

The fallback code makes it look a bit ugly, but hopefully this will be a short-lived transition to have all FAQ content in PCMS.

This gives us the possibility to show multiple tabs of insurances:

https://user-images.githubusercontent.com/1343979/168101978-25d077ea-0d38-4657-89f5-bbf57b3660b8.mov

**Ticket(s): [GRW-1029]**

[GRW-1029]: https://hedvig.atlassian.net/browse/GRW-1029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/pages/OfferNew/FaqSection.tsx
+++ b/src/client/pages/OfferNew/FaqSection.tsx
@@ -4,13 +4,13 @@ import { colorsV3 } from '@hedviginsurance/brand'
 import React from 'react'
 import AnimateHeight from 'react-animate-height'
 import ReactMarkdown from 'react-markdown/with-html'
-import { useFaqsQuery } from 'data/graphql'
-import { getIsoLocale, useCurrentLocale } from 'components/utils/CurrentLocale'
+import { TypeOfContract, useContractFaqQuery } from 'data/graphql'
 import { useTextKeys } from 'utils/textKeys'
 import {
   LARGE_SCREEN_MEDIA_QUERY,
   MEDIUM_SCREEN_MEDIA_QUERY,
 } from 'utils/mediaQueries'
+import { useCurrentLocale } from 'l10n/useCurrentLocale'
 import { Container, Heading, Column, ColumnSpacing } from './components'
 
 const SectionWrapper = styled.div`
@@ -125,7 +125,6 @@ const ExpanderIcon = styled.svg<Openable>`
 `
 
 interface AccordionProps {
-  id: string
   headline: string
   body: string
 }
@@ -151,11 +150,17 @@ export const Accordion: React.FC<AccordionProps> = ({ headline, body }) => {
   )
 }
 
-export const FaqSection: React.FC = () => {
-  const pathLocale = useCurrentLocale()
-  const language = getIsoLocale(pathLocale)
-  const faqs = useFaqsQuery({ variables: { language } })
-  const languageData = faqs?.data?.languages[0]
+type FaqSectionProps = {
+  typeOfContract: TypeOfContract
+}
+
+export const FaqSection = ({ typeOfContract }: FaqSectionProps) => {
+  const { isoLocale } = useCurrentLocale()
+  const { data: faqItems } = useContractFaqQuery({
+    variables: { locale: isoLocale, contractType: 'SE_CAR_HALF' },
+  })
+  // const faqs = useFaqsQuery({ variables: { language } })
+  // const languageData = faqs?.data?.languages[0]
   const textKeys = useTextKeys()
 
   return (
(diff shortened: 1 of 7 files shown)


src/client/pages/OfferNew/FaqSection.tsx
```diff
@@ -151,28 +151,60 @@ export const Accordion: React.FC<AccordionProps> = ({ headline, body }) => {
...
-  const language = getIsoLocale(pathLocale)
-  const faqs = useFaqsQuery({ variables: { language } })
-  const languageData = faqs?.data?.languages[0]
+type FaqPanelProps = {
+  quote: BundledQuote
+}
+
+const FaqPanel = ({ quote }: FaqPanelProps) => {
+  const { isoLocale } = useCurrentLocale()
+  const { data: faqItems } = useContractFaqQuery({
```
simonauner, line 160: Hopefully we can make an endpoint in Giraffe that gets all FAQ content at once, so we can call it in `FaqSection` instead

src/client/pages/OfferNew/FaqSection.tsx
```diff
@@ -121,11 +122,10 @@ const ExpanderIcon = styled.svg<Openable>`
   flex-shrink: 0;
   width: 0.8rem;
   height: 0.8rem;
-  fill: ${colorsV3.white};
+  fill: ${colorsV3.black};
```
gustaveen, line 125: nit. Should probably be `gray900`

src/client/pages/OfferNew/FaqSection.tsx
gustaveen: That would be nice!

gustaveen (commented):
Looks great Simon! Makes sense to split it up in tabs to not have long list with FAQ

simonauner:
Companion PR: https://github.com/HedvigInsurance/giraffe/pull/1596

simonauner:
This PR was enabled by:
- Adding a new property `title` in PCMS and returning the entire data collection, not just an array of `{body: string, headline: string}` objects: https://github.com/HedvigInsurance/promise-cms/pull/105 +  https://github.com/HedvigInsurance/promise-cms/pull/107
- Adding a new corresponding endpoint in giraffe: https://github.com/HedvigInsurance/giraffe/pull/1620

src/client/pages/OfferNew/FaqSection.tsx
```diff
@@ -151,33 +138,99 @@ export const Accordion: React.FC<AccordionProps> = ({ headline, body }) => {
...
+    },
+  })
+
+  const tabItems = React.useMemo(() => {
+    if (!data || !data.contractFaqs?.faqs?.length) return null
+
+    return data.contractFaqs.faqs.map((d) => ({
+      id: d.title || '',
+      // if only one FAQ: show no tab (empty string for 'name')
+      name: data.contractFaqs?.faqs?.length ?? 0 < 2 ? '' : d.title!,
```
robinandeer, line 187: What is happening here? 😅  maybe change to:
```suggestion
      name: (data.contractFaqs?.faqs?.length ?? 0) < 2 ? '' : d.title!,
```

robinandeer (approved):
Nice nice, but let's push for moving all this to PCMS soon :D

src/client/pages/OfferNew/FaqSection.tsx
simonauner: Isn't it crystal clear :D ? 

I agree it's messy!

PR #1001, opened 2022-05-13
GRW-1060 Update adtraction tracking to support new setup

mehdielket:
## What?

Update Adtraction tracking to use `product_categories` instead of specifc `tp` values for every bundle combination

* Change Tp value to a special value per market: 
* Change the am (order_value) to 0
* In the pc (product_categories) build a string of the category ids of the promises in the bundle. A matching between the category ids and promise types will be provided by Sonny Andersson 

The category ids should be concatenated as described in the adTraction docs with amount 0

## Why?

The existing solution was not scalable

**Ticket(s): [GRW-1060]**

[GRW-1060]: https://hedvig.atlassian.net/browse/GRW-1060?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/api/quoteBundleSelectors.ts
```diff
@@ -98,6 +98,10 @@ export const getMainQuote = (bundle: QuoteBundle) => {
   return bundle.quotes[0]
 }
 
+export const getEmail = (bundle: QuoteBundle) => {
+  return bundle.quotes[0].email
```
robinandeer, line 102: nit: should this compose other selectors like: `getMainQuote` instead of assuming things?

robinandeer (approved):
Great work, this is super complicated tho - we should do some work when moving this into Racoon to clean up :)

src/client/api/quoteBundleSelectors.ts
robinandeer: I agree, I guess we solve it better in the new acquisition API:s

guilhermespopolin (approved):

src/client/utils/tracking/adtraction/adtractionQuoteCart.ts
```diff
@@ -1,38 +1,52 @@
 import md5 from 'md5'
 import { QuoteBundle } from 'data/graphql'
-import { quoteBundleTrackingContractType } from 'api/quoteBundleTrackingContractType'
+import { AdTractionMarketConfig } from 'src/client/l10n/adTractionConfigs'
```
robinandeer, line 3:
```suggestion
import { AdTractionMarketConfig } from 'l10n/adTractionConfigs'
```

PR #1004, opened 2022-05-16
feat/useTrackSignedCustomerEvent fix

marian-buc:
## What?
Added getSelectedInsuranceTypes and getCurrentVariation function calls in the tracking handler instead of using their respective hooks

## Why?
To prevent any race conditions where the handler might be called before react updates the internal hook state, same thing we did with useTrackOfferEvent

**Ticket(s): []**

### DIFF
--- a/src/client/utils/hooks/useVariation.ts
+++ b/src/client/utils/hooks/useVariation.ts
@@ -16,19 +16,24 @@ const getVariation = match([
   [match.any(), null],
 ])
 
-export const useVariation = (): Variation | null => {
+const parseVariaton = (search: string): Variation | null => {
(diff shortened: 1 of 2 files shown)


mehdielket (approved):
🚀

robinandeer (approved):
This means that we lose the reactivity on "variation", right?

I guess it's still okey since we don't expect it to change

marian-buc:
@robinandeer no, the handler will now always be called with the latest variation, we just stop depending on the hook for it. same thing happened with selectedInsuranceTypes, which is supposed to change, but react would first fire the handler and only then update the callback.

PR #1006, opened 2022-05-17
GRW-1054 / feat / Adapt Offer page promo sections

simonauner:
## What?

This PR makes the `AppPromotionSection` configurable from the outside, enabling us to pass in different content based on which quote types that are on display.

Once more, it's not the most scalable implementation. For future needs this should perhaps be added to the PCMS?

I copied the `Space` component from Racoon since it's very useful in scenarios like this, when I need to add space between components when there's more than one.

The images are to be updated as well, but that can be in a separate PR.

## Why?

The promotions are different when a car quote is on display, and more importantly there should be two promotions and not just one.

**Ticket(s): [GRW-1054]**

image

[GRW-1054]: https://hedvig.atlassian.net/browse/GRW-1054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/pages/Offer/index.tsx
```diff
@@ -237,7 +241,34 @@ export const OfferPage = ({
...
+                ]
+              : [
+                  {
+                    headingText: textKeys.OFFER_APP_PROMOTION_HEADING(),
+                    bodyText: textKeys.OFFER_APP_PROMOTION_PARAGRAPH(),
+                    imageUrl: '/new-member-assets/offer/app-promotion.jpg',
+                    imageAltText: 'App promotion',
+                  },
+                ]
+          }
```
robinandeer, line 270: nit: extract into a custom hook

src/client/pages/Offer/index.tsx
```diff
@@ -237,7 +241,34 @@ export const OfferPage = ({
...
-        <AppPromotionSection />
+        <AppPromotionSection
+          promotions={
+            hasCarInBundles
+              ? [
+                  {
+                    headingText: textKeys.OFFER_SWITCH_PROMOTION_HEADING_CAR(),
+                    bodyText: textKeys.OFFER_SWITCH_PROMOTION_PARAGRAPH_CAR(),
+                    imageUrl: '/new-member-assets/offer/app-promotion.jpg',
+                    imageAltText: 'App promotion',
```
robinandeer, line 252: Is this image decorational? In that case it should have an empty alt text - doesn't seem like this adds much value to screen, no?

https://www.w3.org/WAI/tutorials/images/decorative/

robinandeer (approved):
Great stuff

src/client/pages/Offer/index.tsx
simonauner: Ah, nice!

src/client/pages/Offer/index.tsx
simonauner: Fixed!

robinandeer (approved):
Very nice with the new hook!

PR #1014, opened 2022-05-19
GRW-1111 Fix details modal & checkout form when House NO

mehdielket:
## Why?

**Ticket(s): []**

### DIFF
--- a/src/client/api/quoteBundleSelectors.ts
+++ b/src/client/api/quoteBundleSelectors.ts
@@ -70,19 +70,15 @@ export const isYouthOffer = (bundle: QuoteBundle): boolean => {
     .every((quote) => quote.data.isYouth === true)
 }
 
-const HOME_HOUSE_INSURANCE_TYPES: Array<InsuranceType> = [
-  InsuranceType.DANISH_HOME_CONTENT,
-  InsuranceType.NORWEGIAN_HOME_CONTENT,
-  InsuranceType.NORWEGIAN_HOUSE,
-  InsuranceType.SWEDISH_APARTMENT,
-  InsuranceType.SWEDISH_HOUSE,
-]
-
 export const getMainQuote = (bundle: QuoteBundle) => {
   if (isMultiQuote(bundle)) {
-    const mainQuoteInBundle = bundle.quotes.find((quote) =>
-      HOME_HOUSE_INSURANCE_TYPES.includes(quote.data.type),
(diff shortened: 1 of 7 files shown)


src/client/api/quoteBundleSelectors.ts
```diff
@@ -70,19 +70,15 @@ export const isYouthOffer = (bundle: QuoteBundle): boolean => {
...
   if (isMultiQuote(bundle)) {
-    const mainQuoteInBundle = bundle.quotes.find((quote) =>
-      HOME_HOUSE_INSURANCE_TYPES.includes(quote.data.type),
+    const houseQuote = bundle.quotes.find((quote) =>
+      quoteSelector.isHouse(quote),
+    )
+    const homeContentsQuote = bundle.quotes.find((quote) =>
+      quoteSelector.isHomeContents(quote),
     )
+    const mainQuoteInBundle = houseQuote || homeContentsQuote
```
robinandeer, line 81: Does this work with Car?

src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
```diff
@@ -378,12 +371,7 @@ export const CheckoutPayment = ({
               currentInsurer: currentInsurer?.id,
               phoneNumber: phoneNumber?.replace(/\s/g, ''),
               dataCollectionId,
-              data: {
-                ...form.data,
-                type,
-                typeOfContract,
-                isStudent,
-              },
+              data,
```
robinandeer, line 374: This is very nice!

robinandeer (approved):
Great work @mehdielket!

This is growing harder and hard to maintain but I still think it's probably better to focus on building new things in Racoon rather than refactor this too much :P

Just had one question if this is compatible with Car

src/client/api/quoteBundleSelectors.ts
simonauner: Not sure what this is used for, I don't think I've used the concept of main quote myself, but perhaps Marian has? Try it out with quote `e99c3e62-fee1-4824-b802-20ba80b61bfc`

mehdielket: Well, this did not work with Car in the first place right?

simonauner: For future reference it can be good to know that car offers are not `isMultiQuote` so for them it will just `return bundle.quotes[0]`

PR #1015, opened 2022-05-19
 added new names for sqm and baths

dendoesit:
## What?

Added the new values for House NO in CHeckout Details Summary 

## Why?

Missing 

** Could not test.


src/client/pages/OfferNew/types.ts
```diff
@@ -73,7 +73,9 @@ export type GenericQuoteData = {
   subType?: string
   apartment?: string
   floor?: number
+  squareMeters?: number
 
+  numberOfWetUnits?: number
   ancillaryArea?: number
   numberOfBathrooms?: number
   yearOfConstruction?: number
```
guilhermespopolin, line 81: House NO quote creation also involves providing `yearOfOwnership`. I'm not sure if we should also display that tho

guilhermespopolin (approved):

src/client/pages/OfferNew/types.ts
dendoesit: good point. I will ask @zakkjellin about this.

PR #1017, opened 2022-05-19
GRW-1041 / feat / Add HeroBackgroundImage component

simonauner:
## What?

Copy the `BackgroundImage` component into a `HeroBackgroundImage` component since we need to style the header on large desktop screen differently to always get it centered in the image. And if we do that to the `BackgroundImage` it will break for `LandingPage`.

## Why?

Without this, the Car in the photo is not in view for large desktop:

https://user-images.githubusercontent.com/1343979/169268892-4e4668a6-f76a-4be4-9d4d-7ccdc90989a8.mov

After this change:

https://user-images.githubusercontent.com/1343979/169268933-75b31e69-39c8-4c3e-a6df-1c5362abfbf3.mov

And for reference, I think it looks better for home insurance as well.

Before: 

https://user-images.githubusercontent.com/1343979/169268990-9d5394ed-92fe-40c0-bb75-509cb62a8dc6.mov

After:

https://user-images.githubusercontent.com/1343979/169269009-3b958013-808b-41ea-8908-998198ac7cca.mov

**Ticket(s): [GRW-1041]**

[GRW-1041]: https://hedvig.atlassian.net/browse/GRW-1041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gustaveen (approved):
LGTM! I suppose it was too messy to pass something like object-position to the background component?

simonauner:
> LGTM! I suppose it was too messy to pass something like object-position to the background component?

Didn't even consider it actually, it felt too complicated with breakpoints and all, but perhaps it's not too messy?

PR #1018, opened 2022-05-19
Bugfix/edit details error inside checkout Details.

dendoesit:
## What?

Removing the throw inside selected date as it breaks the checkout details page in CPAS. 

## Why?

Right now if you edit a value inside /checkout/details edit modal the page breaks as the Select Date component cannot find a quote for some reason. 

## Testing

Go to /checkout/details and open the edit information modal and update one of the values. 

ping @robinandeer, do you remember if there is a specific reason this was implemented like this ?


guilhermespopolin (approved):
Nice

PR #1019, opened 2022-05-20
Center offer hero bg image with flexbox

simonauner:
## What?

This PR partly reverts what PR https://github.com/HedvigInsurance/web-onboarding/pull/1017 introduced. That PR added a `HeroBackgroundImage` because it became difficult to adjust the breakpoints with `object-position` that worked both for `LandingPage` and `OfferPage`. But then I did some more thinking, and thought, why not just center the image with flex-box? It seems to work for all use cases.

So, now we can remove `HeroBackgroundImage` and use `BackgroundImage` like it was used before https://github.com/HedvigInsurance/web-onboarding/pull/1017 was merged. So even though it looks like much has happened here with files being removed and changed and all, it's perhaps easier to just compare how it looked before https://github.com/HedvigInsurance/web-onboarding/pull/1017 and now, which can be done here: 
https://github.com/HedvigInsurance/web-onboarding/compare/f746569b15a5015a6406d4800dd724dc707952ca...e3e82c8e90f11084e39aaeb39f4a78c12939636f

Considering the diff above, this has been added:
* Center image in `BackgroundImage` with flex-box instead of trying to center with `object-position`
* Remove unused `height` property, height is set by styling `BackgroundImage` instead

## Why?

https://github.com/HedvigInsurance/web-onboarding/pull/1017 seemed like a way too complicated solution to the problem. This way we can still have one component and it will be centered for all resolutions.

**Ticket(s): [GRW-1041]**

Car offer page:

https://user-images.githubusercontent.com/1343979/169468429-987dec01-61c1-4d17-844d-04cac749ccb0.mov

Home insurance offer page:

https://user-images.githubusercontent.com/1343979/169468457-41543e6d-b3b3-4370-a5ff-7fa89c0e0075.mov

Landing page:

https://user-images.githubusercontent.com/1343979/169468480-1f1eb895-43ea-4c0d-b7dc-ad6ab9fc19cc.mov

[GRW-1041]: https://hedvig.atlassian.net/browse/GRW-1041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robinandeer (approved):
Very nice! Looks great :D

PR #1020, opened 2022-05-20
fix(GRW-1116): update available options

guilhermespopolin:
## What?

Filter out ExtraBuildings options list on Norway so it matches the one we show during the Embark flow

## Why?

Not all options are available in Norway so there is no point on displaying all of them

**Ticket(s): [GRW-1116](https://hedvig.atlassian.net/browse/GRW-1116)**

## Screenshots / recordings 

image


dendoesit (approved):

mehdielket (approved):
🚀

PR #1025, opened 2022-05-23
Fix / Offer page small fixes car

amandaflod:
## What?
GRW-1126: Compare button text should change size as body text above does, when on small screen
GRW-1127: Align labels for start date in checkout panel
GRW-1123: Phone opening hours text should be grey100

## Why?

**Ticket(s): []**

Screenshot 2022-05-23 at 13 15 07
Screenshot 2022-05-23 at 13 15 31
Screenshot 2022-05-23 at 13 16 07


robinandeer (approved):
Nice clean up :D

PR #1028, opened 2022-05-24
Remove / Forever page

robinandeer:
## What?

Remove unused helpers.
Remove unused GraphQL definitions.
Updates needed for DK House (new from GraphQL).

> I realized that we don't have an Adtraction ID for NO House yet we are live -- do you know about that @guilhermespopolin?

## Why?

We have moved it to Racoon.
DK house updates needed for type checking.

**Ticket(s): []**


simonauner (approved):
🧹 nice! Less code to maintain, and also makes the shift towards Racoon more visible!

PR #1030, opened 2022-05-24
GRW-1131 / feat / Add switcher promo image for offer page

simonauner:
## What?

Adds a new image for the second promo block on Offer page that shows up when a Car quote is displayed on the page.

## Why?

The two blocks had the same image previously, but of course that's not how the design is.

[Figma design for reference](https://www.figma.com/file/HEQXHw3ptHZTltJz38hZ9T/Car-Insurance-%C2%B7-EXPLORATION?node-id=1344%3A19744)

**Ticket(s): [GRW-11-31]**

Screenshot 2022-05-24 at 08 13 25


robinandeer (approved):
Great!

Should we symlink the images since they are all the same 😅

simonauner:
> 

Yes, good idea! Never thought of that! Symlinked now 👍

PR #1033, opened 2022-05-25
Fix/consistent annual mileage translation

simonauner:
## What?

Use same in translation in Checkout component as in Edit modal for annual mileage.

## Why?

Edit modal would say "10,000 km per year" and Checkout component would say "1,000 mil/year". This makes it consistent.

image

**Ticket(s): []**


gustaveen (approved):
🛣

PR #1038, opened 2022-05-27
Fix / Add optional check when getting quote cart id

robinandeer:
## What?

See title.

Update Node engine to allow more recent versions of Node.

## Why?

Saw an error related to "data.data" being null from Sentry.

**Ticket(s): []**


src/server/middleware/quoteCart.ts
```diff
@@ -85,7 +85,13 @@ export const quoteCartSessionMiddleware: Router.IMiddleware<
       variables: { market: apiMarket, locale: isoLocale },
     })
 
-    const newQuoteCartId = result.data.data.onboardingQuoteCart_create.id
+    const newQuoteCartId = result.data.data?.onboardingQuoteCart_create.id
```
simonauner, line 88: Just a nit: `result.data` could be destructured on line 76 to `const { data } = ...` since we don't seem to use anything else from `result`

simonauner (approved):
Nice, simple fix 👍

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "engines": {
-    "node": "14.17",
+    "node": ">=14.17",
```
robinandeer, line 3: Does anyone think this is a bad idea? 

/ping @HedvigInsurance/purchase-journey

simonauner: I think if we start using features that are supported only in a higher version it could become troublesome, no? Like, what if I code something that's only supported in 16, but then someone else is still lagging behind on 14, and we allow both... If we are strict with which (major) version we're running it's easier to stay consistent between team members.

But it's nice to move up to a later LTS of course 👍 But perhaps we should use `>=16.x` in that case? If supported by Heroku.

robinandeer: Ahh yes good point. Could be an idea to upgrade all at once.

But I believe if we would introduce a Node 16-specific feature, that's the time to update supported Node versions.

But as long as we use Node 14 features, why should we not allow Node 16?

simonauner: > But as long as we use Node 14 features, why should we not allow Node 16?

True, we should be able to do that! 

I was just thinking that from a runtime point of view, if I were to be on node 16, develop something that only works in 16, and then push it without testing on 14, I wouldn't know it's not working on 14 until it's either deployed or run by someone else. I would probably be a bit frustrated that `nvm use` would allow me to run 16 but then get problems when I deploy it.

simonauner: (and maybe this is just a highly hypothetical discussion, I don't even know which features I would use that are only available from 16+)

robinandeer: haha yeah I know, I don't don't know either

but I think you make a good point, I will revert those changes and update to Node across deployment/development in one go!

simonauner (approved):

PR #1039, opened 2022-05-27
feat / GRW-1154 / UpperCase RegNo field in edit modal

marian-buc:
## What?
Uppercased the registration number field in the edit edit modal by adding an upperCase prop to CoreInputFieldProps and using it in formik onChange

## Why?
To uppercase the reg no field and it's value, without using a mask sanitize function which would be impractical for this use

**Ticket(s): [GRW-1154]**

[GRW-1154]: https://hedvig.atlassian.net/browse/GRW-1154?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


robinandeer (approved):
Nice!

Did you consider just overriding the specific text input onChange handler? I worry sometimes that we keep introducing props before we know if they will be reused or just needed in one place.

Anyways, nice solution

PR #1040, opened 2022-05-27
Fix / Keep checkout payment page mounted on error

robinandeer:
## What?

Keep checkout payment page mounted while showing error.

## Why?

If we remove Adyen dropin component it doesn't redirect the user.
If we have data we should show the checkout page.

**Ticket(s): []**

### DIFF
--- a/src/client/pages/Checkout/CheckoutDetails/CheckoutDetails.tsx
+++ b/src/client/pages/Checkout/CheckoutDetails/CheckoutDetails.tsx
@@ -42,7 +42,7 @@ export const CheckoutDetails = () => {
   const { path: localePath } = useCurrentLocale()
 
   const trackOfferEvent = useTrackOfferEvent()
-  const data = useQuoteCartData()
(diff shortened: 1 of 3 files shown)


src/client/pages/Checkout/CheckoutDetails/CheckoutDetails.tsx
```diff
@@ -105,6 +106,8 @@ export const CheckoutDetails = () => {
       >
         <PaymentInfo {...priceData} />
       </Footer>
+
+      <CheckoutErrorModal isVisible={error !== undefined} onRetry={onRetry} />
```
simonauner, line 110: A bit late to the party here, but if the server returns `error: null` the modal will not show - is that intended? Maybe the server never sends that?

robinandeer: Hmmm at least it's typed as `ApolloError | undefined` 🤔

PR #1041, opened 2022-05-30
GRW-1094 / Comparison table accordion

amandaflod:
## What?
Accordion in the comparison table that shows the peril description when open. DESKTOP ONLY. 

## Why?

**Ticket(s): [GRW-1094]**

https://user-images.githubusercontent.com/59652803/170940907-6bee36e7-cae0-4cf5-a910-fbf398fd32f2.mov

[GRW-1094]: https://hedvig.atlassian.net/browse/GRW-1094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/pages/OfferNew/ComparisonTable/ComparisonTable.tsx
```diff
@@ -13,13 +15,75 @@ import {
...
+type AccordionProps = {
+  bundles: QuoteBundle[]
+  row: PerilV2
+  isActive: boolean
+  setIsActive: (isActive: boolean) => void
+}
+
+const Accordion = ({ bundles, row, isActive, setIsActive }: AccordionProps) => {
+  return (
+    <AccordionRow key={row.title} onClick={() => setIsActive(!isActive)}>
```
robinandeer: If this element is clickable it should be a button or have a role of a button and a focus state for accessibility reasons :)

But feel free to iterate on that in a future PR!

robinandeer (approved):
Looks great!

I think a couple of things to think about in future iterations:

- Can we make the base accordion more generic so we could e.g. use it also for rendering FAQs?
- Consider supporting full accessibility (more important than animations IMO 😉)
  - http://web-accessibility.carnegiemuseums.org/code/accordions/
  - https://w3c.github.io/aria-practices/examples/accordion/accordion.html
  - OR just implement/style on top of: https://www.radix-ui.com/docs/primitives/components/accordion

package.json
```diff
@@ -76,7 +76,6 @@
     "babel-plugin-dynamic-import-node": "^2.1.0",
     "babel-plugin-emotion": "^10.0.23",
     "bufferutil": "^4.0.1",
-    "caniuse-lite": "^1.0.30001099",
```
robinandeer, line 79: Hmm where does this change come from? Mention in PR description

robinandeer:
@amandaflod sorry I was to quick to review! 😅 

One comment I have on the updated approach is that what you call `Accordion` now is more like an `AccordionItem`, no?

simonauner:
> @amandaflod sorry I was to quick to review! 😅
> 
> One comment I have on the updated approach is that what you call `Accordion` now is more like an `AccordionItem`, no?

Good point. Maybe an `Accordion` component should just set up a context and hold which `AccordionItem` that is currently selected? I think that could be a scalable solution.

package.json
amandaflod: Hm... I only did a yarn install and got this 🤔 Should I remove the changes here as well as the ones on yarn.lock? Because I honestly don't know what they will do 😅

amandaflod:
Thank you guys!! And sorry @robinandeer, I should have changed it to draft as soon as Simon and I started changing so much stuff 😅

amandaflod:
Oh and @simonauner, I'm am not quite sure I understand the Accordion context idea, if we want to implement it here I would need your help describing it! :)

simonauner:
> Oh and @simonauner, I'm am not quite sure I understand the Accordion context idea, if we want to implement it here I would need your help describing it! :)

No, it's out of context (😂 ) for this PR! We can do it together and refactor some time in the future! 🙌

simonauner (approved):

package.json
robinandeer: hmm yeah I wonder why that is -- yarn install doesn't change package.json as far as I know

PR #1043, opened 2022-05-30
Fix/GRW-1118/StartDate getLocaleImport

marian-buc:
##What?
fixed getLocaleImport switch statement

##Why?
since locale is actually en_SE this was breaking translations across the app

**Ticket(s): [GRW-1118]**


dendoesit (approved):
👍

PR #1044, opened 2022-05-30
 tracking for offer sidebar

dendoesit:
## What?

Added tracking for sidebar

## Why?

So we can track HOUSE NO 

https://hedvig.atlassian.net/browse/GRW-1163

### DIFF
--- a/src/client/pages/Offer/Checkout/index.tsx
+++ b/src/client/pages/Offer/Checkout/index.tsx
@@ -39,6 +39,8 @@ import { useSelectedInsuranceTypes } from 'utils/hooks/useSelectedInsuranceTypes
 import { QuoteInput } from 'components/DetailsModal/types'
 import { useTrackSignedCustomerEvent } from 'utils/tracking/hooks/useTrackSignedCustomerEvent'
 import { useDebounce } from 'utils/hooks/useDebounce'
+import { useTrackOfferEvent } from 'utils/tracking/hooks/useTrackOfferEvent'
+import { EventName } from 'utils/tracking/gtm/types'
 import { apolloClient as realApolloClient } from '../../../apolloClient'
 import { isSsnInvalid, checkIsManualReviewRequired } from '../../Checkout/utils'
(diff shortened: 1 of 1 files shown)


src/client/pages/Offer/Checkout/index.tsx
```diff
@@ -429,13 +441,18 @@ export const Checkout = ({
...
+        eventName: EventName.SignError,
+      })
       setSignUiState('FAILED')
       return
     }
   }
 
   const handleSignStart = () => {
+    trackOfferEvent({
+      eventName: EventName.ButtonClick,
```
robinandeer, line 454: shouldn't we add?

```
{
    options: { buttonId: 'SOME_BUTTON_ID' },
}
```

robinandeer (commented):
Nice updates!

I guess we don't know the categories for the sign errors?

src/client/pages/Offer/Checkout/index.tsx
dendoesit: hmm, we could do. but we only have on button in offer sidebar 🤔

robinandeer: hehe well in the tracking tool it will just say "button click" no? 

So maybe `buttonId` could just be `sidebar_cta` or similar?

src/client/pages/Offer/Checkout/index.tsx
```diff
@@ -429,13 +441,19 @@ export const Checkout = ({
         setSignUiState('MANUAL_REVIEW')
         return
       }
-
+      trackOfferEvent({
+        eventName: EventName.SignError,
+      })
```
robinandeer, line 446: Sorry for dragging on but I just wonder why we don't include error type here? Like

```
trackOfferEvent({
  eventName: EventName.SignError,
  options: { error, errorType: ErrorEventType.ManualReviewRequired },
})
```

Seems like if we move it inside the if-statement that would work, no?

robinandeer (approved):
Nice! Thanks for adding the error type

PR #1045, opened 2022-05-30
GRW-1164 / Feature / Add Datadog actions to log checkout

robinandeer:
## What?

Log critical actions to Datadog with some important context.

Next up would be to add logging for errors.

## Why?

It's very hard to follow what's happening in the app currently.

With these changes we will be able to get more context in the logs:

Screenshot 2022-05-30 at 16 21 47

**Ticket(s): []**


simonauner (commented):
Really nice! It will be great to see more of what's happening. If we get decent consistent numbers on these we can probably add alerts as well so see when we accidentally break things.

Should we perhaps have constants for event names? I see that some are using GTM's event names.

simonauner (approved):

PR #1051, opened 2022-06-01
GRW-1178-1172-1173 Fix issues with isStudent in edit modal

mehdielket:
## What?

* First issue is that we choose house insurance MainQuote when it is there, but house insurance does not include the "isStudent" data. That is why the form shows isStudent as "no" even if it is a yes. The solution here is to look through the whole bundle for an insurance that has isStudent.
* Second issue is that we discard form values that are boolean and set to false, so there was no way to set isStudent from true to false

## Why?

Users should be able to edit the student status in the edit details modal

**Ticket(s): [GRW-1178] [GRW-1172] [GRW-1173]**

[GRW-1178]: https://hedvig.atlassian.net/browse/GRW-1178?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GRW-1172]: https://hedvig.atlassian.net/browse/GRW-1172?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[GRW-1173]: https://hedvig.atlassian.net/browse/GRW-1173?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/api/quoteBundleSelectors.ts
+++ b/src/client/api/quoteBundleSelectors.ts
@@ -146,6 +146,14 @@ export const hasCar = (quotes: BundledQuote[]) => {
   return quotes.some(quoteSelector.isCar)
 }
 
+export const isStudent = (quotes: BundledQuote[]) => {
+  return quotes.some(quoteSelector.isStudent)
+}
+
+export const isYouth = (quotes: BundledQuote[]) => {
+  return quotes.some(quoteSelector.isYouth)
+}
+
 const getHouseholdSizeFromBundledQuotes = (
   quotes: ReadonlyArray<BundledQuote>,
 ): number | undefined => {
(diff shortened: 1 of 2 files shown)


marian-buc (approved):
LGTM, thanks for the cleanup

robinandeer (approved):
Looks good to me :)

PR #1053, opened 2022-06-01
fix(GRW-1179): only converte main quote subtype for SE

guilhermespopolin:
## What?

Don't try to convert main quote subtype for non-Swedish quotes

## Why?

For DK we don't convert subtypes to indicate student insurance and we do for SE (E.g subType: RENT --> subType: STUDENT_RENT). Instead we rely on a isStudent field

**Ticket(s): [GRW-1179](https://hedvig.atlassian.net/browse/GRW-1179)**

I know, I know. Another _market if statement_. It sucks but I couldn't find any other way 🤷🏻‍♂️


robinandeer (approved):
Yeah I think the mistake was to handle all markets/forms in the same function so it will be hard to solve without market checks :P

PR #1055, opened 2022-06-01
GRW-1166 / Fix / Remove youth/student option for house bundles

robinandeer:
## What?

Remove option to pick student/youth insurance if bundle contains House insurance.

## Why?

Disucssed with @mehdielket and @guilhermespopolin. We don't have support for rendering these forms based on anything by what you main quote is. Therefore it makes sense to disable this possibility which is only available conditionally.

If we want to support this we should refactor how these forms are build up.

**Ticket(s): []**


guilhermespopolin (approved):

robinandeer:
We should first deploy the Embark changes where we don't ask the "student/youth" question for people that sign House insurance, check with @guilhermespopolin :)

PR #1057, opened 2022-06-01
Fix / Getting insurance summary values

robinandeer:
## What?

Avoid nullish checks since 0 evaluates to false.

Screenshot 2022-06-01 at 15 37 21

## Why?

Some values are missing for House.

**Ticket(s): []**


simonauner (approved):
🎉

PR #1058, opened 2022-06-01
Fix / Avoid defaulting to existing co-insured for falsy values

robinandeer:
## What?

Fix edit details form when editing household size / number co-insured.

Before we defaulted to the current value when `householdSize - 1` was falsy. Since 0 is falsy in JS it meant that you couldn't change to `householdSize = 1` 🙃 

## Why?

You can't set household size to 1 in the edit details form.

**Ticket(s): []**


guilhermespopolin (approved):
The famous falsy thing.

dendoesit (approved):
👍

PR #1060, opened 2022-06-02
Redirect user to download page if payment is already connected

simonauner:
## What?

* Redirect users to Download page if they have an active payment method

Per Robin's suggestion the first step is to just redirect the user straight to the Download page. Second step will be to introduce a new page that is tailored specifically to someone signing another contract.

## Why?

* There's no need to connect payment if user already have an active payment method

**Ticket(s): [GRW-1152](https://hedvig.atlassian.net/browse/GRW-1152)**


guilhermespopolin (approved):
👍🏻

PR #1062, opened 2022-06-02
GRW-1152 / feat / Duplicate Download page into a Confirmation page

simonauner:
## What?

Duplicate Download page into a Confirmation page and redirect already connected users to it instead of the Download page.

In order to avoid a slight case of prop drilling I moved the `Headline` component straight into the `DownloadPageContent` component. It didn't seem like much of an abstraction anyway and it looked a bit messy to just pass props.

Translations are still TBD (made by me for now).

## Why?

We want to show something more appreciative when users have added another insurance, instead of just telling them to download the app (which they probably have already done).

**Ticket(s): [GRW-1152](https://hedvig.atlassian.net/browse/GRW-1152)**

### `/confirmation`

image

### `/download` 

Looks just as before:

image


src/client/pages/ConnectPayment/sections/ConnectPayment.tsx
```diff
@@ -133,6 +131,14 @@ const InstructionItemSeparator = styled.div`
   height: 0.5rem;
 `
 
+const getPaymentAlreadyConnectedUrl = ({
```
dendoesit, line 134: not sure if this is correct. Adyen is the one used in DK and NO, in SE we need to use Trustly so changes should be done there. Car is still only available in SE right?

src/client/pages/Download/components/DownloadPageContent.tsx
```diff
@@ -69,7 +68,36 @@ const ImageSection = styled.div`
...
+  font-size: 2rem;
+  line-height: 40px;
+  color: ${colorsV3.gray900};
+
+  ${LARGE_SCREEN_MEDIA_QUERY} {
+    margin: 0;
+    display: flex;
+    flex-direction: column;
+    font-size: 3rem;
+    line-height: 56px;
```
dendoesit, line 82:
```suggestion
    line-height: 1.16;
```

src/client/pages/ConnectPayment/sections/ConnectPayment.tsx
simonauner: Car is only available for Sweden, yes, but the functionality should be correct for all markets - if payment is already connected there's no need to show the "Connect payment page".. or?

The GraphQL endpoint returns `"ACTIVE`". 

Fredrik Areschoug wrote:
> So many threads so little time. But yeah use `payinMethodStatus` which takes care of both adyen and trustly :)

src/client/pages/Download/components/DownloadPageContent.tsx
```diff
@@ -69,7 +68,36 @@ const ImageSection = styled.div`
   }
 `
 
-export const DownloadPageContent: React.FC = () => {
+const Headline = styled.h1`
+  width: 100%;
+  font-size: 2rem;
+  line-height: 40px;
```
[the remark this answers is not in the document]
simonauner: Nice, it's just moved from the other file, but I'll fix it 👍

src/client/pages/Confirmation/index.tsx
```diff
@@ -0,0 +1,34 @@
...
+export const Confirmation = () => {
+  const textKeys = useTextKeys()
+
+  return (
+    <Page>
+      <SessionTokenGuard>
+        <>
+          <Helmet>
+            <title>{textKeys.CONFIRMATION_PAGE_TITLE()}</title>
+          </Helmet>
```
robinandeer, line 18: I guess we don't need this if we handle it serverside, right?

robinandeer (approved):
Looks good :)

src/client/pages/Confirmation/index.tsx
simonauner: Yeah, I guess `app.use(helmet)` takes care of that on the server..? We can even remove `<title>` since the server renders it for us!

src/client/pages/ConnectPayment/sections/ConnectPayment.tsx
simonauner: @dendoesit please get back to me on the above, I wouldn't want to break anything and you know more about payments than I do 🙃

dendoesit (approved):
good stuff 🔥

PR #1063, opened 2022-06-02
fix(GRW-1188): displays water leakage device only for NO

guilhermespopolin:
## What?

Displays water leakage device only for NO

## Why?

That's not relevante in other market other then NO

**Ticket(s): [GRW-1188](https://hedvig.atlassian.net/browse/GRW-1188)**


src/client/utils/hooks/useQuoteCartData.ts
```diff
@@ -120,18 +124,19 @@ const getHouseDetails = (data: GenericQuoteData) => {
...
+      label: 'CHECKOUT_DETAILS_YEAR_OF_OWNERSHIP',
+      value: { value: data.yearOfOwnership },
+    })
+  }
+
+  if (data.type === InsuranceType.NORWEGIAN_HOUSE) {
+    houseDetails.push({
+      label: 'CHECKOUT_DETAILS_WATER_LEAKAGE',
+      value: {
+        textKey: data.waterLeakageDetector ? 'YES' : 'NO',
```
mehdielket, line 137: Anyway we can check if the value  is null instead. I assume, in NO, the value will always be set ("true" or "false), but it should be null or undefined if it does not exist. That should allow us to avoid the if statement.

mehdielket (approved):
If you think the fix can be doable, that would be great, otherwise we can merge and fix later

src/client/utils/hooks/useQuoteCartData.ts
guilhermespopolin: Thought the same thing but the issue is that is defined like this `waterLeakageDetector?: boolean`. When it's `null` it means 'Not Sure'.

robinandeer: This should be modelled as an enum :) => yes, no, not_sure

guilhermespopolin: That's right

PR #1065, opened 2022-06-03
Fix / GRW-1185 / Prevent stale data checkout

marian-buc:
## What?
Added isLoadingQuoteCart conditional button rendering to prevent signing users with wrong data

## Why?
After editing the details we need to wait for the query to finish before letting users sign.

StartDate fixes coming in a new PR as well

**Ticket(s): [GRW-1185, GRW-1191]**


src/client/pages/OfferNew/Introduction/Sidebar/StickyBottomSidebar.tsx
```diff
@@ -76,7 +78,11 @@ export const StickyBottomSidebar: React.FC<Hidable & {
   return (
     <Wrapper isVisible={reallyIsVisible} displayNone={displayNone}>
       <CtaWrapper>
-        {isConnectPaymentAtSignEnabled ? (
+        {isLoadingQuoteCart ? (
+          <Button size="sm" fullWidth disabled style={{ height: '2.75rem' }}>
+            <LoadingDots color={colorsV3.gray500} />
+          </Button>
+        ) : isConnectPaymentAtSignEnabled ? (
```
robinandeer, line 85: I think we should skip this, it's just a link to open the checkout, right?

Possibly mark the button disabled but no need to show multiple loading state, confused the user

src/client/pages/Offer/Checkout/index.tsx
```diff
@@ -269,7 +270,13 @@ export const Checkout = ({
     createQuoteBundle,
     { loading: isBundleCreationInProgress },
   ] = useCreateQuoteBundleMutation()
-
+  const { loading: isLoadingQuoteCart } = useQuoteCartQuery({
+    variables: {
+      id: quoteCartId,
+      locale: locale.isoLocale,
+    },
+    notifyOnNetworkStatusChange: true,
```
robinandeer, line 278: Should we make this more explicit by adding something like this also:

```
if (networkStatus === NetworkStatus.refetch) return 'Refetching!'
```

https://www.apollographql.com/docs/react/data/queries/

src/client/pages/Offer/Checkout/index.tsx
```diff
@@ -551,6 +558,7 @@ export const Checkout = ({
             canInitiateSign={
               formik.isValid &&
               !isBundleCreationInProgress &&
+              !isLoadingQuoteCart &&
```
robinandeer, line 561: Nice :)

src/client/pages/Offer/Checkout/index.tsx
guilhermespopolin: I guess I didn't get it 😬

marian-buc: not really, since im not using the networkStatus field, but using notifyOnNetworkStatusChange triggers the correct loading value
@guilhermespopolin without this opt, the query just refreshes the data whenever there's a change, but keeps loading:false

src/client/pages/Offer/Introduction/Sidebar/index.tsx
```diff
@@ -206,7 +215,17 @@ export const Sidebar: React.FC<SidebarProps> = ({
...
+                  <Button
+                    size="sm"
+                    fullWidth
+                    style={{ height: '2.75rem' }}
+                    foreground={colorsV3.gray900}
+                    background={colorsV3.purple500}
+                  >
+                    <LoadingDots color={colorsV3.gray500} />
+                  </Button>
+                ) : isConnectPaymentAtSignEnabled ? (
```
robinandeer, line 228: Wanna update this also so we just show the disabled state instead of multiple loading states?

robinandeer (commented):
..otherwise looks good I think!

robinandeer (approved):
Code looks good, haven't tested tho :)

PR #1067, opened 2022-06-04
Added phone number for suport in NO. 

dendoesit:
## What?

Added phone number to NO flow for extra support . 

## Why?

Request from stakeholders.


src/client/l10n/callCenters.ts
```diff
@@ -32,4 +32,22 @@ export const callCenters: CallCenters = {
...
+  NO: {
+    callCenter: {
+      displayNumber: '38 99 41 11',
+      hrefNumber: 'tel:38 99 41 11',
+      opensAt: '09',
+      closesAt: '16',
+      lunchStartsAt: '12',
+      lunchEndsAt: '13',
+    },
+    supportNumber: {
```
simonauner, line 44: The idea here was to keep these as two separate entities, with one being the call center where one can get support of buying an insurance, and the other one is to IEX. But here it looks like the same number, is that intended?

The reason is that the `supportNumber` is being shown on the Payment and Download app screens, which the call center cannot help out with. Or that was the intention from my and @amandaflod at least.

guilhermespopolin (approved):

src/client/l10n/callCenters.ts
dendoesit: Yes, I have confirmed with Oscar that both places should have the same phone number.

simonauner: Ok, great 👍

simonauner (approved):

PR #1068, opened 2022-06-06
Fix/grw 1191/prevent stale data signing

marian-buc:
## What?
Created useStartDate hook and refactored the component a bit in order to get access to the mutation's loading state outside of the date component. Also added isLoading for the campaign queries.

## Why?
To prevent the users starting the checkout before the mutations run on NO and SE markets and to fix the issue where StartDate would sometimes show the old date for a moment.

Before:

https://user-images.githubusercontent.com/101566411/172153640-bab3da1e-c735-463b-aaaa-483d02785dc9.mp4

Now:

https://user-images.githubusercontent.com/101566411/172131466-0618442c-362a-4cba-9521-b21ecbf09e5c.mp4

Though this can't really happen on CPAS, only if the user adds the ssn and goes back and forth very quickly, I've removed the possibility for that as well

**Ticket(s): [GRW-1191]**

[GRW-1191]: https://hedvig.atlassian.net/browse/GRW-1191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/pages/Checkout/CheckoutDetails/CheckoutDetails.tsx
```diff
@@ -106,6 +118,7 @@ export const CheckoutDetails = () => {
         buttonText={textKeys.CHECKOUT_FOOTER_CONTINUE_TO_PAYMENT()}
         buttonLinkTo={paymentPageLink}
         onClick={handleOnClick}
+        isLoading={createQuoteBundleMutation[1].loading}
```
simonauner, line 121: Why is this only for `[1]`? Would it make sense to have `createQuoteBundleMutation.some(m => m.loading)` to get all queries that are loading?

simonauner: Oh, now I understand - it's because you didn't destructure `createQuoteBundleMutation`. Couldn't that be done just for clarity's sake?
`const [, isLoading] = createQuoteBundleMutation`?

simonauner (approved):
Good digging in finding and solving these problems!

I would perhaps have solved it the other way around, by having an `onChange` in the Start Date component and then sending the mutation in the Checkout Details component, to keep the Start Date component "dumb". But I guess it's a matter of taste!

Other than that just a request to destructure a `loading` prop that was a bit hard to understand upon first look.

src/client/pages/Checkout/shared/Footer.tsx
```diff
@@ -100,7 +100,12 @@ export const Footer = ({
         {buttonText && (
           <ButtonWrapper>
             {buttonLinkTo && (
-              <LinkButton {...buttonProps} to={buttonLinkTo} onClick={onClick}>
+              <LinkButton
```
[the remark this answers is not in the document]
marian-buc: it was weird, but the props were there so I think it was just un-intentional omission 😅

src/client/pages/Checkout/CheckoutDetails/CheckoutDetails.tsx
marian-buc: yeap, much better this way

simonauner:
> 

Sure, sounds reasonable! 👍

marian-buc:
I've added the useStartDate hook and cleaned things out nicely

simonauner:
> I've added the useStartDate hook and cleaned things out nicely

Interesting approach with exposing a hook that sets the props for the component that is being created - creative 🎉  Nice job 🙌

marian-buc:
@simonauner this seems to fix the error, please lmk if you're still able to crash it

src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
```diff
@@ -105,15 +105,13 @@ type CancellationOptionsProps = {
...
+  isDisabled,
   quotes,
   onToggleCancellationOption,
 }: CancellationOptionsProps) => {
-  const isDisabled = loadingQuoteIds.length > 0
-
   return (
     <>
-      {quotes.map((quote) => {
+      {quotes?.map((quote) => {
```
robinandeer, line 114: Why does this change? Seems like the prop types don't allow passing `quotes` as undefined, no?

src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
```diff
@@ -456,11 +415,82 @@ export const StartDate = ({
...
+        isDisabled={loadingQuoteIds.length > 0 || isLoading}
         onToggleCancellationOption={(isChecked, quoteId) =>
           handleSelectNewStartDate(isChecked ? null : new Date(), [quoteId])
         }
       />
     </>
   )
 }
+
+export const useStartDate = (): Omit<StartDateProps, 'size' | 'modal'> => {
```
robinandeer, line 427: I would consider renaming the hook some something more explicit. Now it sounds like it would return a start date, while it's actually returning props.

Maybe something like `useStartDateProps`?

src/client/pages/Offer/Introduction/Sidebar/index.tsx
```diff
@@ -248,7 +259,10 @@ export const Sidebar: React.FC<SidebarProps> = ({
                     to={`/${localePath}/new-member/checkout/details/${quoteCartId}`}
                     foreground={colorsV3.gray900}
                     background={colorsV3.purple500}
-                    disabled={isLoadingQuoteCart}
+                    disabled={isLoading}
+                    onClick={(e) => {
+                      if (isLoading) e.preventDefault()
+                    }}
```
robinandeer, line 265: Curious where this comes from. What does it mean that we want to prevent default behavior if loading?

robinandeer (commented):
Nice work @marian-buc! Just left some comments to clarify a few bits

I think we can do a lot in Racoon to make these kinds of things a lot simpler to deal with :)

src/client/pages/Offer/Introduction/Sidebar/index.tsx
marian-buc: Since that's actually a link button, and can be clicked even if disabled={true}, it needed this to prevent from switching to the next page. 😅

guilhermespopolin: 🤔 I thought that by default a link wouldn’t be able to be clicked when it's disabled as it happens with `<button>`s but it seems not to be the case. After some quick search, it seems people use  CSS `pointer-events: none` to implement disabled links, so there's this option as well in case you're up to an alternative solution.

src/client/pages/Offer/Introduction/Sidebar/CancellationOptions.tsx
marian-buc: that's something I forgot to undo while developing, sorry

PR #1072, opened 2022-06-07
feat(onboarding-dk): enables onboarding flow unconditionally

guilhermespopolin:
## What?

enables onboarding-DK flow unconditionally

## Why?

A main component of House DK is a single onboarding flow so as part of the release of it we should make that flow available unconditionally

**Ticket(s): [GRW-876](https://hedvig.atlassian.net/browse/GRW-876)**


simonauner (approved):
Can `isHouseEnabled` flag be removed

PR #1074, opened 2022-06-08
refactor: Remove unnecessary Helmet elements

simonauner:
## What?

Remove unnecessary Helmet elements

## Why?

Since we set `title` and `og` attributes in `routes.tsx` there is no need to also define them in the React page components. I found that while developing some Car stuff, we were defining the same page title inside `<Helmet>` elements in pages as we do in `routes.tsx`. It's bound to cause confusion if you think that you can update the page title in the page component but then you miss updating it in `routes.tsx`.

We still use `<Helmet>` in two places where I didn't find it worth refactoring it away: `Landing.tsx` and `FourOhFour.tsx`.

**Ticket(s): []**


gustaveen (approved):
Great cleanup! 🙏

PR #1077, opened 2022-06-12
Fix/GRW-1198/CPAS race condition fix

marian-buc:
## What?
Added the isDataLoading bool to useSubmitFormOnSsnChange

## Why?
To prevent the user from recreating the quote bundle by editing the SSN while the checkout is running and crashing the mutation

**Ticket(s): [GRW-1198]**

[GRW-1198]: https://hedvig.atlassian.net/browse/GRW-1198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
+++ b/src/client/pages/Checkout/CheckoutPayment/CheckoutPayment.tsx
@@ -166,17 +166,26 @@ const Terms = styled.div`
   }
 `
 
-const useSubmitFormOnSsnChange = (formik: FormikProps<QuoteInput>) => {
+const useSubmitFormOnSsnChange = (
+  formik: FormikProps<QuoteInput>,
+  isDataLoading: boolean,
+) => {
   const sendDatadogAction = useSendDatadogAction()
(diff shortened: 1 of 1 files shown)


marian-buc:
@simonauner this seems to fix the err, lmk if you're still able to crash it 😅

simonauner:
> @simonauner this seems to fix the err, lmk if you're still able to crash it 😅

Not sure if it's related to this fix or not, but I cannot add my credit card details when running this branch locally. I tested with the same quote cart that was used to for the testing session `5b2348c1-12d0-4bc6-af20-3b8ab5f5c709`, and I just get a never-ending spinner.

image

robinandeer (approved):
Tested in DK and works for me :)

I got the adyen dropin to load just as expected

marian-buc:
@simonauner I think that's because that particular quoteCart has already erred? I will merge this since it seems to work for new quoteCarts

guilhermespopolin (approved):

PR #1081, opened 2022-06-13
fix / Use insurable limits from the server as they are

simonauner:
## What?

Use insurable limits from the server as they are, without mapping on type

## Why?

We can sometimes get several insurable limits with the same `type` from BE since there's no restriction on unique-ness there: 

image

This caused some insurable limit items not to show in production because they were accidentally the same type. I think we should allow multiple items of the same type if the BE allows it.

The problem was that we mapped this array into a Map keyed on `type` which meant that we only showed one of the array items on the page:

image

After this PR we show all insurable limits as they are sent from BE:

image

I guess this was used to not get duplicates, but it seems it's not needed for bundles either, still works:

https://user-images.githubusercontent.com/1343979/173360122-b66fe86e-9606-4882-b8de-c34cf637e53a.mov

**Ticket(s): []**

### DIFF
--- a/src/client/pages/OfferNew/Perils/AmountCollection/index.tsx
+++ b/src/client/pages/OfferNew/Perils/AmountCollection/index.tsx
@@ -20,19 +20,17 @@ type Props = {
   offer: OfferQuote
 }
 
-export const AmountCollection: React.FC<Props> = ({ offer }) => {
+export const AmountCollection = ({ offer }: Props) => {
   const textKeys = useTextKeys()
-
-  const limits = [...offer.insurableLimits.entries()]
 
   return (
     <Container>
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         uses: actions/setup-node@v3
         with:
           node-version-file: '.nvmrc'
-          cache: 'yarn'
+          # cache: 'yarn' # TODO: take back once Github solves it's cache problem
```
gustaveen, line 29: 😌

gustaveen (approved):
LGTM 👌

guilhermespopolin (approved):
🚀

simonauner:
Dropped the commit which disabled GH Action's cache 🗑️

PR #1082, opened 2022-06-14
Use flexbox to align items to bottom when height differs

simonauner:
## What?

Use flexbox to bottom-align content when the height (amount of content) differs between cards on the same row.

## Why?

Before:

Screenshot 2022-06-14 at 07 37 11

After (notice first card's content is aligned to bottom):

Screenshot 2022-06-14 at 07 53 45

**Ticket(s): []**


robinandeer (approved):
Nice tweak!

PR #1087, opened 2022-06-15
Fix / Make sure we don't block analytics events

robinandeer:
## What?

Allow domains like:

- region1.google-analytics.com
- region2.google-analytics.com
- www.google-analytics.com

## Why?

**Ticket(s): []**


guilhermespopolin (approved):

dendoesit (approved):
👍

PR #1089, opened 2022-06-17
GRW-1212 / feat / Add EmbarkProvider::externalRedirects::Web 

simonauner:
## What?

Update to latest Embark and add `EmbarkProvider::externalRedirects::Web(path: string)` so that we can redirect to any web route

## Why?

We need to redirect users to a MW page when we cannot create a quote, see Jira ticket below.

**Ticket(s): [GRW-1212](https://hedvig.atlassian.net/browse/GRW-1212)**


guilhermespopolin (approved):
🚀

PR #1092, opened 2022-06-27
Remove Sentry from project

robinandeer:
## What?

Replace error logging with Datadog RUM.
Skip explicit error logging on server - rely on console logs.

## Why?

Use Datadog instead.

**Ticket(s): []**


src/client/pages/Offer/Checkout/InsuranceSummary.tsx
```diff
@@ -40,6 +40,38 @@ export const Row = styled.div`
...
+  hasError: boolean
+}
+
+class ErrorBoundary extends Component<ErrorBoundaryProps, ErrorBoundaryState> {
+  constructor(props: any) {
+    super(props)
+    this.state = { hasError: false }
+  }
+
+  static getDerivedStateFromError() {
```
gustaveen, line 58: Nice! Didn't know about this lifecycle method. For reference: https://reactjs.org/docs/react-component.html#static-getderivedstatefromerror

gustaveen (approved):
🗑 🧘

guilhermespopolin (approved):

src/client/pages/Offer/Checkout/InsuranceSummary.tsx
robinandeer: hehe yeah it's pretty obscure, in most cases I would use `react-error-boundary` but didn't want to add an external dependency

PR #1109, opened 2022-08-02
Grw 1311/fix/update storybook fix build warnings

alebedev:
## What?

Fix many pages of storybook build warnings by updating babel and storybook

## Why?

Removing warnings improve developer experience, we don't want warnings that we ignore

**Ticket(s): [GRW-1311]**

[GRW-1311]: https://hedvig.atlassian.net/browse/GRW-1311?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,7 @@
     "@types/koa-router": "^7.4.1",
     "@types/md5": "^2.1.33",
     "@types/node": "^14.11.8",
-    "@types/react": "^16.9.23",
+    "@types/react": "^16.14.29",
     "@types/react-dom": "^16.9.5",
     "@types/react-facebook-login": "^4.1.0",
     "@types/react-helmet-async": "^0.0.1",
@@ -192,6 +192,7 @@
   },
   "sideEffects": false,
   "resolutions": {
-    "koa-router/path-to-regexp": "6.1.0"
+    "koa-router/path-to-regexp": "6.1.0",
+    "@types/react": "16.14.29"
   }
 }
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -193,6 +192,7 @@
   },
   "sideEffects": false,
   "resolutions": {
-    "koa-router/path-to-regexp": "6.1.0"
+    "koa-router/path-to-regexp": "6.1.0",
+    "@types/react": "16.14.29"
```
robinandeer, line 196: I'm not 100% how this works but do we need this or can we pin the version in the dependencies above?

robinandeer (approved):
Great stuff!

[cathack](https://app.graphite.dev/macros/HedvigInsurance)

guilhermespopolin (approved):
Nice 👷🏻

package.json
alebedev: This forces any direct or transitive dependency to use specific version of package (see https://classic.yarnpkg.com/lang/en/docs/selective-version-resolutions/)

Using resolutions is usually last resort for conflicting transitive dependencies.  This is exactly our case - storybook and friends all want @types/react ^18.something and we don't want to go back to very old storybook that wants react 16

robinandeer: great, thanks :)

robinandeer (approved):

PR #1113, opened 2022-08-11
feat(gGRW-1336): added product selector

guilhermespopolin:
## What?

Add a component to support cross-selling in NO/DK. I'm aware that this implementation counts with some repetition between `MainProductCard` and `AdditionalProductCard`. I've tried to extract the shared code into a `BaseCard` but since they work in a opposing way when it comes about responsivenesses of the UI, it turned out into a _media query_ war that, in my opinion, makes things more difficult to reason about. In any case, I'm open to suggestions.

## Why?

It's required to provide _Petangon House Cross-Sell_  experience

## How to test

Checkout to this branch and run _Storybook_.

## Screenshots / recordings 

image

**Ticket(s): [GRW-1336](https://hedvig.atlassian.net/browse/GRW-1336)**


src/client/pages/Offer/ProductSelector/ProductSlector.tsx
```diff
@@ -0,0 +1,159 @@
...
+                }
+                title={name}
+                price={price}
+                description={description}
+                image={image}
+                checked={selectedProductsByCategory.mainProducts[id]}
+                onClick={() => {
+                  if (isTheOnlySelectedMainProduct) {
+                    inputRef.current?.setCustomValidity(
+                      textKeys.OFFER_PAGE_MISSING_MAIN_COVERAGE_ERROR(),
```
guilhermespopolin, line 88: For some weird reason textkeys is not working here. If I hard code the error message I can see it normally when trying to unselect a main product when I have only that one checked. Any ideas?

src/client/pages/Offer/ProductSelector/ProductSlector.tsx
```diff
@@ -0,0 +1,159 @@
...
+            const isTheOnlySelectedMainProduct =
+              selectedMainProducts.length === 1 &&
+              selectedMainProducts[0] === id
+
+            return (
+              <MainProductCard
+                key={id}
+                checkboxRef={
+                  isTheOnlySelectedMainProduct ? inputRef : undefined
+                }
```
robinandeer, line 79: What are we trying to do here?

src/client/pages/Offer/ProductSelector/ProductSlector.tsx
```diff
@@ -0,0 +1,159 @@
...
+                title={name}
+                price={price}
+                description={description}
+                image={image}
+                checked={selectedProductsByCategory.mainProducts[id]}
+                onClick={() => {
```
robinandeer, line 85: nit: extract this into a names function

robinandeer (commented):
Cool cool! Is this only supposed to work in NO/DK or are we using the same solution for SE?

guilhermespopolin:
> Cool cool! Is this only supposed to work in NO/DK or are we using the same solution for SE?

The way I understood, only for NO/DK. At least during this first lunch.

src/client/pages/Offer/ProductSelector/ProductSlector.tsx
guilhermespopolin: In order to show an error message using the native Web api  (`element.[setCustomValidity()/reportValidity()]`) I need to hold a reference to checkbox that belongs to only currently selected main product. That's because in case the user clicks on that card with the intention of unselecting it I can use that reference and call the API to show the error message.

src/client/pages/Offer/ProductSelector/ProductSlector.tsx
```diff
@@ -0,0 +1,159 @@
...
+                    }))
+                  }
+                }}
+              />
+            )
+          })}
+        </MainCoverageCardGrid>
+      </section>
+      <section>
+        <h2>Additional Coverage</h2>
```
robinandeer, line 107: Are we adding translated keys already?

robinandeer (approved):
Cool, I guess we can investigate if this could also work for SE in the meantime or if that would require a lot of changes.

I guess this is a good approach. The only thing I would consider is to delay the validation. I imagine we can do that when someone presses "checkout" - if they haven't selected a main coverage we can show them an error message. 

But I realize that not usually how we do things in this code base, I just think it would simplify things hehe

guilhermespopolin:
> Cool, I guess we can investigate if this could also work for SE in the meantime or if that would require a lot of changes.
> 
> I guess this is a good approach. The only thing I would consider is to delay the validation. I imagine we can do that when someone presses "checkout" - if they haven't selected a main coverage we can show them an error message.
> 
> But I realize that not usually how we do things in this code base, I just think it would simplify things hehe
I guess it makes sense also investigate the possibility of adding it for SE as well. In fact, that could a way to test if the format we have for possible variations for SE would also work for NO/DK.

🤔 I see. About the validation one thing that I've thought is by only validating when _Checkout_ is pressed should we allow everything be unselected? If that's true, what should we display in the summary sidebar? Or, should we hide it?

src/client/pages/Offer/ProductSelector/ProductSlector.tsx
guilhermespopolin: Good point. I guess we can already add some text keys for those headings. Will do.

src/client/pages/Offer/ProductSelector/ProductSlector.tsx
robinandeer: Right, I see the reason now. Seems a little non-standard to assign the ref conditionally, though.

Another approach would be to assign refs to each checkbox and later call "setCustomValidity" on the relevant one. But okey since this is the web-onboarding repo hehe

robinandeer:
> > Cool, I guess we can investigate if this could also work for SE in the meantime or if that would require a lot of changes.
> > I guess this is a good approach. The only thing I would consider is to delay the validation. I imagine we can do that when someone presses "checkout" - if they haven't selected a main coverage we can show them an error message.
> > But I realize that not usually how we do things in this code base, I just think it would simplify things hehe
> > I guess it makes sense also investigate the possibility of adding it for SE as well. In fact, that could a way to test if the format we have for possible variations for SE would also work for NO/DK.
> 
> About the validation one thing that I've thought is by only validating when _Checkout_ is pressed should we allow everything be unselected? If that's true, what should we display in the summary sidebar? Or, should we hide it?

No, you have to checkout at least one product/insurance. I guess if you would try to checkout no products we should either disable the button or show an error when you click "checkout". But only if that simplifies the code -- otherwise we can continue with the setup you have started already!

src/client/pages/Offer/ProductSelector/ProductSlector.tsx
guilhermespopolin: I've tried a couple of different approaches but they appear more confusing 😢 
So I think I'm gonna leave it like that and iterate over it later.

PR #1115, opened 2022-08-16
styles(Switch): removed undesired line-height from icon

guilhermespopolin:
## What?

Removes undesired line-height from icon since

**Before**
old

**After**
new

## Why?

It adds some vertical spacing take makes difficult align things vertically.
It was added as a side effect of #1113


gustaveen (approved):
🙌

amandaflod (approved):
💅⭐️

PR #1116, opened 2022-08-17
GRW-1352: Embark: try to parse campaign code from URL

simonauner:
This gives us the possibility to pass a campaign code to Embark flows for cross-sells with
reduced prices

## What?

Parse `campaignCode` query string and set it as campaign for quote cart.

## Why?

As part of [GRW-1352](https://hedvig.atlassian.net/browse/GRW-1352) we might want to cross-sell house/car if the user bought the other, maybe with a campaign code attached to get a discount.  So when the user is at the end of the flow we can present a link that takes them to the other Embark flow with a campaign code attached.

This screenshot shows the offer page after going through the Embark flow for SE Home with `campaignCode=Meletis4ever`:

Screenshot 2022-08-17 at 14 38 51

**Ticket(s): [GRW-1352]**

[GRW-1352]: https://hedvig.atlassian.net/browse/GRW-1352?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/pages/Embark/index.tsx
+++ b/src/client/pages/Embark/index.tsx
@@ -6,8 +6,8 @@ import {
   useEmbark,
 } from '@hedviginsurance/embark'
 import { AnimatePresence, motion } from 'framer-motion'
-import React, { useEffect, useState, useCallback } from 'react'
-import { useHistory } from 'react-router'
+import React, { useEffect, useState } from 'react'
+import { useHistory, useLocation } from 'react-router'
 
 import { colorsV3 } from '@hedviginsurance/brand'
 import gql from 'graphql-tag'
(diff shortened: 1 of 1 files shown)


src/client/pages/Embark/index.tsx
```diff
@@ -245,23 +245,28 @@ interface AngelVariables {
 const useInitQuoteCart = ({ skip = false }) => {
   const quoteCartId = useOnboardingQuoteCartId()
   const [addCampaignCode] = useAddCampaignCodeMutation()
+  const { search } = useLocation()
+  const searchParams = useMemo(() => new URLSearchParams(search), [search])
```
alebedev, line 249: If you move search string handling inside initQuoteCart, you won't need useMemo for searchParams.  Less code and slightly more efficient too

src/client/pages/Embark/index.tsx
```diff
@@ -245,23 +245,28 @@ interface AngelVariables {
...
           },
         })
       } catch {
         CampaignCode.remove()
       }
     }
-  }, [addCampaignCode, quoteCartId])
+  }, [addCampaignCode, quoteCartId, searchParams])
 
   useEffect(() => {
```
alebedev, line 271: Nit: you can apply same idea here.  Since effect runs once, you can move initQuoteCart declaration inside it and remove one more hook.  Unrelated to your change, I know.

alebedev (approved):
Looks good

src/client/pages/Embark/index.tsx
simonauner: Great, thanks 👍

src/client/pages/Embark/index.tsx
simonauner: Nice, will fix this as well  👌

simonauner:
This can be closed since we parse `code` and set it as a cookie:
https://github.com/HedvigInsurance/web-onboarding/blob/a00e02e2321d21a27d9c9b575471476248bda6e2/src/server/page.ts#L166
and that cookie is already read by the old code and set by `addCampaignCode`

PR #1118, opened 2022-08-19
Add CrossSell promo Card on DownloadPage

simonauner:
## What?

This PR:
- Refactors `AdditionalProductCard` so that the styles can be re-used for a `LinkCard` that I want to use on the Download app page. This way we get consistency in font sizes, image-to-text ratios and so on.
- Adds a Cross-sell promotion card on the Download app page given that GraphQL query says that it's ok to do so
- Widens the text box on the Download app page since the card gets too small otherwise

## Why?

**Ticket(s): [GRW-1379]**

Desktop:

image

iPad portrait:
image

Mobile:
image

And this is how it looks with the actual data from Robin's query:

image

[GRW-1379]: https://hedvig.atlassian.net/browse/GRW-1379?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/pages/Download/components/DownloadPageContent.tsx
```diff
@@ -6,6 +6,16 @@ import { LARGE_SCREEN_MEDIA_QUERY } from 'utils/mediaQueries'
...
+import { Arrow } from 'components/icons/Arrow'
+import {
+  Description,
+  Header,
+  Image,
+  ImageFrame,
+  LinkCard,
+  Section,
+  Title,
+} from '../../Offer/ProductSelector/AdditionalProductCard'
```
robinandeer, line 18: hmm I wonder if it's worth it to extract this to a global component so it's clear that it's being used in different pages

src/client/pages/Download/components/DownloadPageContent.tsx
```diff
@@ -130,32 +159,49 @@ export const DownloadPageContent = ({
...
+          <ImageFrame>
+            <Image src={crossSell.image} />
+          </ImageFrame>
+          <Section>
+            <Header>
+              <Title>{crossSell.title}</Title>
+            </Header>
+            <Description>{crossSell.description}</Description>
+            <Cta>
+              {crossSell.callToAction} <Arrow direction="forward" />
```
robinandeer, line 200: kinda looks like the arrow is slightly too big to align well with the text - maybe make it a big smaller?

robinandeer (approved):
Good stuff!

src/client/pages/Download/components/DownloadPageContent.tsx
simonauner: Yes, perhaps! But then I think all cards should be global since they're used for the same feature, yes?

src/client/pages/Download/components/DownloadPageContent.tsx
simonauner: Sure, I can tweak the size 👍

src/client/pages/Download/components/DownloadPageContent.tsx
robinandeer: Yeah sounds about right

simonauner: Fixed now 👍

PR #1126, opened 2022-08-25
GRW-1386 / feat / Add links to cross sell cards, hardcoded with campaign code

simonauner:
## What?

Add hard-coded URLs to which page we want users to go to when they click the cross sell cards.

## Why?

We need to know where to forward users. Unfortunately we have no BE support for managing where the links point to, so they're just hard-coded in FE code 😞 

**Ticket(s): [GRW-1386](https://hedvig.atlassian.net/browse/GRW-1386)**

Screenshot 2022-08-25 at 11 48 00

Screenshot 2022-08-25 at 11 47 31

### DIFF
--- a/src/client/pages/Download/components/DownloadPageContent.tsx
+++ b/src/client/pages/Download/components/DownloadPageContent.tsx
@@ -173,11 +173,11 @@ export const DownloadPageContent = ({
 }
 
 const CrossSells = () => {
-  const { loading, crossSells } = useCrossSells()
+  const { crossSells } = useCrossSells()
   const textKeys = useTextKeys()
   const currentLocale = useCurrentLocale()
 
-  if (loading || !crossSells?.length) return null
+  if (!crossSells || crossSells.length === 0) return null
 
   return (
     <>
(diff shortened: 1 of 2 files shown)


src/client/pages/Download/components/DownloadPageContent.tsx
```diff
@@ -174,8 +175,9 @@ export const DownloadPageContent = ({
 const CrossSells = () => {
   const { loading, crossSells } = useCrossSells()
   const textKeys = useTextKeys()
+  const currentLocale = useCurrentLocale()
 
-  if (loading) return null
+  if (loading || !crossSells?.length) return null
```
robinandeer, line 180: nit: do we need to check `loading`?

I would also make the check more explicit =>

```
if (!crossSells || crossSells.length === 0) return null
```

I just find it easier to read. We are actually covering two cases: 1. are we loading data 2. do we have 0 possible cross-sells?

robinandeer (approved):
Let's hope we can live with this shortcut to Coverage Store

Added via Giphy

src/client/pages/Download/components/DownloadPageContent.tsx
simonauner: Yeah, I can change it to your suggestion. The `loading` check was added by you 🙃 
If it would only be `!crossSells?.length` we might briefly show the header "Others also bought" and then remove it if `length === 0`, but your suggestion would solve that 👍

simonauner:
> Let's hope we can live with this shortcut to Coverage Store
> 

Yeah, it hurts

gif

PR #1131, opened 2022-09-01
GRW-1390/ feat / Add switching notice on offer page

simonauner:
## What?

* Add `Shuffle` icon
* Add `Notice` component, similar to [MUI's Alert](https://mui.com/material-ui/react-alert/)
* Add Switching Notice in Offer page, but never show it for now since the `currentInsurer` data is not telling the truth right now

## Why?

We should inform users that we can help them cancel their previous insurance.

**Ticket(s): [GRW-1390](https://hedvig.atlassian.net/browse/GRW-1390)**

The `Notice` component in different variations:

image

As Switching notice on the Offer page:

Screenshot 2022-09-01 at 10 49 34

Data for `currentInsurer` in Test, so we cannot use that I guess.

Screenshot 2022-09-01 at 09 54 03


robinandeer (approved):
Good stuff :D no comments

simonauner:
> Good stuff :D no comments

Achievement unlocked! 🏆 

😀

PR #1132, opened 2022-09-01
GRW-1445 - refact(Selector): derivate state from URL

guilhermespopolin:
## What?

Reactor `<Selector />` component by:
* make its state completely derivated from URL, so no need to keep internal (React) state
* simplify UI/UX used to block users of unselecting the currently only selected standalone product

## Why?

Simplify the logic used to get it done.

Screenshot 2022-09-01 at 13 07 58

**Ticket(s): [GRW-1445](https://hedvig.atlassian.net/browse/GRW-1445)**

PR #1139, opened 2022-09-02
Revert "refact(Selector): derivate state from URL"

guilhermespopolin:
## What?

This reverts #1132 

## Why?

It wasn't approved by design

### DIFF
--- a/src/client/pages/Offer/ProductSelector/Selector.tsx
+++ b/src/client/pages/Offer/ProductSelector/Selector.tsx
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useState, useRef, useMemo } from 'react'
 import { useLocation, useHistory } from 'react-router'
 import styled from '@emotion/styled'
 import { useTextKeys } from 'utils/textKeys'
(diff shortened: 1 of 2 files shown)


src/client/pages/Offer/ProductSelector/Selector.tsx
```diff
@@ -22,6 +22,57 @@ export type SelectorProps = {
...
+}: {
+  standaloneProducts: Array<Product>
+  additionalProducts: Array<Product>
+  searchParams: string
+}): State => {
+  const selectedInsuranceTypes = new Set(
+    new URLSearchParams(searchParams).getAll(SEARCH_PARAM),
+  )
+
+  const getProductIdByValueMap = (quotes: Array<Product>) =>
```
alebedev, line 43: Nit: Constructing with `Object.fromEntries()` might be easier to read than manual reduce.  And it's more efficient since you don't copy on every iteration

src/client/pages/Offer/ProductSelector/Selector.tsx
```diff
@@ -69,18 +103,41 @@ export const Selector = ({
...
+                checkboxRef={isTheOnlyStandaloneProduct ? inputRef : undefined}
                 title={name}
                 price={price}
                 description={description}
                 image={image}
-                checked={selectedProductIds.includes(id)}
-                disabled={isTheOnlyStandaloneProduct}
-                onClick={() => handleClick(id)}
+                checked={selectedProductsByCategory.standaloneProducts[id]}
+                onClick={() => {
```
alebedev, line 117: Looks too big to inline into JSX.  Could you extract it?

src/client/pages/Offer/ProductSelector/Selector.tsx
```diff
@@ -32,34 +83,17 @@ export const Selector = ({
...
 
-  const selectedProductIds = new URLSearchParams(searchParams).getAll(
-    SEARCH_PARAM,
-  )
-  const selectedStandaloneProducts = standaloneProducts.filter(({ id }) =>
-    selectedProductIds.includes(id),
+  const inputRef = useRef<HTMLInputElement | null>(null)
+
+  const [selectedProductsByCategory, setSelectedProductsByCategory] = useState(
+    getStateFromURL({ standaloneProducts, additionalProducts, searchParams }),
```
alebedev, line 89: Can we watch URL changes instead of local state?

alebedev (approved):
Let's try removing state sync by removing the state, should simplify the code

src/client/pages/Offer/ProductSelector/Selector.tsx
robinandeer: hmm yeah I don't think this PR contains all the changes the @guilhermespopolin and I made 🤔

robinandeer (approved):
Hmm so sad we can't simplify it this way... what was the reason design didn't like it @guilhermespopolin?

guilhermespopolin:
> Hmm so sad we can't simplify it this way... what was the reason design didn't like it @guilhermespopolin?

According to Zak, with the new UI it's not so clear what's selected.

guilhermespopolin:
> > Hmm so sad we can't simplify it this way... what was the reason design didn't like it @guilhermespopolin?
> 
> According to Zak, with the new UI it's not so clear what's selected.

PR #1141, opened 2022-09-06
Use new web-specific content on PotentialCrossSells

simonauner:
## What?

* Use web-specific content for cross-sells from PCMS

## Why?

Because we want different content in the cross-sell promos on the web compared to what is shown in the apps.

**Ticket(s): [GRW-1452](https://hedvig.atlassian.net/browse/GRW-1452)**

### DIFF
--- a/src/client/graphql/PotentialCrossSells.graphql
+++ b/src/client/graphql/PotentialCrossSells.graphql
@@ -3,8 +3,8 @@ query PotentialCrossSells($locale: Locale!) {
     potentialCrossSells(locale: $locale) {
       type
       contractType
-      title
-      description
+      webTitle
+      webDescription
       imageUrl
       callToAction
     }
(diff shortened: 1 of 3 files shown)


robinandeer (approved):

src/client/graphql/PotentialCrossSells.graphql
```diff
@@ -3,8 +3,8 @@ query PotentialCrossSells($locale: Locale!) {
     potentialCrossSells(locale: $locale) {
```
robinandeer, line 3: Nothing to consider now but another way to accomplish this is to add a filter to this query like `potentialCrossSells(locale: $locale, filter: WEB)` -- what do you think about that?

simonauner: Yeah, and then in Giraffe swap out the fields from PCMS? I like that solution more 👍 
Especially since we also might have to have different cross-sells in the different platforms..

robinandeer: Yeah exactly, seem slightly more scalable -- but we can start with this if you want!

simonauner: yeah, let's start here. And I'll probably introduce `filter` for when we need web-specific promos altogether.

PR #1145, opened 2022-09-07
Add page for initializing Car cancellation via BankSignering

simonauner:
## What?

* Add page for initializing Car cancellation via BankSignering

This is the page that users will land on when they click the link in the email to initialize cancellation of their previous car insurance:
image

We take the `contractId` query parameter and pass that to BE to get a launch URL back, which we redirect the user to.

## Why?

We need a page that can redirect the user to BankSignering.

**Ticket(s): [GRW-1391]**

Error screen is reusing existing components and translations:

image

[GRW-1391]: https://hedvig.atlassian.net/browse/GRW-1391?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/pages/IntiateCarCancellation/InitiateCarCancellationPage.tsx
```diff
@@ -0,0 +1,57 @@
...
+        {loading && <LoadingDots color={colorsV3.gray500} />}
+      </CenteredWrapper>
+      <ErrorModal isVisible={!!error}>
+        <ErrorHeading>{textKeys.GENERIC_ERROR_HEADING()}</ErrorHeading>
+        <ErrorText>
+          <ReactMarkdown source={textKeys.GENERIC_ERROR_TEXT()} />
+        </ErrorText>
+      </ErrorModal>
+      {data?.initiateCancelOldCarInsurance?.url && (
+        <Redirect to={data?.initiateCancelOldCarInsurance?.url} />
```
guilhermespopolin, line 53: Extremely nit:
I guess you don't need optional chaining here since you've already checked for the presence of that value.

```suggestion
        <Redirect to={data.initiateCancelOldCarInsurance.url} />
```

guilhermespopolin (approved):

src/client/pages/IntiateCarCancellation/InitiateCarCancellationPage.tsx
```diff
@@ -0,0 +1,57 @@
...
+import ReactMarkdown from 'react-markdown'
+import { useInitiateCarCancellationMutation } from 'data/graphql'
+import { LoadingDots } from 'components/LoadingDots/LoadingDots'
+import { Page } from 'components/utils/Page'
+import { ErrorHeading, ErrorModal, ErrorText } from 'components/ErrorModal'
+import { useTextKeys } from 'utils/textKeys'
+
+const CenteredWrapper = styled.div({
+  width: '100%',
+  height: '90vh', // 100 shows scroll bars since we have padding on parent elements
```
alebedev, line 14: If 10vh is less than paddings, UI may break.  Does it work on small mobile screen?

alebedev (approved):

src/client/pages/IntiateCarCancellation/InitiateCarCancellationPage.tsx
simonauner: Ah, true, that's a remnant from old code 👍

src/client/pages/IntiateCarCancellation/InitiateCarCancellationPage.tsx
simonauner: Yes, it works on mobile screen. The only problem that might arise if 10vh is less than paddings is that we get a scroll bar on the page while the request to BE is loading, so I don't see that as a huge issue if it were to happen.

simonauner:
Now that the Giraffe endpoint is actually working it's clear that `<Redirect>` doesn't handle external redirects, so I swapped it to a `window.location.href`.

Plus: added translation for page title.

PR #1147, opened 2022-09-09
GRW-1421 / feat / Close modal with escape

simonauner:
## What?

Just some small updates to modal handling in web-onboarding in regards to working with GRW-1421. 

1. Close modals with Escape button press
2. Move button definitions of Error Modal into the actual component

## Why?

1. It's annoying not to be able to close Modals with Escape and it's a common user interaction pattern 
2. They're duplicated across the code base and I don't want to duplicate them into yet another component (GRW-1421's modal)

**Ticket(s): sort of [GRW-1421]**

### DIFF
--- a/src/client/components/ErrorModal.tsx
+++ b/src/client/components/ErrorModal.tsx
@@ -4,7 +4,7 @@ import { colorsV3 } from '@hedviginsurance/brand'
 import { MEDIUM_SMALL_SCREEN_MEDIA_QUERY } from 'utils/mediaQueries'
 import { WarningTriangleOutlined } from './icons/WarningTriangle'
 import { Modal, ModalProps } from './ModalNew'
-import { Button } from './CloseButton/CloseButton'
+import { Button } from './buttons'
 
 type Props = ModalProps & {
   children: React.ReactNode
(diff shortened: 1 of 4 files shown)


src/client/components/ErrorModal.tsx
```diff
@@ -66,16 +66,40 @@ const ErrorModalWrapper = styled(Modal)`
     max-width: min(36rem, 100% - 2rem);
     min-height: auto;
   }
-  ${Button} {
```
simonauner, line 69: These style the close button, but if that button is used it's rendered in ModalNew and should be styled there

src/client/components/ModalNew.tsx
```diff
@@ -95,6 +95,18 @@ export const Modal = ({
 
   const hasOnCloseFunction = Boolean(onClose)
 
+  useEffect(() => {
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key === 'Escape' && !!onClose) {
```
simonauner, line 100: Using `hasOnCloseFunction` above doesn't satisfy TS 🤷

alebedev: What if it were ```const hasOnCloseFunction = typeof onClose === 'function'```?  

It might be that Boolean does not narrow it enough

alebedev (approved):
Nice, always a pleasure to see more code removed than added

src/client/components/ModalNew.tsx
simonauner: Yes, nice, that worked ⭐

simonauner: erhm.. Worked i VSCode but tsc failed it when trying to push :trollface:

gustaveen (approved):
Wohoo! A small thing that prevents user from raging! (me at least :D)

guilhermespopolin (approved):
Nice one :)

PR #1149, opened 2022-09-09
GRW-1421 / feat / Warn if date is too soon

simonauner:
## What?

1. Change the label of the date selector when Switcher **and** Car to "Select date" instead of "When current insurance expires"
2. Warn the user if the selected date is "too soon" (less than 7 days into the future). 

## Why?

This was originally going to be a Slack post but I guess things will be better documented for future reference here.

**Background:**

We cannot guarantee that a Car insurance with start date less than one week in the future can be switched by BankSignering on time. We want to prompt users if the date is too soon. 

**Current state:**

We have the date picker for when an insurance should start on the Offer page. When creating a car quote, the start date is returned as `null` from BE, which by default is turned into `new Date()` in the Offer page. This is not the case for switchers. Prior to this PR (and after, still), `null` dates for a switcher indicates "when my old insurance expires", and `null` is an allowed value for switchers.

**Problem:**

Since we don't know the expiration date of the old insurance we cannot guarantee that we can switch the user in time, thus we cannot allow the users to select "When my old insurance expires". The user _must_ guess the date so that we can prompt them if the date is too soon. 

We also can't show the warning by default, that would be an annoying experience, as it is designed for this PR. (Maybe we should reconsider the design!)

**Solution:**

1. If Car and Switcher, show a different form label that says "Select date" instead of "When old insurance expires". We need to once again make components aware of that they're rendering something for Car.
2. Disallow `null` for Car switchers

**Other thoughts:**

* We cannot guarantee a switch for car within 7 days - but we can for other insurances 🤔 ? This logic should probably apply to all switcher cases?
* What happens when we implement Insurely for Car? Then we will get the same data as we have for switchers for other insurances? Maybe much of this PR will change then anyway?

⚠️ I have not implemented a validation that says that `null` is not allowed for Car switchers.

## Pics

The label:
image

We cannot allow `null` for Car switchers (but for other types of switchers):
image

Warning notice:
image

**Ticket(s): [GRW-1421]**


src/client/components/ModalNew.tsx
```diff
@@ -95,6 +95,18 @@ export const Modal = ({
 
   const hasOnCloseFunction = Boolean(onClose)
 
+  useEffect(() => {
+    const handleKeyDown = (e: KeyboardEvent) => {
+      if (e.key === 'Escape' && !!onClose) {
```
alebedev, line 100: Nit: can use optional chaining instead
```
if (e.key === 'Escape') {
  onClose?.()
}
```

src/client/components/ModalNew.tsx
```diff
@@ -95,6 +95,18 @@ export const Modal = ({
...
+        onClose()
+      }
+    }
+    window.addEventListener('keydown', handleKeyDown)
+    return () => {
+      window.removeEventListener('keydown', handleKeyDown)
+    }
+  }, [hasOnCloseFunction, onClose])
```
alebedev, line 108: hasOnCloseFunction isn't used in this hook.  Let's investigate why this isn't a lint error, it should be

src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
```diff
@@ -158,13 +173,13 @@ const DateInputModalWrapper = styled.div<{
...
   startDate: string | null,
-  currentInsurer?: CurrentInsurer,
+  isSwitcher?: boolean,
 ) => {
   if (startDate) {
     return parse(startDate, 'yyyy-MM-dd', new Date())
   }
 
-  if (currentInsurer?.switchable) {
+  if (isSwitcher) {
```
alebedev, line 182: Let's comment why

src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
```diff
@@ -207,16 +222,20 @@ const DateForm = ({
...
+    getDefaultDateValue(startDate, isSwitcher),
   )
   const [dateLocale, setDateLocale] = useState<Locale | null>(null)
 
   const [datePickerOpen, setDatePickerOpen] = useState(false)
 
   useEffect(() => {
-    setDateValue(getDefaultDateValue(startDate, currentInsurer))
-  }, [startDate, currentInsurer])
+    setDateValue(getDefaultDateValue(startDate, isSwitcher))
```
alebedev, line 237: Nit: dateValue is calculated from props, we never set it other way.  We can probably simplify code by calculating it and removing the state

alebedev, line 237: Nit: this file is already huh

src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
```diff
@@ -424,6 +462,31 @@ export const StartDate = ({
   )
 }
 
+type TooSoonSwitcherErrorModalProps = {
```
alebedev, line 465: Nit: this file is already huge, maybe we can split most of modal markup and styles into separate module?

alebedev (commented):
Looks good, some potential future improvements in comments

src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
alebedev, line 465: Nit: this file is already huge, maybe we can split error modal and related utils to separate module?

src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
simonauner: Nice, didn't even notice 👍  Will fix!

src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
simonauner: Sure, that's non unreasonable 🙃

src/client/components/ModalNew.tsx
simonauner: Ah, yes, forgot about that one, nice!

src/client/components/ModalNew.tsx
simonauner: Yeah, didn't even think about it!

alebedev (approved):
Looks good!

PR #1152, opened 2022-09-13
GRW-1453 / Feature / Add car switching notice to Download page along with "too soon" warning

alebedev:
## What?

- Add car switching notice to Download page, take data from active contract bundle
- Show "switching too soon" warning

## Why?

Screenshot 2022-09-13 at 12 35 38

**Ticket(s): [GRW-1453]**

[GRW-1453]: https://hedvig.atlassian.net/browse/GRW-1453?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
```diff
@@ -331,8 +319,9 @@ export const StartDate = ({
     quoteIds: Array<string>,
   ) => {
     try {
-      const dateIsTooSoon = isLessThanOneWeekAway(newDateValue)
-      setShowToSoonError(userIsCarSwitcher && dateIsTooSoon)
+      setShowToSoonError(
```
simonauner, line 322: It shouldn't happen, but now if `newDateValue` is `null`, this evaluates to `true` and we show the error message. Before, it didn't. 

But it might be an improvement actually since a `null` date definitely would want us to show the warning 💡

simonauner (approved):
Feels bit like

image

but approved from me 😄

src/client/pages/Offer/Introduction/Sidebar/StartDate.tsx
alebedev: Fixed this and made check more readable.  If we want warnings for nulls this should be handled elsewhere

PR #1153, opened 2022-09-13
GRW-1421 / Set default start date for Car switchers to new Date()

simonauner:
## What?

Set default start date for Car switchers to `new Date()`. Now it's the same as it was before a previous PR (https://github.com/HedvigInsurance/web-onboarding/pull/1149) introduced the logic for Car switcher dates to be default `null`.

This also means that we do not need a custom translation for the date label for Car switchers.

## Why?

BE does not accept `null` as start date for Car switchers.

**Ticket(s): [GRW-1421]**

Screenshot 2022-09-13 at 14 19 52


gustaveen (approved):
🤜🤛

PR #1155, opened 2022-09-14
GRW-1480 - refact(quoteCart_editQuote): safely update apollo cache after mutation

guilhermespopolin:
## What?

* Update `quoteCart_editQuote` mutation usage so apollo client cache gets updated safely. It also improves performance since now there's no need for refetching `QuoteCart` after editing a quote in order to proper update apollo's local cache (and consecutively, the UI)

## Why?

I've noticed a warning after setting start date:

edit-quote-mutation

It makes sense because we're updating the `QuoteCart.bundle` field (`QuoteBundle` type) as a result of executing `quoteCart_editQuote` mutation. Since `QuoteBundle` is not identifiable, `QuoteCart.bundle` field gets overridden by the return of `quoteCart_editQuote` mutation, which can be dangerous. At the end of that day, it was working because we were refetching `QuoteCart` query right after `quoteCart_editQuote` mutation, so even though we're cleaning some stuff from the cache after the mutation, the lost data were being restored with the result of refetching `QuoteCart` query.

In order to solve it all we need to do is to make `quoteCart_editQuote` mutation to return the changed quote --> `BundledQuote` type. Since `BundledQuote` are identifiable, they get normalised by apollo client's cache. That means that apollo can automatically update the cache with the updated quote right after `quoteCart_editQuote` mutation, which means no need for refetching `QuoteCart`.

Observe tho, that this solution is not perfect. As the response for `quoteCart_editQuote` mutation we are returning all of the quotes that belongs to a specific quote cart. As the name suggests, `quoteCart_editQuote` changes **one single** quote. So taking a quote cart with 4 quotes as an example, after editing one of them with `quoteCart_editQuote` mutation  we'll get back a list that contains the updated quote + 3 other quotes that didn't change, which is not ideal. The way I see we could update the schema to improve that from this

```
union EditQuoteResult = QuoteCart | QuoteBundleError

quoteCart_editQuote(id: ID!, quoteId: ID!, payload: JSON!): EditQuoteResult!
```

to this:

```
union EditQuoteResult = BundledQuote | QuoteBundleError

quoteCart_editQuote(id: ID!, quoteId: ID!, payload: JSON!): EditQuoteResult!
```

So instead of returning a list of quotes, we'd be returning the updated quote. However I think we can do that in another iteration since it's not exactly the scope of this ticket and that would be a breaking change for other teams that are using `quoteCart_editQuote` today.

**Ticket(s): [GRW-1480](https://hedvig.atlassian.net/browse/GRW-1480)**


guilhermespopolin:
* master
  * **PR #1155** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/web-onboarding/1155" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/web-onboarding/1155?utm_source=stack-comment).

robinandeer (approved):
[clean](https://app.graphite.dev/macros/HedvigInsurance)

PR #1159, opened 2022-09-15
GRW-1491 / feat / Fix positioning for Sidebar for small screens

simonauner:
## What?

Introduce a `useHasScrolledPastElement` hook and use that to set positioning for Sidebar.

This commit sets the Sidebar to absolutely positioned by default instead of fixed, and then sets fixed to the bottom of the screen when scrolled past. For desktop, to emulate the old fixed positioning we set fixed to top when scrolled past instead.

## Why?

There's a concern that we might have a too long Sidebar now with the Car switching content, and that users might not be able to reach the "Continue to checkout" button if they have a small screen.

**Ticket(s): [GRW-1491]**

### Small screen

https://user-images.githubusercontent.com/1343979/190357647-253fbbe4-74b7-4a26-a9ad-5b4f4e35af92.mov

### Regular desktop screen

https://user-images.githubusercontent.com/1343979/190357804-f1bdef3b-9691-4d1a-847d-c9df4585fad4.mov

[GRW-1491]: https://hedvig.atlassian.net/browse/GRW-1491?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/pages/Offer/Introduction/Sidebar/Sidebar.tsx
+++ b/src/client/pages/Offer/Introduction/Sidebar/Sidebar.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useCallback, useEffect } from 'react'
+import React, { useState, useCallback, useEffect, useRef } from 'react'
 import styled from '@emotion/styled'
 import ReactVisibilitySensor from 'react-visibility-sensor'
 import { colorsV3 } from '@hedviginsurance/brand'
@@ -37,9 +37,11 @@ import { StickyBottomSidebar } from './StickyBottomSidebar'
 import { CampaignCodeModal } from './CampaignCodeModal'
 import { StartDate, useStartDateProps } from './StartDate'
 import { SwitchingNotice } from './SwitchingNotice'
+import { useHasScrolledPastElement } from './useHasScrolledPastElement'
 
 const SIDEBAR_WIDTH = '26rem'
 const SIDEBAR_SPACING_LEFT = '2rem'
+const FIXED_MARGIN = 16
 
 const Wrapper = styled.div`
   width: calc(${SIDEBAR_WIDTH} + ${SIDEBAR_SPACING_LEFT});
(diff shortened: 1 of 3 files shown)


src/client/pages/Offer/Introduction/Sidebar/useHasScrolledPastElement.ts
```diff
@@ -0,0 +1,68 @@
+import React, { useCallback, useEffect, useState } from 'react'
+import { throttle } from 'utils/throttle'
+
+type StickyOptions = {
+  /**
+   * The margin that should be used to to calculate the full element height + paddings + margins
+   **/
+  margin?: number
```
simonauner, line 8: I'm thinking that we could extend this to `marginTop` and `marginBottom` if we see the need in the future

src/client/pages/Offer/Introduction/Sidebar/Sidebar.tsx
```diff
@@ -52,21 +54,40 @@ const Wrapper = styled.div`
...
+    width: SIDEBAR_WIDTH,
+    maxWidth: '100%',
+    padding: '1.25rem 1rem 1rem 1rem',
+    marginTop: '3rem',
+    backgroundColor: colorsV3.white,
+    borderRadius: '8px',
+    boxShadow: '0 8px 16px rgba(0, 0, 0, 0.08)',
+
+    [LARGE_SCREEN_MEDIA_QUERY]: stickyToTop
+      ? {
```
simonauner, line 73: This isn't very compact or beautiful, but does the job. Feel free to suggest an improvement 🙃

simonauner: And I still get to use multi-terneries in web-onboarding
gif

robinandeer (approved):
Tricky situation -- let's remember to avoid this UI pattern in coverage store 😅

Do you have a reference/link to the custom hooks?

simonauner:
> Tricky situation -- let's remember to avoid this UI pattern in coverage store 😅
> 
> Do you have a reference/link to the custom hooks?

Yes, let's put that one in the books 🔖 

Not sure what you're asking for regarding reference. I did not copy-paste it from anywhere else, but used some different sources for ideas although none of them covered exactly what I wanted to achieve since we need to stick to both top and bottom.

PR #1160, opened 2022-09-15
GRW-1480 - refact(Checkout): make usage of quoteCart_editQuote mutation

guilhermespopolin:
## What?

Update `Checkout` component so it makes usage of `quoteCart_editQuote` mutation

## Why?

It consists on a better method for updating quotes rather than recreating them

**Ticket(s): [GRW-1480](https://hedvig.atlassian.net/browse/GRW-1480)**

PR #1158, opened 2022-09-15
GRW-1480 - refact(CheckoutPayment): make usage of quoteCart_editQuote mutation

guilhermespopolin:
## What?

Update `CheckoutPayment` page so it makes usage of `quoteCart_editQuote` mutation

## Why?

It consists on a better method for updating quotes rather than recreating them

**Ticket(s): [GRW-1480](https://hedvig.atlassian.net/browse/GRW-1480)**

PR #1161, opened 2022-09-16
GRW-1480 - refact(DetailsModal): make usage of quoteCart_editQuote mutation

guilhermespopolin:
## What?

* Update `DetailsModal` component so it makes usage of `quoteCart_editQuote` mutation
## Why?

It consists on a better method for updating quotes rather than recreating them

**Ticket(s): [GRW-1480](https://hedvig.atlassian.net/browse/GRW-1480)**


guilhermespopolin:
* master
  * **PR #1155** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/web-onboarding/1155" target="_blank">Graphite</a> 
    * **PR #1158** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/web-onboarding/1158" target="_blank">Graphite</a> 
      * **PR #1160** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/web-onboarding/1160" target="_blank">Graphite</a> 
        * **PR #1161** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/web-onboarding/1161" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/web-onboarding/1161?utm_source=stack-comment).

src/client/components/DetailsModal/index.tsx
```diff
@@ -228,109 +227,108 @@ export const DetailsModal = ({
...
-          case 'subType':
-            if (marketLabel === 'SE') {
-              acc[key] =
-                mainQuote.id === quote.id
-                  ? getSubType(form.data)
-                  : getSubType(data)
-            } else {
+    for (const quote of allQuotes) {
+      const { id, data } = quote
+      const payload = {
```
guilhermespopolin, line 245: I understand it would be better to include in the payload only the fields (info) that has been changed, but there's no easy way to do it with formik (even with regular form I guess). I would need to deep compare `initialValues` with `values` passed into formik's `onSubmit` function. Do you guys think it is worth a try?

guilhermespopolin, line 245: I understand it would be better to include in the payload only the fields (info) that has been changed, but there's no easy way to do it with formik (even with regular form I guess). I would need to deep compare `initialValues` with `values` passed into formik's `onSubmit` function. Do you guys think it is worth a try?

src/client/components/DetailsModal/index.tsx
```diff
@@ -228,109 +227,108 @@ export const DetailsModal = ({
...
-            return acc
-          case 'subType':
-            if (marketLabel === 'SE') {
-              acc[key] =
-                mainQuote.id === quote.id
-                  ? getSubType(form.data)
-                  : getSubType(data)
-            } else {
+    for (const quote of allQuotes) {
+      const { id, data } = quote
```
robinandeer, line 244: nit: maybe rename id here so it's easier to understand

```suggestion
      const { id: quoteId, data } = quote
```

src/client/components/DetailsModal/index.tsx
robinandeer: I think it's more straightforward to send all the data every time :)

At least I can't really see why that would be a bad idea in practise 🤔

robinandeer (approved):
Fantastic job, this has been a long time coming! But everyone except you was too intimidated to touch this code 😅

[clap](https://app.graphite.dev/macros/HedvigInsurance)

src/client/components/DetailsModal/index.tsx
```diff
@@ -228,109 +227,108 @@ export const DetailsModal = ({
...
+                data['livingSpace'] ||
+                data['squareMeters']
+              acc.livingSpace =
+                form.data['livingSpace'] ||
+                form.data['squareMeters'] ||
+                data['livingSpace'] ||
+                data['squareMeters']
+              return acc
+            case 'subType':
+              if (marketLabel === 'SE') {
```
alebedev, line 271: Can you comment what's difference between SE and other markets?

src/client/components/DetailsModal/index.tsx
alebedev, line 245: Nit: I know this has not changed, but this reduce is hard to read.  Can iterative code be more readable?

alebedev (approved):

src/client/components/DetailsModal/index.tsx
guilhermespopolin: It seems we configure `subType` differently for SE. No idea why tho. Do you know @robinandeer ?

src/client/components/DetailsModal/index.tsx
guilhermespopolin: I've tried but it seems to make things even more complicated 😅

PR #1167, opened 2022-09-20
GRW-1495 / feat / Add tracking for when users click Cross sell promo on Download screen

simonauner:
## What?

* Track clicks on Cross sell promos on Download page

## Why?

* Otherwise we won't know if they're useful or not

**Ticket(s): [GRW-1495]**

[GRW-1495]: https://hedvig.atlassian.net/browse/GRW-1495?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/client/pages/Download/components/CrossSells.tsx
+++ b/src/client/pages/Download/components/CrossSells.tsx
@@ -13,6 +13,7 @@ import { useCurrentLocale } from 'l10n/useCurrentLocale'
 import { useTextKeys } from 'utils/textKeys'
 import { Arrow } from 'components/icons/Arrow'
 import { MEDIUM_SCREEN_MEDIA_QUERY } from 'utils/mediaQueries'
+import { useTrackEvent } from 'utils/tracking/gtm/useTrackEvent'
 import { useActiveContractBundles } from '../useActiveContractBundles'
 
 const CrossSellSectionHeader = styled.p({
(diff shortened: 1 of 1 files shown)


robinandeer (approved):

src/client/pages/Download/components/CrossSells.tsx
```diff
@@ -42,6 +44,13 @@ export const CrossSells = () => {
   const { crossSells } = useActiveContractBundles()
   const textKeys = useTextKeys()
   const currentLocale = useCurrentLocale()
+  const [trackEvent] = useTrackEvent('cross_sell_click')
+
+  const handleOnClick = (type: CrossSellType) => {
```
alebedev, line 49: Why extra wrapper that does not add value?

You can add `useCallback` or directly use `onClick={() => trackEven(crossSell.type)}`

alebedev (approved):
Looks good!

src/client/pages/Download/components/CrossSells.tsx
simonauner: Yeah, you're right! I did some debugging in that function before, but removed it before the PR and now it makes no sense to have that extra code. Will remove it 👍

PR #1170, opened 2022-09-27
Fix/update phone hours

amandaflod:
## What?

Updated phone hours and made lunch hour optional (IEX NO and SE answer phones during lunch as well). Also added DK number!

## Why?

Requested by IEX.

**Ticket(s): []**


alebedev (approved):
Nice working hours :)

PR #1180, opened 2022-10-07
Handle Trustly-payments from Hedvig.com

robinandeer:
## What?

Add new Trustly Payment page. This can be used by Hedvig.com as a payment widget:

1. Create authenticated session using `_hvsession`
2. Redirect to this new page `/new-member/payment?redirect_url=/se/confirmation`
3. Web Onboarding allows member to connect payment with Trustly
4. Web Onboarding redirects the user back to the `redirect_url` with a suffix: `/success | /fail`

Also did some refactoring in the current `ConnectPayment` page to e.g. use a proper `.graphql` mutation definition.

> There's still some design elements missing and string are not translated but let's use this to try out the solution!

## Why?

This way we can use this current setup to make payment connections with Giraffe in Hedvig.com.

**Ticket(s): [GRW-1607]**

[GRW-1607]: https://hedvig.atlassian.net/browse/GRW-1607?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/client/pages/Payment/useTrustlyURL.ts
```diff
@@ -0,0 +1,25 @@
+import { useEffect } from 'react'
+import { useLocation } from 'react-router'
+import { useRegisterDirectDebitMutation } from 'data/graphql'
+import { SUCCESS_SUFFIX, FAIL_SUFFIX } from './Payment.constants'
+
+export const useTrustlyURL = () => {
+  const { pathname } = useLocation()
+  const baseURL = `${window.location.origin}${pathname}`
```
guilhermespopolin, line 8: I wonder if there's any difference of doing this instead 🤔 

```suggestion
  const baseURL = window.location.href
```

I guess `window.location.href` will also include query parameters right?! It feels "weird" getting `baseURL` by using "2 location objects". Ah, whatever. No actions required here; just a random thought.

src/client/pages/ConnectPayment/components/TrustlyCheckout.tsx
```diff
@@ -45,21 +45,20 @@ export const TrustlyCheckout: React.FC<Props> = ({ onSuccess }) => {
     return res.data.registerDirectDebit.url
   }
 
+  const handleClick = async () => {
+    setTrustlyModalIsOpen(true)
+    const url = await generateTrustlyUrl()
+    if (!url) return
```
alebedev, line 51: ```if (url) setTrustlyUrl(url)```

And how should we handle empty URL?

src/client/pages/Payment/useRedirectURL.ts
```diff
@@ -0,0 +1,31 @@
+import { useLocation } from 'react-router'
+import { useMemo, useEffect } from 'react'
+
+const QUERY_PARAMETER = 'redirect_url'
+const SESSION_STORAGE_KEY = `_hedvig_${QUERY_PARAMETER}`
+
+export const useRedirectURL = () => {
```
alebedev, line 7: Let's explain why we're using localStorage as backup

src/client/pages/Payment/useRedirectURL.ts
```diff
@@ -0,0 +1,31 @@
...
+  }, [query])
+
+  return useMemo(() => {
+    const value = query.get(QUERY_PARAMETER)
+    if (value) return value
+
+    const storeValue = window.sessionStorage.getItem(SESSION_STORAGE_KEY)
+    if (storeValue) return storeValue
+
+    return undefined
```
alebedev, line 24: we can just return `storeValue`

src/client/pages/Payment/useRedirectURL.ts
```diff
@@ -0,0 +1,31 @@
...
+  }, [query])
+}
+
+const useQuery = () => {
+  const { search } = useLocation()
+  return useMemo(() => new URLSearchParams(search), [search])
```
alebedev, line 30: Cost of `new URLSearchParams` is trivial, why not just create it and extract redirect URL on every render?

src/client/pages/Payment/TrustlyIframe.tsx
```diff
@@ -0,0 +1,48 @@
...
+      }
+    }
+
+    window.addEventListener('message', handler)
+    return () => window.removeEventListener('message', handler)
+  }, [url])
+
+  const handleLoad: ReactEventHandler<HTMLIFrameElement> = (event) => {
+    const href = event.currentTarget.contentWindow?.location.href
+    if (href?.endsWith(SUCCESS_SUFFIX)) {
```
guilhermespopolin, line 31: Should we consider the eventual existence of query params into trustly iframe URL? 🤔 
I mean, if someday one is present this logic will fail since it's based on `href?.endsWith`. Maybe use `pathname` in this case.

src/client/pages/Payment/useTrustlyURL.ts
```diff
@@ -0,0 +1,25 @@
...
+  const [registerDirectDebit, { data }] = useRegisterDirectDebitMutation({
+    variables: {
+      clientContext: {
+        successUrl: [baseURL, SUCCESS_SUFFIX].join(''),
+        failureUrl: [baseURL, FAIL_SUFFIX].join(''),
+      },
+    },
+  })
+
+  useEffect(() => {
```
alebedev, line 18: It non-obvious for caller that useTrustlyUrl mutates something.  Can we call register on caller side?

src/client/pages/Payment/Payment.tsx
```diff
@@ -0,0 +1,82 @@
...
+
+  return (
+    <SessionTokenGuard>
+      <TopBar textColorVariant="dark">
+        <LanguagePicker color="black" />
+      </TopBar>
+      <Wrapper>
+        <Header>
+          <Heading>Connect direct debit</Heading>
+          <Paragraph>
```
alebedev, line 31: Planning to localize that later?

alebedev (approved):
Looks promising, can't wait to see full flow!

guilhermespopolin (approved):
Nicely done 💪🏻

src/client/pages/ConnectPayment/components/TrustlyCheckout.tsx
robinandeer: Hmm that would have to be an error messages of some kind... "try again" -- let's ask design!

src/client/pages/Payment/Payment.tsx
robinandeer: hehe of course ;)

src/client/pages/Payment/TrustlyIframe.tsx
robinandeer: right good point, it's been working until now though but perhaps we can take a stab at it anyways

src/client/pages/Payment/useRedirectURL.ts
robinandeer: true that :)

I'll write here for now:

We need to remember the redirect URL even if Trustly does a redirect to sign with BankID in the middle of the flow.

src/client/pages/Payment/useRedirectURL.ts
robinandeer: good point!

src/client/pages/Payment/useRedirectURL.ts
robinandeer: haha just picked this from react route docs

src/client/pages/Payment/useTrustlyURL.ts
robinandeer: hmm interesting -- we need `window.location` for the origin

then I use react router location since it will make the hook reactive but perhaps it's not really needed...

src/client/pages/Payment/useTrustlyURL.ts
robinandeer: right, I was confused about how to implement this -- it doesn't feel like it fits in a useEffect -- do you have a better idea?

I can also rename this function but I feel like it nice to co-locate it

src/client/pages/ConnectPayment/components/TrustlyCheckout.tsx
robinandeer: This is the OLD solution tho so didn't plan on changing much here

src/client/pages/Payment/useTrustlyURL.ts
guilhermespopolin: Yeah, we probably want it to be reactive so the correct `baseUrl` gets passed to `useRegisterDirectDebitMutation`. That's fine 😄

src/client/pages/Payment/useRedirectURL.ts
robinandeer: I have to wrap in memo though, otherwise it will cause useEffect to run over and over....

alebedev: Probably can be simplified like this
```
const search = useLocation()
const redirectUrl = // extract from search
useEffect(() => ..., [redirectUrl]
```
No memo needed since redirectUrl is a primitive

src/client/pages/Payment/useTrustlyURL.ts
alebedev: Maybe we can solve this by naming only
```
// Caller side
const trustlyUrl = useResigsterTrustlyPaument()
```
Now it's clear that hook initiates a mutation

alebedev (approved):

guilhermespopolin (approved):

src/client/pages/Payment/Payment.tsx
```diff
@@ -0,0 +1,98 @@
...
+    <SessionTokenGuard>
+      <Wrapper>
+        <PageHeader>
+          <LogoLink href={'/' + pathLocale}>
+            <HedvigLogo width={82} />
+          </LogoLink>
+        </PageHeader>
+        <Header>
+          <Heading>{textKeys.ONBOARDING_CONNECT_DD_CTA()}</Heading>
+          <Paragraph>
```
alebedev, line 36: When do you plan to translate text below?

src/client/pages/Payment/Payment.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { TrustlyIframe } from './TrustlyIframe'
+import { useStoredRedirectURL } from './useStoredRedirectURL'
+
+export const PaymentPage = () => {
+  const textKeys = useTextKeys()
+  const { path: pathLocale } = useCurrentLocale()
+  const redirectURL = useStoredRedirectURL()
+  const [, trustlyURL] = useCreateTrustlyURL()
+
+  const handleSuccess = (baseRedirectURL: string) => {
```
alebedev, line 18: Nit: if you capture redirectUrl here, you can simplify usage.  It would also enable adding `useCallback` when needed
```
onSuccess={handleSuccess}
```

alebedev (approved):

PR #1198, opened 2022-11-01
GRW-1721 / Feature / Support A/B testing Embark flows

alebedev:
## What?

Allow to A/B test Embark purchase flow.  No experiment_impression reports, that would be next PR

Very similar to A/B redirects in market-web
- Hardcoded redirect config
- Feature flag
- Env variable for weights configuration
- Set cookie to assign experiment variant to user, then redirect based on that cookie
- Remove dev/prod split of GTM config and use latest version instead of pinning specific revision

## Why?

First step to A/B test new Price matching flow that's going to be developed soon

**Ticket(s): []**


gustaveen (approved):
Cool! 🔀

guilhermespopolin (approved):

PR #1199, opened 2022-11-01
GRW-1721 / Feature / Track AB experiments

alebedev:
## What?

Send experiment_impression event when first showing embark flow that matches experiment cookie

## Why?

Allows us to track performance of A/B testing purchase flows in Google Optimize

**Ticket(s): []**


KarnellSchultz (approved):
lgtm

PR #1210, opened 2022-11-24
GRW-1798 / Feature / Exchange authorizationCode to accessToken when connecting payments for hedvig-com

alebedev:
## What?

Added /new-member/exchange-token ednpoint that gets accessToken, saves it cookie, then redirects to local URL specified by `next` param

## Why?

Safe way to do cross-domain token transfer that does not rely on shared cookie and therefore works in any environment

**Ticket(s): []**

### DIFF
--- a/src/server/exchangeToken.ts
+++ b/src/server/exchangeToken.ts
@@ -6,7 +6,7 @@ import { ServerCookieStorage } from 'utils/storage/ServerCookieStorage'
 import { WithLoggerState } from './middleware/enhancers'
 
 enum SearchParams {
-  AuthorizationCode = 'authorizationCode',
+  AuthorizationCode = 'authorization_code',
   Next = 'next',
 }
 
--- a/src/server/exchangeToken.ts
+++ b/src/server/exchangeToken.ts
@@ -16,7 +16,7 @@ export const exchangeTokenMiddleware: Router.IMiddleware<
 > = async (ctx: ParameterizedContext) => {
   const nextPath = ctx.URL.searchParams.get(SearchParams.Next)
   if (!nextPath) {
-    return ctx.throw(400, `Search param ${SearchParams.Next} is required`)
+    throw new Error(`Search param ${SearchParams.Next} is required`)
   }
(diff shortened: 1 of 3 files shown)


src/server/exchangeToken.ts
```diff
@@ -0,0 +1,68 @@
+import Axios from 'axios'
+import Router from 'koa-router'
+import { ParameterizedContext } from 'koa'
+import { createSession } from 'shared/sessionStorage'
+import { ServerCookieStorage } from 'utils/storage/ServerCookieStorage'
+import { WithLoggerState } from './middleware/enhancers'
+
+enum SearchParams {
+  AuthorizationCode = 'authorizationCode',
```
robinandeer, line 9: <i class='graphite__hidden'>[Re: lines 9 to 9]<br/></i>

Didn't you define this as `authorization_code` in racoon?<span class='graphite__hidden'><br/><br/>See this comment inline on <a href="https://app.graphite.dev/github/pr/HedvigInsurance/web-onboarding/1210?utm_source=unchanged-line-comment">Graphite</a>.</span>

src/server/exchangeToken.ts
```diff
@@ -0,0 +1,68 @@
...
+  nextUrl.searchParams.delete(SearchParams.AuthorizationCode)
+
+  const authorizationCode = ctx.URL.searchParams.get(
+    SearchParams.AuthorizationCode,
+  )
+  if (!authorizationCode) {
+    return ctx.throw(
+      400,
+      `Search param ${SearchParams.AuthorizationCode} is required`,
+    )
```
robinandeer, line 34: What happens when a user ends up here? Just thinking it would be nice to show an error page in the future...

robinandeer (approved):
Looks good! Just double check the search param names so they match :)

[looksgood](https://app.graphite.dev/macros/HedvigInsurance)

src/server/exchangeToken.ts
alebedev: Good catch!

PR #1220, opened 2022-12-02
Fix / Make AB variant cookie have 1w TTL as intended

alebedev:
## What?

Propery handle AB variant cookie lifetime.  Was using incorrect abstraction for it

## Why?

Clients should see consisten variant of AB test

**Ticket(s): []**


robinandeer (approved):
Nice! And every simpler :D

PR #1221, opened 2022-12-02
Fix / Report experiment_impression

alebedev:
## What?

Fix experiment_impression reports for Embark AB tests

## Why?

Previous MR made experiment cookie HTTP-only and we lost client-side reporting

**Ticket(s): []**


amandaflod (approved):
👍

PR #1227, opened 2022-12-21
GRW-1963 / Feature / Track experiment_impression events for new vs old site

alebedev:
## What?

Send Analytics event when we see user selected for old site during AB testing

Simple hardcoded config, no feature flags - sending event is always safe, worst case is that it's ignored

## Why?

Need to track conversion for new vs old site in Google Optimize

**Ticket(s): []**


robinandeer (approved):
Looks good afaict

PR #1228, opened 2022-12-27
Configure public host to correctly redirect from server

robinandeer:
## What?

Configure public host name to be able to correctly redirect from the server.

## Why?

We do redirect after performing access token exchange. It was redirecting to the wrong host name.

**Ticket(s): []**

### DIFF
--- a/src/server/exchangeToken.ts
+++ b/src/server/exchangeToken.ts
@@ -14,12 +14,15 @@ export const exchangeTokenMiddleware: Router.IMiddleware<
   WithLoggerState,
   any
 > = async (ctx: ParameterizedContext) => {
+  const logger = ctx.state.getLogger('exchangeTokenMiddleware')
+
   const nextPath = ctx.URL.searchParams.get(SearchParams.Next)
   if (!nextPath) {
     throw new Error(`Search param ${SearchParams.Next} is required`)
   }
-  // Safe way of building redirect URL.  We don't want to redirect to external site
+  // Safe way of building redirect URL. We don't want to redirect to external site
   const nextUrl = new URL('', ctx.URL)
+  nextUrl.host = process.env.HOST_URL || nextUrl.host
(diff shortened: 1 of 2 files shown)


robinandeer:
* master
  * **PR #1228** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/web-onboarding/1228" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/web-onboarding/1228?utm_source=stack-comment).

src/server/exchangeToken.ts
```diff
@@ -14,12 +14,15 @@ export const exchangeTokenMiddleware: Router.IMiddleware<
...
+  const logger = ctx.state.getLogger('exchangeTokenMiddleware')
+
   const nextPath = ctx.URL.searchParams.get(SearchParams.Next)
   if (!nextPath) {
     throw new Error(`Search param ${SearchParams.Next} is required`)
   }
-  // Safe way of building redirect URL.  We don't want to redirect to external site
+  // Safe way of building redirect URL. We don't want to redirect to external site
   const nextUrl = new URL('', ctx.URL)
+  nextUrl.host = process.env.HOST_URL || nextUrl.host
```
alebedev, line 25: We cannot use any VERCEL_* params for this, right?

alebedev (approved):
Looks good! 

If we manage our preview domains in Vercel, would it prevent such problems in the future?

src/server/exchangeToken.ts
robinandeer: This is deployed on Heroku 🤔

robinandeer:
@alebedev we don't have any plans to migrate this to Vercel - so no.

Our public-facing domain is managed outside of Heroku so I believe we need to pass this in

PR #1231, opened 2023-01-09
Update car switch notice to include link to BankSignering

robinandeer:
## What?

Update download page with new copy for switching notice and link to BankSignering.

Update type of contracts (trials replaced with group).

## Why?

So member is able to initate the flow to switch to BankSignering instead of getting email.

**Ticket(s): []**


amandaflod (approved):
🙌

guilhermespopolin (approved):

PR #1239, opened 2023-02-15
Add feature flag to toggle Intercom

robinandeer:
## What?

Add new feature flag to toggle intercom per market.

Conditionally show/hide related buttons and components.

## Why?

We want to disable intercom for sweden to ease load on IEX.

**Ticket(s): []**


robinandeer:
* master
  * **PR #1239** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/web-onboarding/1239" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/web-onboarding/1239?utm_source=stack-comment).

.env.example
```diff
@@ -14,6 +14,7 @@ FEATURE_COMPARISON_TABLE=SE
 FEATURE_CAR_CANCELLATION=SE
 FEATURE_CAR_PRICE_MATCHING=SE
 FEATURE_AB_TEST_PURCHASE_FLOWS=SE
+FEATURE_INTERCOM=NO,DK
```
alebedev, line 17: Just checking - was the idea to remove only from Sweden?

src/client/pages/Checkout/shared/CheckoutIntercomVariation.tsx
```diff
@@ -2,26 +2,33 @@ import React from 'react'
...
-          bottom: '87px!important',
-        },
-      }}
-    />
-    <TextLinkContainer>
-      <Intercom.TextLinkVariation />
-    </TextLinkContainer>
-  </>
-)
+export const CheckoutIntercomVariation = () => {
```
alebedev, line 14: Can we render this conditionally instead of returning null?

alebedev (approved):
Looks good, some questions inside

.env.example
robinandeer: yes exactly

src/client/pages/Checkout/shared/CheckoutIntercomVariation.tsx
robinandeer: hmm but then we need to add the check in a lot more places

I think this looked fine when testing it out - but it's a workaround

PR #1250, opened 2023-04-28
Directly call Giraffe endpoint rather than using a proxy

robinandeer:
## What?

Directly call Giraffe from browser rather than using a proxy

## Why?

Easier to setup with unified domain

No need to setup a proxy since we handle the CORS headers

**Ticket(s): []**

### DIFF
--- a/src/server/page.ts
+++ b/src/server/page.ts
@@ -40,7 +40,7 @@ const clientConfig: ClientConfig = {
   adyenClientKey: ADYEN_CLIENT_KEY,
   contentServiceEndpoint: CONTENT_SERVICE_ENDPOINT,
   giraffeHost: GIRAFFE_HOST,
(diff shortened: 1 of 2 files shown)


robinandeer:
* master
  * **PR #1250** <a href="https://app.graphite.dev/github/pr/HedvigInsurance/web-onboarding/1250" target="_blank">Graphite</a>  👈

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/HedvigInsurance/web-onboarding/1250?utm_source=stack-comment).

alebedev (approved):
Nice!  Can we delete koa-server-http-proxy now?

robinandeer:
> Nice!  Can we delete koa-server-http-proxy now?

no, I checked that but we still use it for some other use cases

PR #1251, opened 2023-05-02
Update call center info

gustaveen:
## What?
Update call center info

## Why?
We will no longer use Key solutions for phone sales

**Ticket(s): []**


alebedev (approved):
Looks good

PR #1252, opened 2023-05-12
Don't do token exchange on HTTP HEAD requests

alebedev:
## What?

Add early HTTP 204 to token exchange route to HEAD requests, we should not consume authorization_code in this case

## Why?

Potentially fixes a problem a lot of members see

**Ticket(s): []**


gustaveen (approved):
Let's try it!

